PR #1246, opened 2019-10-29
Add E-mail and MIME lexer

tzing:
This PR add lexer to parse raw E-mail source and MIME data.

It was originally designed for E-mail, but MIME format is common used in E-mail.
So a minimal MIME lexer is provided, it could handle the nested `multipart` format but it only processes few headers.
(Currently it only supports `Content-Type` and `Content-Transfer-Encoding`, which are necessary to recognize the data in the payload.)

Also, the `multipart/form-data` format, which is common used in HTTP transfer is not handled.

### DIFF
--- a/pygments/lexers/mime.py
+++ b/pygments/lexers/mime.py
@@ -51,10 +51,9 @@ class MIMELexer(RegexLexer):
 
     def __init__(self, **options):
         RegexLexer.__init__(self, **options)
-        self.content_type = options.get("Content_Type")
         self.boundary = options.get("boundary")
         self.content_transfer_encoding = options.get("Content_Transfer_Encoding")
-
+        self.content_type = options.get("Content_Type", "text/plain")
         self.max_nested_level = get_int_opt(options, "MIME-max-level", -1)
 
     def analyse_text(text):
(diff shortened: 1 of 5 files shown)


Anteru:
Queued for the next release, thanks!

pygments/lexers/mime.py
```diff
@@ -0,0 +1,224 @@
...
+        self.content_type = options.get("Content_Type", "text/plain")
+        self.max_nested_level = get_int_opt(options, "MIME-max-level", -1)
+
+    def analyse_text(text):
+        try:
+            header, body = text.strip().split("\n\n", 1)
+            if not body.strip():
+                return 0.1
+
+            invalid_headers = MIMELexer.tokens["header"].sub("", header)
```
alanhamlett, line 71: This line is raising an exception. What's the intention of `sub("", header)`? Is that meant to be `re.sub` or just a typo?

tzing: I guess this line is for removing all lines that could be a valid header.
Probably mistakenly treat `MIMELexer.tokens["header"]` as a re pattern.

alanhamlett: Oh, I see `MIMELexer.tokens["header"]` is a list of tuples that contain regex strings. Should it loop over that list and return any regex matches against `header`?

birkenfeld: I've removed this `analyse_text` for now.  It is not easy to find a good implementation that is not too broad for such a generic format as MIME.

Anyone: If you have one, please open a PR.

PR #1257, opened 2019-11-19
Add Sieve lexer

sblondon:
This PR adds a Sieve lexer.
Sieve is a language to filter incoming e-mails into the user mailbox.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -32,7 +32,7 @@ Other contributors, listed alphabetically, are:
 * Sébastien Bigaret -- QVT Operational lexer
 * Jarrett Billingsley -- MiniD lexer
 * Adam Blinkinsop -- Haskell, Redcode lexers
-* Stéphane Blondon -- SGF lexer
+* Stéphane Blondon -- SGF and Sieve lexers
 * Frits van Bommel -- assembler lexers
 * Pierre Bourdon -- bugfixes
 * Matthias Bussonnier -- ANSI style handling for terminal-256 formatter
--- a/pygments/lexers/_mapping.py
+++ b/pygments/lexers/_mapping.py
@@ -394,6 +394,7 @@ LEXERS = {
     'ScssLexer': ('pygments.lexers.css', 'SCSS', ('scss',), ('*.scss',), ('text/x-scss',)),
     'ShExCLexer': ('pygments.lexers.rdf', 'ShExC', ('shexc', 'shex'), ('*.shex',), ('text/shex',)),
     'ShenLexer': ('pygments.lexers.lisp', 'Shen', ('shen',), ('*.shen',), ('text/x-shen', 'application/x-shen')),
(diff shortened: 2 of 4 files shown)


Anteru:
Thanks for your contribution -- looks like it's causing some test failures though. Could you please take a look at them? If you need any help, please let me know!

sblondon:
@Anteru I made some improvements but it still does not pass the test. I don't understand why because running:
`./pygmentize -v -O full tests/examplefiles/example.sieve`
works properly.

I get the same error like Github CI when I run `tox`:
```
E           AssertionError: lexer <pygments.lexers.SieveLexer> generated error token for /home/stephane/src/pygments/tests/examplefiles/example.sieve: '.' at position 1109
E           assert Token.Error != Token.Error
```

The tests are green if the file `./tests/examplefiles/example.sieve` is empty.
If `example.sieve` contains only:
```
require
```
I get another error, perhaps it's an hint about the issue:
```
[...]
>               assert False, absfn
E               AssertionError: /home/stephane/src/pygments/tests/examplefiles/example.sieve
E               assert False

tests/test_examplefiles.py:141: AssertionError
---------------------------------------------------------------- Captured stdout call ----------------------------------------------------------------
--- 

+++ 

@@ -1 +1 @@

-[(Token.Punctuation, '\n')]
+[(Token.Keyword.Namespace, 'require'), (Token.Punctuation, '\n')]
```

Do you have any idea?

sblondon:
After `rm -rf tests/examplefiles/output`, the error generated by `example.sieve` file disappears.

There are two errors in python3.7  with `example.praat` and `example.rkt` files which are equivalent (missing token). For `example.praat`:
```
- (Token.Name, '#%A|||b #true C\n\\|ｄ\u2002'),
- (Token.Text, ''),
+ (Token.Name, '#%A|||b #true C\n\\|ｄ'),
+ (Token.Text, ''),
+ (Token.Text, '\u2002'),
```
These errors are also in `master` branch of `pygments`.

birkenfeld:
You saw the failure generated by cached output in `tests/examplefiles/output` because a special test mode was active which shouldn't be in the normal case, I've fixed that in d338695666a8255ec1dd188a2fa653fc42516aa3.

This shouldn't matter for GitHub CI since it always starts with a clean checkout.

Running the tests locally with your branch, I now get the following error:
```
E           AssertionError: lexer <pygments.lexers.SieveLexer> generated error token for /home/gbr/devel/pygments/tests/examplefiles/example.sieve: '.' at position 1109
```
which also occurs on the CI.

sblondon:
@birkenfeld Thanks for your help! I rebased the branch and I get the same error too. I will investigate it.

I wonder if `make clean` should remove the `tests/examplefiles/output/` directory. I can do it in another PR.

birkenfeld:
> I wonder if `make clean` should remove the `tests/examplefiles/output/` directory. I can do it in another PR.

Yes, that's a good idea in any case.

sblondon:
I fixed the test. `$` was not understood as end-of-line in regular expression in 'text' block so I replaced it by `\n`.

PR #1261, opened 2019-11-24
Update Logtalk lexer for the latest language specification

pmoura:
Tested using the sample Logtalk source file available at:
https://github.com/LogtalkDotOrg/logtalk3/tree/master/coding/tests

### DIFF
--- a/pygments/lexers/prolog.py
+++ b/pygments/lexers/prolog.py
@@ -107,19 +107,19 @@ class LogtalkLexer(RegexLexer):
             (r'\n', Text),
             (r'\s+', Text),
             # Numbers
-            (r"0'.", Number),
+            (r"0'[\\]?.", Number),
             (r'0b[01]+', Number.Bin),
             (r'0o[0-7]+', Number.Oct),
             (r'0x[0-9a-fA-F]+', Number.Hex),
             (r'\d+\.?\d*((e|E)(\+|-)?\d+)?', Number),
             # Variables
-            (r'([A-Z_]\w*)', Name.Variable),
+            (r'([A-Z_][a-zA-Z0-9_]*)', Name.Variable),
             # Event handlers
             (r'(after|before)(?=[(])', Keyword),
             # Message forwarding handler
             (r'forward(?=[(])', Keyword),
             # Execution-context methods
-            (r'(parameter|this|se(lf|nder))(?=[(])', Keyword),
+            (r'(context|parameter|this|se(lf|nder))(?=[(])', Keyword),
             # Reflection
             (r'(current_predicate|predicate_property)(?=[(])', Keyword),
             # DCGs and term expansion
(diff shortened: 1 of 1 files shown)


pygments/lexers/prolog.py
```diff
@@ -226,16 +228,16 @@ class LogtalkLexer(RegexLexer):
...
-            (r"[a-z]\w*", Text),
-            (r"'", String, 'quoted_atom'),
+            (r"[a-z][a-zA-Z0-9_]*", Text),
+            (r"[']", String, 'quoted_atom'),
         ],
 
         'quoted_atom': [
-            (r"''", String),
-            (r"'", String, '#pop'),
+            (r"['][']", String),
```
Anteru: Can you please shed some light on this? Why is a character class selector?

pygments/lexers/prolog.py
```diff
@@ -226,16 +228,16 @@ class LogtalkLexer(RegexLexer):
             (r'\^', Operator),
             # Strings
             (r'"(\\\\|\\"|[^"])*"', String),
-            # Ponctuation
+            # Punctuation
             (r'[()\[\],.|]', Text),
             # Atoms
-            (r"[a-z]\w*", Text),
-            (r"'", String, 'quoted_atom'),
+            (r"[a-z][a-zA-Z0-9_]*", Text),
```
Anteru: ` \w` should be equivalent? Why this change?

pygments/lexers/prolog.py
```diff
@@ -244,44 +246,43 @@ class LogtalkLexer(RegexLexer):
...
-            (r'(alias|d(ynamic|iscontiguous)|m(eta_(non_terminal|predicate)|ode|ultifile)|'
-             r's(et_(logtalk|prolog)_flag|ynchronized))(?=[(])', Keyword, 'root'),
+            (r'(built_in|dynamic|synchronized|threaded)(?=[.])', Keyword, 'root'),
+            (r'(alias|d(ynamic|iscontiguous)|m(eta_(non_terminal|predicate)|ode|ultifile)|s(et_(logtalk|prolog)_flag|ynchronized))(?=[(])', Keyword, 'root'),
             (r'op(?=[(])', Keyword, 'root'),
             (r'(c(alls|oinductive)|module|reexport|use(s|_module))(?=[(])', Keyword, 'root'),
-            (r'[a-z]\w*(?=[(])', Text, 'root'),
-            (r'[a-z]\w*[.]', Text, 'root'),
+            (r'[a-z][a-zA-Z0-9_]*(?=[(])', Text, 'root'),
+            (r'[a-z][a-zA-Z0-9_]*(?=[.])', Text, 'root'),
```
Anteru: Same as above -- is there a reason to not use `\w`?

pmoura: Same reason as above: avoid possible issues with Unicode code points.

pygments/lexers/prolog.py
pmoura: Is my understanding that `\w` can include Unicode code points besides ASCII characters, which may result in an invalid atom.

pygments/lexers/prolog.py
pmoura: No longer remember the motif. These changes are old (the original pull request was on BitBucket more than one year ago). But it works fine as I'm using this submitted version on a regular basis on a patched Pygments installation.

pygments/lexers/prolog.py
Anteru: Ah yes, correct. I'm not sure we do unicode yet, but I guess if the language is limited to ASCII, doesn't hurt to make this explicit. Thanks!

pygments/lexers/prolog.py
Anteru: Can you please give it a shot with `"''"` again? Other than that, should be ok to merge.

pmoura: Replace occurrences of `[']` with `'` and re-run the tests. No issues. Pushed the changes.

pygments/lexers/prolog.py
birkenfeld: Well, Unicode is explicitly selected by re.UNICODE being in the flags for the lexer. If that's not accurate it can just be deleted.

pygments/lexers/prolog.py
```diff
@@ -244,44 +246,43 @@ class LogtalkLexer(RegexLexer):
         'directive': [
             # Conditional compilation directives
             (r'(el)?if(?=[(])', Keyword, 'root'),
-            (r'(e(lse|ndif))[.]', Keyword, 'root'),
+            (r'(e(lse|ndif))(?=[.])', Keyword, 'root'),
```
birkenfeld: If a dot always follows, can it be highlighted in this rule?

pmoura: As you may know, Logtalk (and Prolog) source files are made of *terms* (directives and clauses; clauses can be facts or rules) that must always end with a dot. Thus, consistent highlighting would require to all ending dots be highlighted, not just in this particular case.

pygments/lexers/prolog.py
pmoura: That explicit selection is done only in the Prolog lexer, not in the Logtalk lexer. Even there, is not clear. The ISO Prolog standards are mute on Unicode. Some Prolog systems support it, most don't. Among the Prolog systems that do support Unicode, there's no common de facto standard, particularly regarding using of non ASCII code points in identifiers.

PR #1262, opened 2019-11-24
Make Python 3 the default language for "PythonLexer".

birkenfeld:


birkenfeld:
Thanks for the review!

eric-wieser:
Does Pygments have a general policy about reassigning aliases like this?

My guess is that there are many people with old latex documents using pygments via `minted`, and when used via platforms like overleaf the user is not exactly in control or even really aware of the pygments version they are using. As a result, from their perspective their document suddenly "breaks" when shifting to a newer TeXlive and highlights the wrong language.

Of course, almost all python2 programs will render identically with the python3 lexer, so here the cost is probably justified; and at any rate, the ship sailed 6 years ago! But I'm curious if any discussions happened that could apply to similar renames happening in future.

Anteru:
We don't have a firm policy, and we had a similar case with Perl I think. The general assumption is that you _can_ somehow control your Python/Pygments version, if you really care -- as in, if you find a 10 year old LaTeX document, you can also reconstruct what Pygments version you used. Eventually we are dropping old Python versions, too (as in, Pygments won't support Python 3.2 for example any more), and there's no hope to maintain compatibility that far back for languages (note that also the languages themselves change, so if you wrote some code 10 years ago about missing keywords in some language which were correctly highlighted as errors back then, but added in the meantime based on your work, you'd also get a broken document, even though nothing else changed.)

PR #1264, opened 2019-11-24
Add Notmuch lexer

dan-santana:


pygments/lexers/textfmts.py
```diff
@@ -295,3 +297,86 @@ class TodotxtLexer(RegexLexer):
...
+        'part': [
+            (r'\f(?:part|attachment)}\n', Keyword, '#pop'),
+            (r'\f(?:part|attachment){\s*', Keyword, ('#push', 'part-attr')),
+            (r'^Non-text part: .*\n', Comment),
+            (r'(?s)(.*?(?=\f(?:part|attachment)}\n))', _highlight_code),
+        ],
+    }
+
+    def analyse_text(text):
+        return 1.0 if text.startswith('\fmessage{') else 0.0
```
Anteru: That seems overly confident? Probably better to return `0.01` in that case?

birkenfeld: If that's really a formfeed character, confidence is probably pretty well justified :smile:

Anteru: Oh, missed the `\f`. I guess that's solved then, we just need a sample file (starting with a formfeed.)

dan-santana:
> Thanks for the new lexer! Can you please provide some simple example file so we can ensure it works?

Ooops, I forgot to commit the example file! it's there now :)

pygments/lexers/textfmts.py
```diff
@@ -295,3 +297,86 @@ class TodotxtLexer(RegexLexer):
             (r'\s+', IncompleteTaskText),
         ],
     }
+
+
+class NotmuchLexer(RegexLexer):
+    """
+    For `Notmuch <https://notmuchmail.org/>`_ email text format.
+
+    .. versionadded:: 2.4
```
birkenfeld: it's 2.5 now :)

pygments/lexers/textfmts.py
```diff
@@ -295,3 +297,86 @@ class TodotxtLexer(RegexLexer):
...
+
+        try:
+            if self.body_lexer:
+                lexer = get_lexer_by_name(self.body_lexer)
+            else:
+                lexer = guess_lexer(code.strip())
+        except ClassNotFound:
+            lexer = get_lexer_by_name('text')
+
+        for item in do_insertions([], lexer.get_tokens_unprocessed(code)):
```
birkenfeld: I don't think `do_insertions` with an empty insertions list does anything other than a `for ... in lexer.get_tokens_...`.

dan-santana: fixed

pygments/lexers/textfmts.py
dan-santana: fixed!

PR #1266, opened 2019-11-24
Fix #1253.

### LINKED ISSUE #1253: Console lexer cannot detect virtual environment names
When a Python virtual environment is activated, by default it puts the virtual environment's name to the left of the prompt in a shell session. E.g.:

```console
~$ source myenv/bin/activate
(myenv) ~$
```

This causes Pygments to fail to detect the prompt in the `console`/`shell-session` lexer. Compare:

Without the virtual environment:

```console
$ pygmentize -l console -f html <<<"[~/project]$ foo -h"
<div class="highlight"><pre><span></span><span class="gp">[~/project]$</span> foo -h
</pre></div>
```

With the virtual environment:
```console
$ pygmentize -l console -f html <<<"(env) [~/project]$ foo -h"
<div class="highlight"><pre><span></span><span class="go">(env) [~/project]$ foo -h</span>
</pre></div>
```

Note that it *is* detected if there is nothing after the environment name or possibly a user@host:dir without brackets:

```console
$ pygmentize -l shell-session -f html <<<"(env)a@b:x$ foo -h"
<div class="highlight"><pre><span></span><span class="gp">(env)a@b:x$</span> foo -h
</pre></div>
```

**Context:**

Some software I develop is documented with Sphinx. Since I encourage users of my software to use virtual environments, I would like to illustrate their use in my documentation, but the lexer fails to parse my examples if I use the virtual environment names. I think this is just a matter of accounting for an optional virtual environment name in BashSessionLexer._ps1rgx, but there may be complexities I'm not considering.

Anteru:
This seems to do the trick, but I'm not 100% certain if that's the correct fix.

### DIFF
--- a/pygments/lexers/shell.py
+++ b/pygments/lexers/shell.py
@@ -154,6 +154,9 @@ class ShellSessionBaseLexer(Lexer):
 
     .. versionadded:: 2.1
     """
+
+    _venv = re.compile(r'^(\([^)]*\))(\s*)')
+
     def get_tokens_unprocessed(self, text):
         innerlexer = self._innerLexerCls(**self.options)
 
@@ -164,11 +167,24 @@ class ShellSessionBaseLexer(Lexer):
 
         for match in line_re.finditer(text):
             line = match.group()
-            m = re.match(self._ps1rgx, line)
             if backslash_continuation:
                 curcode += line
                 backslash_continuation = curcode.endswith('\\\n')
-            elif m:
+                continue
+            
+            venv_match = self._venv.match(line)
+            if venv_match:
+                venv = venv_match.group(1)
+                venv_whitespace = venv_match.group(2)
+                insertions.append((len(curcode),
+                    [(0, Generic.Prompt.VirtualEnv, venv)]))
+                if venv_whitespace:
+                    insertions.append((len(curcode),
+                        [(0, Text, venv_whitespace)]))
+                line = line[venv_match.end():]
+
+            m = self._ps1rgx.match(line)
+            if m:
                 # To support output lexers (say diff output), the output
                 # needs to be broken by prompts whenever the output lexer
                 # changes.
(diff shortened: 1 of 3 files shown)


pygments/lexers/shell.py
```diff
@@ -165,6 +168,7 @@ def get_tokens_unprocessed(self, text):
         for match in line_re.finditer(text):
             line = match.group()
             m = re.match(self._ps1rgx, line)
+            venv_re = re.match(self._venv, line)
```
birkenfeld: better use a precompiled regex. and the result should be e.g. `venv_match` not `venv_re`.

Anteru:
No, that's not intentional. I do get some highlighting there (`Token.Operator`, `[`) -- how can I get the lexer to process everything "as-if" I have cut the front and just return "normal tokens"?

pygments/lexers/shell.py
Anteru: Changed to a precompiled regex (should this also apply to `_ps1rgx`? That's a hefty one in comparison) and renamed as requested.

birkenfeld:
> No, that's not intentional. I do get some highlighting there (`Token.Operator`, `[`) -- how can I get the lexer to process everything "as-if" I have cut the front and just return "normal tokens"?

I think you need to check the venv match before the "m" match, and still jump into the rest of the conditions. So in short:
```
if continuation: ... continue
if venv: ... cut off something from line
m = ps1.match(line)
if m: ...
... rest of clauses
```

pygments/lexers/shell.py
birkenfeld: yeah, ps1rgx should be precompiled too.

It's all a bit less important because re has a cache for regexes passed in as strings, but we're using lots of them and may exhaust the cache.

Anteru:
Either I'm dumb, or there's more to this. I've changed it to:

```python
if backslash_continuation:
    # existing code

venv_match = self._venv.match(line)
if venv_match:
    venv_match = venv_match.group(0)
    insertions.append((len(curcode),
        [(0, Generic.Prompt, line[:len(venv_match)])]))
    # Cut line
    line = line[len(venv_match):]

m = self._ps1rgx.match(line)
```

but that produces:

```
Token.Generic.Prompt    u'(env)'
Token.Generic.Output    u' [~/project]$ foo -h\n'
```

If I modify `curcode` to add the line (after cutting), I get the same result as before. This is analogous to how `_ps2` is handled, and I'm getting really confused here.

birkenfeld:
I suspect that the PS1 regex doesn't like the leading space. This should go in the venv regex.

Anteru:
I think that did the trick -- thanks! Feel free to merge if that looks good to you now :)

Anteru:
Correct on the trailing space. Looks like I sorted this out, the `backslash_continuation` should also trigger a `continue` from what I can tell, and that was the missing ingredient.

PR #1269, opened 2019-11-24
Add Zeek lexer based on the Bro lexer

jsiwek:
Bro has been renamed to Zeek, but the language is essentially the
same without any different treatment of .zeek files from .bro files.

This change also adds general improvements to the lexer.

Not sure if `.. versionadded:: 2.5` for `ZeekLexer` was the right thing for me to add.

Original PR: https://bitbucket.org/birkenfeld/pygments-main/pull-requests/824

### DIFF
--- a/pygments/lexers/dsls.py
+++ b/pygments/lexers/dsls.py
@@ -16,7 +16,7 @@ from pygments.lexer import ExtendedRegexLexer, RegexLexer, bygroups, words, \
 from pygments.token import Text, Comment, Operator, Keyword, Name, String, \
     Number, Punctuation, Literal, Whitespace
 
-__all__ = ['ProtoBufLexer', 'BroLexer', 'PuppetLexer', 'RslLexer',
+__all__ = ['ProtoBufLexer', 'ZeekLexer', 'BroLexer', 'PuppetLexer', 'RslLexer',
            'MscgenLexer', 'VGLLexer', 'AlloyLexer', 'PanLexer',
            'CrmshLexer', 'ThriftLexer', 'FlatlineLexer', 'SnowballLexer']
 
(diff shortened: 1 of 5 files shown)


pygments/lexers/dsls.py
```diff
@@ -188,84 +188,174 @@ class ThriftLexer(RegexLexer):
...
+        'operators': [
+            (r'[!%*/+<=>~|&^-]', Operator),
             (r'([-+=&|]{2}|[+=!><-]=)', Operator),
-            (r'(in|match)\b', Operator.Word),
-            (r'[{}()\[\]$.,;]', Punctuation),
-            # Identfier
-            (r'([_a-zA-Z]\w*)(::)', bygroups(Name, Name.Namespace)),
+            (r'\b(in|as|is|of)\b', Operator.Word),
+            (r'\??\$', Operator),
+            # Technically, colons are often used for punctuation/sepration.
```
Anteru: Typo: Should be punctuation/separation

jsiwek: Thanks, pushed a new commit to fix.

jsiwek:
> I'm not super happy with the IPv4/IPv6 patterns, but assuming it works ...

They are pretty gnarly, but the same patterns for Zeek highlighting are at least being used in other various projects without complaint so far: SublimeText (same rules also used in linguist / GitHub highlighting), Vim, Ace editor

The "not 100% correct" comment related to IPv6 I recall is related to it being more lenient than it technically should be in some cases, like in terms of allowing more than a single `::` zero-compressed sequence.

pygments/lexers/dsls.py
```diff
@@ -188,84 +188,174 @@ class ThriftLexer(RegexLexer):
...
-        ]
+            (r'.', String.Regex),
+        ],
     }
 
 
+class BroLexer(ZeekLexer):
+    """
+    For `Zeek <https://www.zeek.org/>`_ scripts.
+    Note, this is equivalent to ZeekLexer (Bro is the old name for Zeek).
```
birkenfeld: I think this should be done by moving the alias/filename pattern to ZeekLexer, and just having a `BroLexer = ZeekLexer` for compatibility.

pygments/lexers/other.py
```diff
@@ -27,8 +27,8 @@
 from pygments.lexers.business import ABAPLexer, OpenEdgeLexer, \
     GoodDataCLLexer, MaqlLexer
 from pygments.lexers.automation import AutoItLexer, AutohotkeyLexer
-from pygments.lexers.dsls import ProtoBufLexer, BroLexer, PuppetLexer, \
-    MscgenLexer, VGLLexer
+from pygments.lexers.dsls import ProtoBufLexer, ZeekLexer, BroLexer, \
```
birkenfeld: Since this is a compatibility module, ZeekLexer shouldn't go here.

pygments/lexers/dsls.py
```diff
@@ -188,84 +188,174 @@ class ThriftLexer(RegexLexer):
...
-            (r'((\d+|:)(' + _hex + r'|:)*)?::(' + _hex + r'|:)*', Number),
-            (r'(\d+\.\d+\.|(\d+\.){2}\d+)', Number),
+        ],
+
+        'comments': [
+            (r'#.*$', Comment),
+        ],
+
+        'directives': [
+            (r'(@(load-plugin|load-sigs|load|unload))\b.*$', Comment.Preproc),
```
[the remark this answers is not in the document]
jsiwek: Didn't look like it, now removed.

pygments/lexers/dsls.py
```diff
@@ -188,84 +188,174 @@ class ThriftLexer(RegexLexer):
...
+            (r'\b(T|F)\b', Keyword.Constant),
+
+            # Port
+            (r'\b\d{1,5}/(udp|tcp|icmp|unknown)\b', Number),
+
+            # IPv4 Address
+            (r'\b(25[0-5]|2[0-4][0-9]|[0-1][0-9]{2}|[0-9]{1,2})\.(25[0-5]|2[0-4][0-9]|[0-1][0-9]{2}|[0-9]{1,2})\.(25[0-5]|2[0-4][0-9]|[0-1][0-9]{2}|[0-9]{1,2})\.(25[0-5]|2[0-4][0-9]|[0-1][0-9]{2}|[0-9]{1,2})\b', Number),
+
+            # IPv6 Address (not 100% correct: that takes more effort)
+            (r'\[([0-9a-fA-F]{0,4}:){2,7}([0-9a-fA-F]{0,4})?((25[0-5]|2[0-4][0-9]|[0-1][0-9]{2}|[0-9]{1,2})\.(25[0-5]|2[0-4][0-9]|[0-1][0-9]{2}|[0-9]{1,2})\.(25[0-5]|2[0-4][0-9]|[0-1][0-9]{2}|[0-9]{1,2})\.(25[0-5]|2[0-4][0-9]|[0-1][0-9]{2}|[0-9]{1,2}))?\]', Number),
```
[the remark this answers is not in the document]
jsiwek: The IPv4 part (now using simpler pattern per previous comment) is optional and there to match the [IPv4-mapped IPv6 format](https://en.wikipedia.org/wiki/IPv6#IPv4-mapped_IPv6_addresses):

> For example, ::ffff:192.0.2.128 represents the IPv4 address 192.0.2.128

pygments/lexers/dsls.py
```diff
@@ -188,84 +188,174 @@ class ThriftLexer(RegexLexer):
...
             (r'([-+=&|]{2}|[+=!><-]=)', Operator),
-            (r'(in|match)\b', Operator.Word),
-            (r'[{}()\[\]$.,;]', Punctuation),
-            # Identfier
-            (r'([_a-zA-Z]\w*)(::)', bygroups(Name, Name.Namespace)),
+            (r'\b(in|as|is|of)\b', Operator.Word),
+            (r'\??\$', Operator),
+            # Technically, colons are often used for punctuation/separation.
+            # E.g. field name/type separation.
+            (r'[?:]', Operator),
```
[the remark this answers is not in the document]
jsiwek: Sure, but either ways seems to end up deserving of a code-comment: the problem was that `:` is used either in "ternary if" like `foo ? expr1 : expr2`  or in variable declarations like `local foo: int;`.  The later is most common and disambiguating requires more full-blown parsing treatment, so I do like the idea of just using the less-prominent Punctuation for both `?` and `:` for sake of consistency and now changed it to that.

pygments/lexers/dsls.py
jsiwek: I've hit a point where I may need to modify a unit test to get that way to work, so could you help take a look at what you'd like done here or else point me at another example where this has been done?

Here's the patch I tried:

```diff
diff --git a/pygments/lexers/_mapping.py b/pygments/lexers/_mapping.py
index 92aaca39..5405785f 100644
--- a/pygments/lexers/_mapping.py
+++ b/pygments/lexers/_mapping.py
@@ -65,7 +65,7 @@ LEXERS = {
     'BooLexer': ('pygments.lexers.dotnet', 'Boo', ('boo',), ('*.boo',), ('text/x-boo',)),
     'BoogieLexer': ('pygments.lexers.verification', 'Boogie', ('boogie',), ('*.bpl',), ()),
     'BrainfuckLexer': ('pygments.lexers.esoteric', 'Brainfuck', ('brainfuck', 'bf'), ('*.bf', '*.b'), ('application/x-brainfuck',)),
-    'BroLexer': ('pygments.lexers.dsls', 'Bro', ('bro',), ('*.bro',), ()),
+    'BroLexer': ('pygments.lexers.dsls', 'Zeek', ('zeek', 'bro'), ('*.zeek', '*.bro'), ()),
     'BugsLexer': ('pygments.lexers.modeling', 'BUGS', ('bugs', 'winbugs', 'openbugs'), ('*.bug',), ()),
     'CAmkESLexer': ('pygments.lexers.esoteric', 'CAmkES', ('camkes', 'idl4'), ('*.camkes', '*.idl4'), ()),
     'CLexer': ('pygments.lexers.c_cpp', 'C', ('c',), ('*.c', '*.h', '*.idc'), ('text/x-chdr', 'text/x-csrc')),
@@ -474,7 +474,7 @@ LEXERS = {
     'XtlangLexer': ('pygments.lexers.lisp', 'xtlang', ('extempore',), ('*.xtm',), ()),
     'YamlJinjaLexer': ('pygments.lexers.templates', 'YAML+Jinja', ('yaml+jinja', 'salt', 'sls'), ('*.sls',), ('text/x-yaml+jinja', 'text/x-sls')),
     'YamlLexer': ('pygments.lexers.data', 'YAML', ('yaml',), ('*.yaml', '*.yml'), ('text/x-yaml',)),
-    'ZeekLexer': ('pygments.lexers.dsls', 'Zeek', ('zeek',), ('*.zeek',), ()),
+    'ZeekLexer': ('pygments.lexers.dsls', 'Zeek', ('zeek', 'bro'), ('*.zeek', '*.bro'), ()),
     'ZephirLexer': ('pygments.lexers.php', 'Zephir', ('zephir',), ('*.zep',), ()),
     'ZigLexer': ('pygments.lexers.zig', 'Zig', ('zig',), ('*.zig',), ('text/zig',)),
 }
diff --git a/pygments/lexers/dsls.py b/pygments/lexers/dsls.py
index d88bac99..5894606f 100644
--- a/pygments/lexers/dsls.py
+++ b/pygments/lexers/dsls.py
@@ -195,8 +195,8 @@ class ZeekLexer(RegexLexer):
     .. versionadded:: 2.5
     """
     name = 'Zeek'
-    aliases = ['zeek']
-    filenames = ['*.zeek']
+    aliases = ['zeek', 'bro']
+    filenames = ['*.zeek', '*.bro']
 
     _hex = r'[0-9a-fA-F]'
     _float = r'((\d*\.?\d+)|(\d+\.?\d*))([eE][-+]?\d+)?'
@@ -345,16 +345,7 @@ class ZeekLexer(RegexLexer):
     }
 
 
-class BroLexer(ZeekLexer):
-    """
-    For `Zeek <https://www.zeek.org/>`_ scripts.
-    Note, this is equivalent to ZeekLexer (Bro is the old name for Zeek).
-
-    .. versionadded:: 1.5
-    """
-    name = 'Bro'
-    aliases = ['bro']
-    filenames = ['*.bro']
+BroLexer = ZeekLexer
 
 
 class PuppetLexer(RegexLexer):
```

Here's the failing unit test:

```
    def test_get_lexers():
        # test that the lexers functions work
        for func, args in [(lexers.get_lexer_by_name, ("python",)),
                           (lexers.get_lexer_for_filename, ("test.py",)),
                           (lexers.get_lexer_for_mimetype, ("text/x-python",)),
                           (lexers.guess_lexer, ("#!/usr/bin/python -O\nprint",)),
                           (lexers.guess_lexer_for_filename, ("a.py", "<%= @foo %>"))
                           ]:
            x = func(opt='val', *args)
            assert isinstance(x, lexers.PythonLexer)
            assert x.options["opt"] == "val"
    
        for cls, (_, lname, aliases, _, mimetypes) in lexers.LEXERS.items():
>           assert cls == lexers.find_lexer_class(lname).__name__
E           AssertionError: assert 'BroLexer' == 'ZeekLexer'
E             - BroLexer
E             + ZeekLexer

tests/test_basic_api.py:124: AssertionError
```

Except that test does sound legit the way it is: if `BroLexer` remains in `LEXERS` at all (think it needs to for compat), then it's going to need a unique `name` attribute ?

jsiwek:
@birkenfeld thanks for the feedback, I pushed a new commit to address all of it except the `BroLexer = ZeekLexer` issues.

pygments/lexers/dsls.py
birkenfeld: I don't think the old one needs to stay in `LEXERS`, as long as the old aliases and filenames work with the new one...

jsiwek: But the `_mapping.py` file containing the `LEXERS` definition auto-generates itself, so either need to change the generation code to ignore `BroLexer` somehow or else move it to a different place where that logic won't find it?

birkenfeld: Just needs to be removed from `__all__`.

jsiwek: Thanks, sounds good, new commit does that.  Let me know if there's anything further.

PR #1271, opened 2019-11-24
Improvements to Matlab lexer

anntzer:
- Detect `.m` files starting with a function definition as MATLAB, not
  ObjC.
- Require word boundaries in regexes matching numbers and floats, to
  avoid mishighlighting `load 123file` as starting with a number.
- MATLAB treats a bare word followed by arguments such as `foo bar baz` as
  the function call `foo('bar', 'baz')`.  As such, treat everything that
  follows the bare word as a string.

Just a revival of https://bitbucket.org/birkenfeld/pygments-main/pull-requests/676/matlab-detection-float-boundaries/diff.

### DIFF
--- a/pygments/lexers/matlab.py
+++ b/pygments/lexers/matlab.py
@@ -99,6 +99,12 @@ class MatlabLexer(RegexLexer):
             # operators requiring escape for re:
             (r'\.\*|\*|\+|\.\^|\.\\|\.\/|\/|\\', Operator),
 
+            # numbers (must come before punctuation to handle `.5`; cannot use
+            # `\b` due to e.g. `5. + .5`).
+            (r'(?<!\w)((\d+\.\d*)|(\d*\.\d+))([eEf][+-]?\d+)?(?!\w)', Number.Float),
+            (r'\b\d+[eEf][+-]?[0-9]+\b', Number.Float),
+            (r'\b\d+\b', Number.Integer),
+
             # punctuation:
             (r'\[|\]|\(|\)|\{|\}|:|@|\.|,', Punctuation),
             (r'=|:|;', Punctuation),
@@ -106,10 +112,6 @@ class MatlabLexer(RegexLexer):
             # quote can be transpose, instead of string:
             # (not great, but handles common cases...)
             (r'(?<=[\w)\].])\'+', Operator),
-
-            (r'(\d+\.\d*|\d*\.\d+)([eEf][+-]?[0-9]+)?', Number.Float),
-            (r'\d+[eEf][+-]?[0-9]+', Number.Float),
-            (r'\d+', Number.Integer),
 
             (r'(?<![\w)\].])\'', String, 'string'),
             (r'[a-zA-Z_]\w*', Name),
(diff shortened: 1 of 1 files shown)


pygments/lexers/matlab.py
```diff
@@ -72,6 +72,8 @@ class MatlabLexer(RegexLexer):
              "hilb", "invhilb", "magic", "pascal", "rosser", "toeplitz", "vander",
              "wilkinson")
 
+    _operators = r'-|==|~=|<|>|<=|>=|&&|&|~|\|\|?|\.\*|\*|\+|\.\^|\.\\|\.\/|\/|\\'
```
birkenfeld: this will find `<` before `<=`, the order has to be different (I know you just moved the definition)

Or, better, use `words()` which will automatically escape and order the regex to avoid this problem.

pygments/lexers/matlab.py
```diff
@@ -131,12 +138,26 @@ class MatlabLexer(RegexLexer):
...
+            ("[^';\n]+", String),
+            (";?\n?", Punctuation, '#pop'),
+        ]
     }
 
     def analyse_text(text):
-        if re.match(r'^\s*%', text, re.M):  # comment
+        # function declaration.
+        if next(line for line in text.splitlines()
+                if not re.match(r'^\s*%', text)).strip().startswith('function'):
```
birkenfeld: I think you can just `re.search(r'^function\b')` with `re.MULTILINE`.

Although this is a bit problematic for safe identification, since e.g. JavaScript functions are also defined that way.  (analyse_text can also be used when no filename is known.)

Maybe you can also match the equals sign which I gather is always part of the function header?

pygments/lexers/matlab.py
```diff
@@ -94,10 +96,22 @@ class MatlabLexer(RegexLexer):
             # line continuation with following comment:
             (r'\.\.\..*$', Comment),
 
+            # command form:
+            # "How MATLAB Recognizes Command Syntax" specifies that an operator
+            # is recognized if it is either surrounded by spaces or by no
+            # spaces on both sides; only the former case matters for us.  (This
+            # allows distinguishing `cd ./foo` from `cd ./ foo`.)
+            (r'(?:(?<=^)|(?<=;))\s*\w+\s+(?!=|\(|(%s)\s+)' % _operators, Name,
```
birkenfeld: No lookbehind is needed for `^`.

pygments/lexers/matlab.py
anntzer: I don't think the single regex is equivalent, the point is that the *first* line that isn't fully commented should start with `function`.
No, there isn't always an equal sign (if the function doesn't return anything).

pygments/lexers/matlab.py
anntzer: I just reordered the elements, will pass on the bigger refactoring.

pygments/lexers/matlab.py
anntzer: fair point.

anntzer:
thanks, all comments handled.

pygments/lexers/matlab.py
birkenfeld: ah, I missed that, sorry. Still very possible in Javascript though... maybe exclude if `{` comes later?

anntzer: sure, done

birkenfeld:
Thanks!

PR #1274, opened 2019-11-24
Assume setuptools is available

jdufresne:
On modern Python installations, setuptools should always be available
and should always be preferred. Can remove the fallback to distutils and
the pygmentize script workaround.

### DIFF
--- a/pygments/__main__.py
+++ b/pygments/__main__.py
@@ -1,7 +1,6 @@
-#!/usr/bin/env python2
-
 import sys
 import pygments.cmdline
+
 try:
     sys.exit(pygments.cmdline.main(sys.argv))
 except KeyboardInterrupt:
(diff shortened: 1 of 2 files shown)


Anteru:
@birkenfeld You can still do that using `pip install -e .`, which will make `pygmentize` available in the current environment. Can you give that a try? If that works for you as well, there's really no upside in keeping the `pygmentize` script as-is.

birkenfeld:
Sorry, I don't like all that "pip develop" stuff that changes anything outside the project. I'm also not keen on having to use a virtualenv for a package like Pygments, which has no other dependencies.

I would be ok with moving the contents into `pygments/__main__.py`, then I can use `python -m pygments ...`

jdufresne:
I like the idea of adding a `__main__.py`. I've done that in the latest revision.

birkenfeld:
Nice!

PR #1276, opened 2019-11-24
Based on the Vim theme: http://www.vim.org/scripts/script.php?script_…

ideasman42:
Based on the Vim theme: http://www.vim.org/scripts/script.php?script_id=1143

Example output: http://members.iinet.net.au/~ideasman42/dotemacs/init.html

The theme is also bundled with qt-creator and available from emacs package manager.

Also, there aren't many nice dark themes for Pygments at the moment.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -23,6 +23,10 @@ Version 2.5.0
   * Python3 (PR#1255)
   * YAML (#1528)
 
+- Added styles:
+
+  * Inkpot (#1276)
+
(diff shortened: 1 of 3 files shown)


birkenfeld:
Thanks! Do you want to add an entry to CHANGES?

ideasman42:
@birkenfeld done.

birkenfeld:
Thanks!

PR #1278, opened 2019-11-25
Fixes #1124

### LINKED ISSUE #1124: Matlab double quoted string is not highlighted
(Original issue 1420 created by None on 2018-03-01T18:25:18.198712+00:00)

Matlab double quoted string is not highlighted. Matlab now supports double quoted strings. It is very desirable that such strings are highlighted.

a = 'string'  => this is highlighted

b = "string" => this is NOT highlighted

wallysalami:
Highlights Matlab double quoted string

### DIFF
--- a/pygments/lexers/matlab.py
+++ b/pygments/lexers/matlab.py
@@ -121,6 +121,8 @@ class MatlabLexer(RegexLexer):
             # (not great, but handles common cases...)
             (r'(?<=[\w)\].])\'+', Operator),
 
+            (r'"[^"]*"', String),
+
             (r'(?<![\w)\].])\'', String, 'string'),
             (r'[a-zA-Z_]\w*', Name),
             (r'.', Text),


birkenfeld:
Is there any escaping of double quotes inside these strings possible?

wallysalami:
Matlab escapes double quotes using two of them consecutively, like this:

```matlab
text = "Total of ""Hello there"" memes"; % → Total of "Hello there" memes
```

I still don't know how that regex accepts this case (I copied it from Octave's lexer), but it works!

Anteru:
That should get highlighted as three separate tokens -- `"Total of"`, `"Hello there"`, `" memes"`, instead of one token with the `""` being part of the string. It will not break though with a single `""` inside the string, but it's not the correct solution either :)

wallysalami:
I thought that would happen, but I tested in my website via [python-markdown2](https://github.com/trentm/python-markdown2) and I got just one token:

Captura de Tela 2019-11-26 às 12 03 13

Maybe markdown2 made some magic there? I don't know. How can I test it using only pygments?

Anteru:
Assuming you have it working locally, you can check using:

`python -m pygments -v -l console -f raw <<<"your matlab code here"`

wallysalami:
`python3 -m pygments -v -l console -f raw <<< 'text = "Total of ""Hello there"" memes'`

results in 

`Token.Generic.Output	'text = "Total of ""Hello there"" memes\n'`

I should define the language as "Matlab" somehow, right?

Anteru:
Sorry, yes, I edited my comment. `console` should have been `matlab`, `-l foo` sets the lexer to `foo`.

wallysalami:
Ok, got it. The following command

`python3 -m pygments -v -l matlab -f raw <<< 'text = "Total of ""Hello there"" memes'`

results in 

```
Token.Name	'text'
Token.Text	' '
Token.Punctuation	'='
Token.Text	' '
Token.Literal.String	'"Total of "'
Token.Literal.String	'"Hello there"'
Token.Text	'"'
Token.Text	' '
Token.Name	'memes'
Token.Text	'\n'
```

It looks like it was some magic of markdown2 after all. Bummer.

birkenfeld:
Well, the HTML formatter coalesces parts with the same token type (actually, with the same CSS style; multiple token types can map to the same CSS style).  The raw formatter does not.

So in the end, I don't think it matters too much.

birkenfeld:
I've fixed up the regex to recognize escapes now.

PR #1279, opened 2019-11-25
reStructuredText: improve handling of field lists

EricFromCanada:
The current field list rules only find a field body if it starts on the same line as the field marker, as shown by this [code sample](http://www.sphinx-doc.org/en/stable/domains.html#the-javascript-domain) in the Sphinx docs. This replacement rule covers any field marker (including colons if escaped) and an optional field body consisting of one or more paragraphs, starting on either the same line or a following line, as long as it's indented more than the marker.

### DIFF
--- a/pygments/lexers/markup.py
+++ b/pygments/lexers/markup.py
@@ -222,10 +222,9 @@ class RstLexer(RegexLexer):
                       Punctuation, Text, using(this, state='inline'))),
             # Comments
             (r'^ *\.\..*(\n( +.*\n|\n)+)?', Comment.Preproc),
-            # Field list
-            (r'^( *)(:[a-zA-Z-]+:)(\s*)$', bygroups(Text, Name.Class, Text)),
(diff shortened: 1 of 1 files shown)


birkenfeld:
Hmm, IIUC now the field body is uniformly highlighted, but it can (especially if multiline) contain markup to highlight itself.

I'd rather just ensure that field names on their own (with the body following on the next line) are highlighted properly, and maybe remove the special highlight of in-line field bodies.

EricFromCanada:
Just so we're clear, by "can contain markup to highlight itself", you mean that because the field body may contain its own inline markup like roles or emphasis, you'd rather have it skip highlighting the field body entirely? I can do that.

birkenfeld:
Yes, the field body should not be consumed in the field rule.

At least for multiline fields; I don't know if the lexer likes starting in the middle of a line, I remember that it was pretty line-by-line-oriented.

EricFromCanada:
Done; it now only highlights field list markers.

PR #1290, opened 2019-11-27
Add USD lexer

ColinKennedy:
Ported from: https://bitbucket.org/birkenfeld/pygments-main/pull-requests/840/add-usd-lexer

## Current Status
The code below was run against roughly 300 USD files that ship with Pixar's test suite. Everything is 100% working except for one exception case which I was hoping to get advice on.

Any time there's a multi-line string, it seems, non-alphanumeric characters register as an error.

e.g.

```
    int[] creaseLengths = [] (
        doc = """The length of this array specifies the number of creases on the
        surface. Each element gives the number of (must be adjacent) vertices in
        each crease, whose indices are linearly laid out in the 'creaseIndices'
        attribute. Since each crease must be at least one edge long, each
        element of this array should be greater than one.""")
```

The ```'```s in this paragraph 

As do characters like ```->``` and ```\\```

```
    inherits = </APISchemaBase>
    doc = """API schema to support discovery and publishing of lights in a scene.

    \\section UsdLuxListAPI_Discovery Discovering Lights via Traversal

    To motivate this API, consider what is required to discover all
    lights in a scene.  We must load all payloads and traverse all prims:

    \\code
    01  // Load everything on the stage so we can find all lights,
    02  // including those inside payloads
    03  stage->Load();
    04
    05  // Traverse all prims, checking if they are of type UsdLuxLight
    06  // (Note: ignoring instancing and a few other things for simplicity)
    07  SdfPathVector lights;
    08  for (UsdPrim prim: stage->Traverse()) {
    09      if (prim.IsA<UsdLuxLight>()) {
    10          lights.push_back(i->GetPath());
    11      }
    12  }
    \\endcode
```

Any advice on how to fix this would be very appreciated. Otherwise that issue, please let me know what else you'd like to have changed. Thank you!


pygments/lexers/usd.py
```diff
@@ -0,0 +1,69 @@
...
+def _keywords(words, type_):
+    """Declare the given words as full-match tokens."""
+    return [(lexer.words(words, prefix=r"\b", suffix=r"\b"), type_)]
+
+
+class UsdLexer(lexer.RegexLexer):
+    """
+    A lexer that parses Pixar's Universal Scene Description file format.
+
+    .. versionadded:: 2.5.0
```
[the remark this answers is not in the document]
ColinKennedy: Whoops. Done!

ColinKennedy:
Hi @Anteru 

Just checked on my end and everything is getting captured correctly! Thanks for your suggestion. I tried that regex rule but I put it in the wrong position so it didn't work in every case.

I've also added single-quote and double-quote variations and everything seems to work great now. Please let me know if you have any other notes.

Anteru:
For the last error you're seeing, the fix should be as easy as adding `r` in front of the string so you don't need to escape `\`.

ColinKennedy:
Tests added and CI passes now. Thanks for your quick replies, @Anteru ! Please let me know if you'd like anything else changed

ColinKennedy:
My pleasure! I don't have write access so I'll leave it to someone who does to handle the merge

pygments/lexers/usd.py
```diff
@@ -0,0 +1,136 @@
...
+            (",", token.Generic),
+            (";", token.Generic),  # ";"s are allowed to combine separate metadata lines
+            ("=", token.Operator),
+            ("[-]?([0-9]*[.])?[0-9]+", token.Number),
+            (r"'''(?:.|\n)*?'''", token.String),
+            (r'"""(?:.|\n)*?"""', token.String),
+            (r"'.*'", token.String),
+            (r'".*"', token.String),
+            (r"<(\.\./)*([\w/]+|[\w/]+\.\w+[\w:]*)>", token.Name.Namespace),
+            (r"@.*@", token.String.Interpol),
```
birkenfeld: these rules usually have to be written with `.*?` to avoid greedy highlighting of a line like `@firststring@ something else @secondstring@`.

Same for double- and single quoted strings.

pygments/lexers/usd.py
```diff
@@ -0,0 +1,136 @@
...
+        + _keywords(_usd_builtins.SPECIAL_NAMES, token.Name.Builtins)
+        + _keywords(_usd_builtins.COMMON_ATTRIBUTES, token.Name.Attribute)
+        + [(r"\b\w+:[\w:]+\b", token.Name.Attribute)]
+        + _keywords(_usd_builtins.OPERATORS, token.Operator)  # more attributes
+        + [(type_ + r"\[\]", token.Keyword.Type) for type_ in _usd_builtins.TYPES]
+        + _keywords(_usd_builtins.TYPES, token.Keyword.Type)
+        + _PUNCTUATION
+        + [
+            ("#.*?$", token.Comment.Single),
+            (",", token.Generic),
```
birkenfeld: should be either `Text` (no highlighting at all) or `Punctuation`

pygments/lexers/usd.py
```diff
@@ -0,0 +1,136 @@
...
+            (r"'''(?:.|\n)*?'''", token.String),
+            (r'"""(?:.|\n)*?"""', token.String),
+            (r"'.*'", token.String),
+            (r'".*"', token.String),
+            (r"<(\.\./)*([\w/]+|[\w/]+\.\w+[\w:]*)>", token.Name.Namespace),
+            (r"@.*@", token.String.Interpol),
+            (r'\(.*"[.\\n]*".*\)', token.String.Doc),
+            (r"\A#usda .+$", token.Comment.Hashbang),
+            (r"\s+", token.Text),
+            (r"[\w|_|:|\.]+", token.Generic),
```
birkenfeld: I think this should be `[\w_:.]+`; the `|` does nothing in character classes (except add itself to the class).

pygments/lexers/usd.py
```diff
@@ -0,0 +1,136 @@
...
+            (";", token.Generic),  # ";"s are allowed to combine separate metadata lines
+            ("=", token.Operator),
+            ("[-]?([0-9]*[.])?[0-9]+", token.Number),
+            (r"'''(?:.|\n)*?'''", token.String),
+            (r'"""(?:.|\n)*?"""', token.String),
+            (r"'.*'", token.String),
+            (r'".*"', token.String),
+            (r"<(\.\./)*([\w/]+|[\w/]+\.\w+[\w:]*)>", token.Name.Namespace),
+            (r"@.*@", token.String.Interpol),
+            (r'\(.*"[.\\n]*".*\)', token.String.Doc),
```
birkenfeld: again, the `.*` here is a bit dangerous.

pygments/lexers/usd.py
```diff
@@ -0,0 +1,136 @@
...
+
+    name = "USD"
+    aliases = ["usd", "usda"]
+    filenames = ["*.usd", "*.usda"]
+
+    tokens = {
+        "root": [
+            (
+                r"(custom){_WHITESPACE}(uniform)(\s+){}(\s+){}(\s*)(=)".format(
+                    _TYPE, _BASE_ATTRIBUTE, _WHITESPACE=_WHITESPACE
```
birkenfeld: What's the difference here between using `WHITESPACE` and `\s+`?

ColinKennedy: When \s was used, \n was getting matched. This isn't allowed by USD. It has to be separated by space or tabs. If there's an alternative that's more expressive, please suggest it and I'll add it

pygments/lexers/usd.py
ColinKennedy: Ah I forgot about that. I'll remove them

pygments/lexers/usd.py
ColinKennedy: Oh, didn't know about that. Thanks - done on all accounts + added a test for this

pygments/lexers/usd.py
ColinKennedy: I'll add that for ; too. Looking at it again, the `(r"[\w_:\.]+", Generic)` rule should be Text, as well. Done

pygments/lexers/usd.py
```diff
@@ -0,0 +1,136 @@
+#!/usr/bin/env python
```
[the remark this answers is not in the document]
ColinKennedy: > use the standard file header

Sure, you mean how the other modules just define encoding and don't provide a shebang line, right? Done

> import conventions (e.g. import tokens as from token import Punctuation)

I can't tell if you're suggesting to

- not use `tokens.Tokens.Foo` and instead use the re-defined names as `tokens.Foo`
- change the imports to absolute 
- use imports of class / function names, instead of module names.

or some combination of the 3. The first item is my bad and is changed. But the other two I think you should reconsider. (I'm not trying to start an argument or anything, just saying you should consider it).

pygments/lexers/compiled.py
```diff
@@ -30,5 +30,6 @@
 from pygments.lexers.felix import FelixLexer
 from pygments.lexers.nimrod import NimrodLexer
 from pygments.lexers.crystal import CrystalLexer
+from pygments.lexers.usd import UsdLexer
```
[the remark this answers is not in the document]
ColinKennedy: Okay, removed

pygments/lexers/usd.py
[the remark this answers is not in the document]
birkenfeld: For the file header, I meant the standard docstring with module name etc. But that's a detail, I'll add that post merging.

About the imports I'm sorry I was a bit vague. But your latest version is what I wanted, so I'm happy.

pygments/lexers/usd.py
birkenfeld: ok, I overlooked that difference!

pygments/lexers/usd.py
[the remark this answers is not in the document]
ColinKennedy: :+1:

ColinKennedy:
Sweet, thanks for the merge. Really excited for the next Pygments release :)

PR #1291, opened 2019-11-27
GAS: Add support for C style comments.

lrh2000:
GNU AS supports C style comments, but some errors occur when I try to highlight them using pygments. So I try to add support for it.

Here are some examples (x64 assembly).

The following code is considered illegal by GNU AS,
```
movq %rax, /* comments
  comments */ %rbx
```
Instead, the following code is considered legal,
```
movq %rax, %rbx /* comments
  comments */ movq %rax, %rbx
```
and
```
movq %rax, /* comments */ %rbx
```
It is corresponding to my pull request code.

In the end, I highlight this code:
```
.section ".text"
.global main
.code64

/*
 * Hello world!
 */

/* comments */ main: // comments
  lock addq $0, /* comments */ (%rsp) /*
    comments
  */ xorq %rax, %rax
  /* comments */retq
```
The result is
a
It seems that there are no problems.

### DIFF
--- a/pygments/lexers/asm.py
+++ b/pygments/lexers/asm.py
@@ -55,9 +55,9 @@ class GasLexer(RegexLexer):
             (number, Number.Integer),
             (register, Name.Variable),
             (r'[\r\n]+', Text, '#pop'),
-            (r'([;#]|//).*?\n', Comment, '#pop'),
-            (r'/[*].*?[*]/', Comment),
-            (r'/[*].*?\n[\w\W]*?[*]/', Comment, '#pop'),
+            (r'([;#]|//).*?\n', Comment.Single, '#pop'),
+            (r'/[*].*?[*]/', Comment.Multiline),
+            (r'/[*].*?\n[\w\W]*?[*]/', Comment.Multiline, '#pop'),
 
             include('punctuation'),
             include('whitespace')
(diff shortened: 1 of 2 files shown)


Anteru:
Thanks for the PR! This pops the comment state when you have a line comment (`//`), is this desired? Can you please try it on something like this
```c
/* this is
// a comment
surrounded by a comment */```

Does that get highlighted correctly? Could you also please add comments to the example file or a test for this?

File: pygments/lexers/asm.py [hunk 1]
```diff
@@ -55,7 +55,9 @@ class GasLexer(RegexLexer):
             (number, Number.Integer),
             (register, Name.Variable),
             (r'[\r\n]+', Text, '#pop'),
-            (r'[;#].*?\n', Comment, '#pop'),
+            (r'([;#]|//).*?\n', Comment, '#pop'),
+            (r'/[*].*?[*]/', Comment),
+            (r'/[*].*?\n[\w\W]*?[*]/', Comment, '#pop'),
```
Anteru: Is this `#pop` needed here? I'm looking at the C lexer, and it uses
```python
(r'/[*](.|\n)*?[*]/', Comment.Multiline),
```

for multi-line comments, which I presume would be sufficient  for GAS as well? Plus there are two separate rules for what should be the same type of comment?

lrh2000: Yes, I think it is necessary to `#pop` here.

It is because GAS thinks the current instruction is ended and it will parse the following code as the next instruction when GAS sees `\n` in multi-line comments. Please see the first two examples I mentioned in the pull request for detail.

The case is different in C language. It will never care whether there is `\n` in multi-line comments or not.

lrh2000:
> Thanks for the PR! Can you please try it on something like this
> 
> ```c
> /* this is
> // a comment
> surrounded by a comment */```
> 
> Does that get highlighted correctly? Could you also please add comments to the example file or a test for this?
> ```

I think that it will be correct. And I also added some test code just now. Hope there are no problems.

What's more, I also noticed that commit 7827966acdb6431636520d20fc3c148ce52de59b ignores the assembly language (Why?). Now it should be fixed.

PR #1293, opened 2019-11-28
Added support for the Kappa language.

hmedina:
Kappa is a rule transformation language for simulating biochemical signaling cascades. The tokens used in other languages were not suitable for Kappa, so I created new ones. This necessitated a new style-sheet, for which two are presented.

If it would be desirable to declare the tokens elsewhere, please advise.

[this PR is taking up where the old mercurial PR-811 left off]

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -133,6 +133,7 @@ Other contributors, listed alphabetically, are:
 * Brian McKenna -- F# lexer
 * Charles McLaughlin -- Puppet lexer
 * Kurt McKee -- Tera Term macro lexer
+* Hector Medina -- Kappa lexer and styles
 * Lukas Meuser -- BBCode formatter, Lua lexer
 * Cat Miller -- Pig lexer
 * Paul Miller -- LiveScript lexer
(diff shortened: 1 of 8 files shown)


birkenfeld:
Hmm, not using the standard tokens is a no-go, I'm afraid. I really don't want the multitude of tokens to expand further, and have different styles that are only usable with one lexer.

If you can't use the standard types, I think you're better off with keeping this as a 3rd party package.

hmedina:
I see. I'll repackage this as a plugin so the idiosyncrasies are kept at bay. On that note, the links on the Pygments documentation are broken, specifically those on https://pygments.org/docs/plugins/ pointing to the documentation for setup tools; they're no longer hosted at peak[dot]telecommunity[dot]com.

A MWE would be most welcome on that page, as even the entry-point ini-style labeling within triple-quoted strings is buried in the setup tools docs.

birkenfeld:
Thanks for the pointer, I'll make necessary updates to the page and include examples.

PR #1295, opened 2019-11-28
Added ThingsDB Language support

birkenfeld:
Ported from https://bitbucket.org/birkenfeld/pygments-main/pull-requests/839


Anteru:
@birkenfeld Anything else we need to do here?

PR #1299, opened 2019-11-28
Fixed split lines errors; moved contract keywords to a separate categ…

birkenfeld:
…ory; added a rule for triggers.

Ported from https://bitbucket.org/birkenfeld/pygments-main/pull-requests/758


birkenfeld:
@aterga thanks for the PR!

PR #1303, opened 2019-11-28
Update C and C++ lexers to support more literal formats

birkenfeld:
Ported from https://bitbucket.org/birkenfeld/pygments-main/pull-requests/723


Anteru:
Latest master supports hexadecimal float literals, closing.

mosra:
Hi @Anteru, sorry for commenting on a long-closed PR, but these patches added much more than just hex float literals. In particular it added support for arbitrary C++11 user-defined literals (say, `123.0_foo`).

The C++ lexer code probably changed significantly since then so the old patch would probably not apply anymore, nevertheless would it be possible to integrate the rest? Thank you :)

It would also resolve issues like https://github.com/pygments/pygments/issues/1847.

PR #1310, opened 2019-11-28
This commit adds a lexer for linux kernel logs as outputted by `dmesg`

birkenfeld:
It supports output from `dmesg`, in that case it highlights based on
keywords in the line

It can also highlight `dmesg -x` output. In that case it uses the
loglevels from the kernel to highlight the lines.

Ported from https://bitbucket.org/birkenfeld/pygments-main/pull-requests/815


pygments/lexers/textfmts.py
```diff
@@ -380,3 +380,52 @@ def analyse_text(text):
...
+            include('base'),
+            (r'.+\n', Comment, '#pop')
+        ],
+        'info': [
+            include('base'),
+            (r'.+\n', Text, '#pop')
+        ],
+        'warn': [
+            include('base'),
+            (r'.+', Generic.Strong, '#pop')
```
Anteru: Why does this not use `\n` at the end like the other categories?

birkenfeld: Looks like an oversight...

PR #1313, opened 2019-11-28
Address issue 829: highlight `self` and `parent` as classes, and func…

birkenfeld:
…tion-calls as functions

Ported from https://bitbucket.org/birkenfeld/pygments-main/pull-requests/661


birkenfeld:
This is maybe somewhat inconsistently used, but the original intent was for `Name.Class` and `Name.Function` to be used only at the *definition* sites, and as a result the highlighting for those tends to be kind of heavy in styles...

Anteru:
Rebased this on master, fixed the tests. I see no harm in merging this, it just changes `Name.Builtin` to `Name.Function` in a bunch of places.

PR #1314, opened 2019-11-28
[Scala] Fix #972: Resurrect patch from bug tracker

### LINKED ISSUE #972: Scala lexer incorrectly highlights code containing type parameters
(Original issue 1265 created by acieroid on 2016-07-26T14:21:37.111928+00:00)

See the following fragment: http://pygments.org/demo/5862224/ (repeated below for completeness)
Code appearing after a trait definition with a type parameter is incorrectly highlighted.

```
#!scala


// Incorrect highlighting (on `case`)
trait Foo[A]
case class Foo1(x: A) extends Foo
// Correct highlighting
case class Foo2(x: A) extends Foo

// Incorrect highlighting (on `extends`)
trait Bar[A] extends Foo[A]
// Correct highlighting
trait Bar extends Foo
```

Blaisorblade:
Issue: Scala lexer incorrectly highlights code containing type parameters.

This is the patch from https://github.com/pygments/pygments/issues/972#issuecomment-526852069, tested by
https://github.com/pygments/pygments/issues/972#issuecomment-526852076.

### DIFF
--- a/pygments/lexers/jvm.py
+++ b/pygments/lexers/jvm.py
@@ -271,8 +271,7 @@ class ScalaLexer(RegexLexer):
             # method names
             (r'(class|trait|object)(\s+)', bygroups(Keyword, Text), 'class'),
             (r'[^\S\n]+', Text),
-            (r'//.*?\n', Comment.Single),
-            (r'/\*', Comment.Multiline, 'comment'),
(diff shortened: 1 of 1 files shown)


birkenfeld:
merged as part of #1314

PR #1315, opened 2019-11-28
[Scala] Fix type highlighting

Blaisorblade:
On top of #1314, fix the following issue: `type` is recognized as a keyword even if inside a word, which isn't correct. To wit, below `types` is an identifier, not the `type` keyword followed by an `s`:

```
trait Symbol(val tpe: types.Type)
trait Symbol(val tpe: Option[types.Type])
```

image

I expect `forSome` needs the same treatment, but I haven't tested that.

Also, I'm a Scala dev, not a Python one, and I might be doing something silly.


birkenfeld:
Looks good, thanks!

PR #1316, opened 2019-11-28
Ensure `setup.py` works again

Blaisorblade:
While `pip3 install .` still works, `python3 setup.py install` does not (and
ditto for Python2), because subpackages of `pygments` don't get installed. Even
running `find_packages` on its own has consistent behavior:

```
$ python3
>>> from setuptools import setup, find_packages
>>> find_packages(exclude=['tests'])
['pygments', 'pygments.filters', 'pygments.lexers', 'pygments.formatters', 'pygments.styles']
>>> find_packages(include=['pygments'])
['pygments']
```

The bug seems to have been introduced in
8d0828bbfc5b8deb7525a19e1037704f644563da.

Tested on macOS 10.14 with Homebrew Python 2.7.17 and 3.7.5.

I'm no Python expert, but my fix is based on examples in https://setuptools.readthedocs.io/en/latest/setuptools.html#using-find-packages.

That `packages` does not recurse automatically is made clear by distutils docs:
https://docs.python.org/3.7/distutils/setupscript.html#listing-whole-packages.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,7 @@ setup(
     description = 'Pygments is a syntax highlighting package written in Python.',
     long_description = __doc__,
     keywords = 'syntax highlighting',
-    packages = find_packages(include=['pygments']),
+    packages = find_packages(exclude=['tests']),
     entry_points = {
         'console_scripts': ['pygmentize = pygments.cmdline:main'],
     },


birkenfeld:
I tested this and was able to `setup.py install` fine inside a virtualenv, and all the subpackages were installed, and `pygmentize` worked.

But I only tested Py3.6 and a newish version of setuptools, so there might be some version specifics going on, and obviously you're right that `packages` lists have to be comprehensive.

So I'll give in and use `exclude` after all.

birkenfeld:
2.5.2 is released with this change.

Anteru:
Uh, should this be maybe `include=["pygments.*", "pygments"]`? Does that work? Reading the documentation, that should do it. For exclusion, it should be `["tests", "tests.*"]` -- see https://setuptools.readthedocs.io/en/latest/setuptools.html#using-find-packages.

@birkenfeld Do you want to change this yet again or leave as is?

birkenfeld:
Feel free :)

Anteru:
Ok :)

Anteru:
Should be fixed for real in 01f05b4dc5155745d401f31545fc1a089e8a3710. Tried it locally using Python 2.7.

birkenfeld:
No need for a new release this time.

PR #1301, opened 2019-11-28
Ride language support

birkenfeld:
Ported from https://bitbucket.org/birkenfeld/pygments-main/pull-requests/819

PR #1319, opened 2019-11-30
Support Ride language (Waves blockchain smart-contracts)

potan:


birkenfeld:
Thanks for the PR! Does this supersede #1301?

potan:
> Thanks for the PR! Does this supersede #1301?

Yes, it does.

birkenfeld:
Ah, and I also added https://github.com/wavesplatform/ride-examples/blob/master/auction/ride/auction.ride as an example file for the test suite.  Had to fix allowing `!` as an operator...

PR #1320, opened 2019-12-01
Rust: overhaul, cover more builtins and newer syntax

birkenfeld:


birkenfeld:
I'm not sure if the macro rules were meant to catch macro definitions or calls, since the rule isn't correct for either one.  In any case, they never fired because the initial [A-z]\w+ would have been detected as a `Name` first. 

And it's a good thing, because in the `macro*` states, only braces would have been allowed which doesn't make sense.

PR #1322, opened 2019-12-02
Parse HTTP status-line even when reason-phrase not included.

jakubroztocil:
Background: https://github.com/jakubroztocil/httpie/issues/811


jakubroztocil:
@birkenfeld a non-capturing group is indeed a better solution — I’ve update the code.

birkenfeld:
Thanks!

PR #1323, opened 2019-12-03
Add Linux and Windows configuration aliases for org-mode (Emacs) and minted compatibility

Phundrak:
Minted, a LaTeX package, relies on pygments to generate syntax highlighting in code blocks sent to it. In org-mode, code blocks must be declared accordingly with their mode in which they are edited in Emacs, and this mode name will be exported in LaTeX as the name of the language Minted has to take care of. For instance, this:
```org
#+BEGIN_SRC conf-windows
[color]
#+END_SRC
```
gets exported to this:
```latex
\begin{minted}[]{conf-windows}
[color]
\end{minted}
```
However, minted sends the code to be colored as code with the `conf-windows` syntax, which is not understood by pygments, and this results in the code blocks not being exported in the LaTeX export from org-mode. Same goes for `conf` source blocks which are not recognized as code with the Linux configuration syntax.

This pull request aims to add compatibility between org-mode and pygments when the latter is used through minted as a result of LaTeX export.

I have tried to use this modified version on my local machine (running an up-to-date Arch Linux) and I haven’t encountered any issue so far.


birkenfeld:
I'm not sure that the Kconfig lexer is the right choice for "conf" on Linux. It's a very specific format for configuring the kernel build system, not generic "Linux config files".

PR #1326, opened 2019-12-04
757/germanriano/mosel

jcmuel:
I deleted the test file that had an incompatible license and created a new test file. Moreover, I also added several keywords and types.


pygments/lexers/mosel.py
```diff
@@ -27,123 +27,400 @@ class MoselLexer(RegexLexer):
...
-            (r'(integer|is_binary|real)', Keyword.Type),
-            (r'(\+|\-|\*|/|=|<=|>=|==|\||\^|<|>|<>|\.\.|:=|:)', Operator),
+            (words((
+                'range', 'array', 'set', 'list', 'mpvar', 'mpproblem', 'linctr', 'nlctr', 'integer',
+                'string', 'real', 'boolean', 'text', 'time', 'date', 'datetime', 'returned', 'Model',
+                'Mosel', 'counter', 'xmldoc',
+                'is_sos1', 'is_sos2', 'is_integer', 'is_binary', 'is_continuous',
+                'is_free', 'is_semcont', 'is_semint', 'is_partint'
+                )), Keyword.Type),
+            (r'(\+|\-|\*|/|=|<=|>=|==|\||\^|<|>|<>|\.\.|\.|:=|:|::|in|mod|div)', Operator),
```
birkenfeld: `:` and `::` need to swap places to have `::` recognized.

pygments/lexers/mosel.py
```diff
@@ -27,123 +27,400 @@ class MoselLexer(RegexLexer):
...
             (r'[()\[\]{},;:]+', Punctuation),
```
birkenfeld: Is `:` still needed here as punctuation?

jcmuel: You are right. It's not needed. I just removed it.

pygments/lexers/mosel.py
jcmuel: Thanks for the hint. I just reordered the `:` and `::`.

birkenfeld:
Thanks!

Issue #1327, opened 2019-12-04
2.5.1 MatlabLexer regression


florczakraf:
https://github.com/pygments/pygments/pull/1271 introduced a regression to `MatlabLexer`.

### Reproduction script:
```
import pygments
from pygments.lexers.matlab import MatlabLexer

print(list(pygments.lex('foo bar', MatlabLexer())))
print(list(pygments.lex('foo', MatlabLexer())))
```

### Before (2.5.1 with 612fb2b reverted):
```
[(Token.Name, 'foo'), (Token.Text, ' '), (Token.Name, 'bar'), (Token.Text, '\n')]
[(Token.Name, 'foo'), (Token.Text, '\n')]
```

### After:
There are invalid characters (` `, `\n`) in `Name`. ~~Moreover, `bar` is of a wrong type now.~~
```
[(Token.Name, 'foo '), (Token.Literal.String, 'bar'), (Token.Punctuation, '\n')]
[(Token.Name, 'foo\n'), (Token.Punctuation, '')]
```

@birkenfeld @anntzer could you please have another look at those changes?

birkenfeld:
I believe the change in `bar` was intended by @anntzer, since this is "command syntax" with arguments.

The `Name` should be fixed to not include the newline, of course.

anntzer:
"bar" is effectively a string from MATLAB's POV (`foo bar` is equivalent to `foo('bar')`) and I would say should be highlighted as such.
I guess the invalid characters can be fixed by making the \s+ in https://github.com/pygments/pygments/commit/612fb2b33d6f48965661e53c6a88ec78b5dd54e3#diff-bc1be5d33c811ece546521ae70944206R104 part of the lookahead and by removing the newline from https://github.com/pygments/pygments/commit/612fb2b33d6f48965661e53c6a88ec78b5dd54e3#diff-bc1be5d33c811ece546521ae70944206R146?

birkenfeld:
SGTM.

florczakraf:
> "bar" is effectively a string from MATLAB's POV (foo bar is equivalent to foo('bar')) and I would say should be highlighted as such.

You are absolutely right here! I'm still a little bit confused with matlab's calling conventions...

anntzer:
[...]

PR #1328, opened 2019-12-05
More fixes to MATLAB lexer.

anntzer:
Per discussion at #1327.

### DIFF
--- a/pygments/lexers/matlab.py
+++ b/pygments/lexers/matlab.py
@@ -101,8 +101,8 @@ class MatlabLexer(RegexLexer):
             # is recognized if it is either surrounded by spaces or by no
             # spaces on both sides; only the former case matters for us.  (This
             # allows distinguishing `cd ./foo` from `cd ./ foo`.)
-            (r'(?:^|(?<=;))\s*\w+\s+(?!=|\(|(%s)\s+)' % _operators, Name,
-             'commandargs'),
+            (r'(?:^|(?<=;))(\s*)(\w+)(\s+)(?!=|\(|(%s)\s+)' % _operators,
+             bygroups(Text, Name, Text), 'commandargs'),
 
             # operators:
             (_operators, Operator),
@@ -142,9 +142,10 @@ class MatlabLexer(RegexLexer):
             (r"[^']*'", String, '#pop'),
         ],
         'commandargs': [
+            (r"[ \t]+", Text),
             ("'[^']*'", String),
-            ("[^';\n]+", String),
-            (";?\n?", Punctuation, '#pop'),
+            (r"[^';\s]+", String),
+            (";?", Punctuation, '#pop'),
         ]
     }
 


pygments/lexers/matlab.py
```diff
@@ -142,9 +142,10 @@ class MatlabLexer(RegexLexer):
             (r"[^']*'", String, '#pop'),
         ],
         'commandargs': [
+            (r"\s+", Text),
             ("'[^']*'", String),
             ("[^';\n]+", String),
-            (";?\n?", Punctuation, '#pop'),
+            (";?(?=\n)?", Punctuation, '#pop'),
```
birkenfeld: I think this doesn't do anything, since the lookahead is optional.

I guess use `bygroups()` instead to have the `\n` as Text.

pygments/lexers/matlab.py
```diff
@@ -101,8 +101,8 @@ class MatlabLexer(RegexLexer):
             # is recognized if it is either surrounded by spaces or by no
             # spaces on both sides; only the former case matters for us.  (This
             # allows distinguishing `cd ./foo` from `cd ./ foo`.)
-            (r'(?:^|(?<=;))\s*\w+\s+(?!=|\(|(%s)\s+)' % _operators, Name,
-             'commandargs'),
+            (r'(?:^|(?<=;))(\s*)(\w+)(?=\s)(?!\s+(=|\(|(%s)\s+))' % _operators,
+             bygroups(Text, Name), 'commandargs'),
```
birkenfeld: The whitespace that's now in the lookahead could also be in its own Text group with bygroups().

pygments/lexers/matlab.py
anntzer: With the lookahead the \n will never be labeled as Punctuation, but a single \n will correctly #pop the state (without being consumed).

pygments/lexers/matlab.py
anntzer: I can just let a later match handle that?

birkenfeld: Yes, sure. I just don't like to have lookaheads or lookbehinds when normal matching does the same job; they complicate the regex needlessly.

pygments/lexers/matlab.py
birkenfeld: But since the lookahead is optional, I don't see what the difference is vs. removing it completely.

anntzer: Because otherwise you stay in commandargs mode even after the newline.

pygments/lexers/matlab.py
anntzer:
```patch
-            (r'(?:^|(?<=;))(\s*)(\w+)(?=\s)(?!\s+(=|\(|(%s)\s+))' % _operators,
-             bygroups(Text, Name), 'commandargs'),
+            (r'(?:^|(?<=;))(\s*)(\w+)(\s+)(?!=|\(|(%s)\s+)' % _operators,
+             bygroups(Text, Name, Text), 'commandargs'),
```
works for you?

birkenfeld: That's what I meant, yes!

pygments/lexers/matlab.py
birkenfeld: Again, I claim that you can remove `(?=\n)?` and there is no observable difference. I may be missing the forest for the trees, so do you have an example that shows otherwise?

pygments/lexers/matlab.py
anntzer: pushed

pygments/lexers/matlab.py
anntzer: A bad case is
```
In [1]: print(list(pygments.lex('foo bar baz\nquux', MatlabLexer())))
[(Token.Name, 'foo'), (Token.Text, ' '), (Token.Literal.String, 'bar baz'), (Token.Text, '\n'), (Token.Literal.String, 'quux'), (Token.Text, '\n'), (Token.Punctuation, '')]
```
note how the `quux` on the second line is also interpreted as a Literal.String rather than a Name.
...
except that I realized that even with the (?=\n)? this still fails because the \s+ coming first globs everything; and there's another bug (`bar baz` is two strings).

I think
```
        'commandargs': [
            (r"[ \t]+", Text),
            ("'[^']*'", String),
            (r"[^';\s]+", String),
            (";?", Punctuation, '#pop'),
        ]
```
may work better?

birkenfeld: Yes, that should leave commandargs mode after anything that isn't space, string or word.

anntzer: pushed

PR #1332, opened 2019-12-06
Improve docs.

Anteru:
Add a contributing guide for newcomers, and remove references to pygmentize in the development docs. During development, there is no pygmentize.

### DIFF
--- a/Contributing.md
+++ b/Contributing.md
@@ -1,0 +1,44 @@
+Licensing
+=========
+
+The code is distributed under the BSD 2-clause license. Contributors making pull
+requests must agree that they are able and willing to put their contributions
+under that license.
+
+Contribution checklist
+======================
+
+* Check the documentation for how to write
+  [a new lexer](https://pygments.org/docs/lexerdevelopment/),
+  [a new formatter](https://pygments.org/docs/formatterdevelopment/) or
+  [a new filter](https://pygments.org/docs/filterdevelopment/)
+* When writing rules, try to merge simple rules. For instance, combine:
+
+  ```python
+  _PUNCTUATION = [
+    (r"\(", token.Punctuation),
+    (r"\)", token.Punctuation),
+    (r"\[", token.Punctuation),
+    (r"\]", token.Punctuation),
+    ("{", token.Punctuation),
+    ("}", token.Punctuation),
+  ]
+  ```
+
+  into:
+
+  ```python
+  (r"[\(\)\[\]{}]", token.Punctuation)
+  ```
+* Be careful with ``.*``. This matches greedily as much as it can. For instance,
+  rule like ``@.*@`` will match the whole string ``@first@ second @third@``,
+  instead of matching ``@first@`` and ``@second@``. You can use ``@.*?@`` in
+  this case to stop early. The ``?`` tries to match _as few times_ as possible.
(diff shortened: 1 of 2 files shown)


Contributing.md
```diff
@@ -0,0 +1,31 @@
+Licensing
+=========
+
+The code is distributed under the BSD 2-clause license. Contributors making pull requests must agree that they are able and willing to put their contributions under that license.
```
birkenfeld: I'd prefer to keep lines broken at ~80 chars also in md/rst files.

Contributing.md
```diff
@@ -0,0 +1,31 @@
...
+  ]
+  ```
+
+  into:
+
+  ```python
+  (r"[\(\)\[\]{}]", token.Punctuation)
+  ```
+* Be careful with ``.*``. This matches greedily as much as it can. For instance, a rule like ``@.*@`` will match the whole string ``@first@ second @third@``, instead of matching ``@first@`` and ``@second@``. You can use ``@.*?@`` in this case to stop early. The ``?`` tries to match _as few times_ as possible.
+* Don't add your new lexer to ``compiled.py`` -- this file is there for backwards compatibility.
```
birkenfeld: this can be expanded into "don't add imports of your new lexer anywhere".

Contributing.md
```diff
@@ -0,0 +1,31 @@
...
+* Use the standard importing convention: ``from token import Punctuation``
+* If you have a tricky case, you can use the ``testcase`` formatter to produce an unit test quickly. Run ``python -m pygments -l lua -f testcase <<< "local a = 5"``. This will produce a test case function skeleton.
```
birkenfeld: nice!

doc/docs/lexerdevelopment.rst
```diff
@@ -164,16 +164,16 @@ To test the new lexer, store an example file with the proper extension in
...
 
-    $ ./pygmentize -O full -f html -o /tmp/example.html tests/examplefiles/example.diff
+    $ python -m pygments -O full -f html -o /tmp/example.html tests/examplefiles/example.diff
 
-Note that this explicitly calls the ``pygmentize`` in the current directory
-by preceding it with ``./``. This ensures your modifications are used.
-Otherwise a possibly already installed, unmodified version without your new
-lexer would have been called from the system search path (``$PATH``).
+Note that this explicitly calls the ``pygments`` module in the current
+directory. This ensures your modifications are used. Otherwise a possibly
```
birkenfeld: Now that this isn't bound to ``./pygmentize`` existing, it's better to state that this must be run from the root of the checkout.  The "Otherwise..." is still useful in this case.

Contributing.md
Anteru: Done!

Contributing.md
Anteru: Rephrased, hope it's clear now.

Contributing.md
Anteru: :+1:

doc/docs/lexerdevelopment.rst
Anteru: Improved!

birkenfeld:
New commit seems to be the same as the old?

Anteru:
Indeed, because I forgot to add the changed files :)

birkenfeld:
Nice!

PR #1336, opened 2019-12-11
Add a PEG (Parsing Expression Grammar) lexer

goodmami:
[Parsing Expression Grammars](https://en.wikipedia.org/wiki/Parsing_expression_grammar) (PEGs) have been around for about 15 years and they are fairly popular these days for describing non-ambiguous grammars, such as for programming languages and data formats.

This lexer (`pygments.lexers.grammar_notation.PegLexer`) builds on the original PEG syntax described by Bryan Ford (see [here](https://bford.info/pub/lang/peg.pdf)) with some common extensions seen in various implementations of PEG, such as (optionally) using `|` for choices instead of `/`, `=` or `:` instead of `<-` for rule definitions, cut operators (`^` or `~`), and string modifiers, such as `r"a regex"`.

I've added `tests/test_grammar_notation.py`, but only included tests relevant to this PR instead of going back and adding tests for `BnfLexer`, etc.

### DIFF
--- a/pygments/lexers/grammar_notation.py
+++ b/pygments/lexers/grammar_notation.py
@@ -230,6 +230,8 @@ class PegLexer(RegexLexer):
     * A single `a-z` character immediately before a string, or
       multiple `a-z` characters following a string, are part of the
       string (e.g., `r"..."` or `"..."ilmsuxa`).
+
+    .. versionadded:: 2.6
     """
 
     name = 'PEG'
--- a/pygments/lexers/grammar_notation.py
+++ b/pygments/lexers/grammar_notation.py
@@ -15,7 +15,7 @@ from pygments.lexer import RegexLexer, bygroups, include, this, using, words
 from pygments.token import Comment, Keyword, Literal, Name, Number, \
     Operator, Punctuation, String, Text
 
-__all__ = ['BnfLexer', 'AbnfLexer', 'JsgfLexer']
+__all__ = ['BnfLexer', 'AbnfLexer', 'JsgfLexer', 'PegLexer']
 
 
 class BnfLexer(RegexLexer):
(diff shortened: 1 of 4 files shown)


pygments/lexers/grammar_notation.py
```diff
@@ -211,3 +211,58 @@ class JsgfLexer(RegexLexer):
...
+    * `<-`, `←`, `:`, and `=` are all accepted as rule operators.
+
+    * Both `|` and `/` are choice operators.
+
+    * `^`, `↑`, and `~` are cut operators.
+
+    * A single `a-z` character immediately before a string, or
+      multiple `a-z` characters following a string, are part of the
+      string (e.g., `r"..."` or `"..."ilmsuxa`).
+    """
```
Anteru: Can you please add ``.. versionadded:: 2.6`` at the end?

goodmami:
Added the `versionadded` tag and added the language to `doc/languages.rst`, as requested.

There are two remaining things about the lexer that I thought about and I'll write them here for posterity:
* At least one syntax variant (used in [Parsimonious](https://github.com/erikrose/parsimonious)) uses `~"..."`, `~r"..."`, etc. for regular expressions, as an extension. This was not compatible with `~` as a cut operator (which is used in the new [pegen](https://github.com/gvanrossum/pegen) and older, but not exactly PEG, [TatSu](https://github.com/neogeny/TatSu) libraries), but the impact is minimal: it will just be highlighted differently and does not break parsing.
* I did not account for extensions that put semantic actions between `{` and `}` at the end of the rule (as in the previously mentioned pegen, as well as [Rats!](https://cs.nyu.edu/rgrimm/xtc/rats-intro.html) and some other parsers. I chose not to do this because I didn't see such support in the other grammar notation lexers, but I can add it if requested.

goodmami:
@Anteru do you wish me to resolve the conflicts in `pygments/lexers/_mapping.py` or does the person merging handle this? And is there anything else you're waiting on me for?

Anteru:
Hi, that would be appreciated, but I regenerate the `_mapping.py` file quite often myself already. I just didn't get around to merging this yet, will try to do it this week.

goodmami:
Ok I've rebased to the current master branch. Let me know if you need anything else.

Anteru:

PR #1337, opened 2019-12-16
Update the Tera Term macro lexer (new 'setspeed' command)

kurtmckee:
I've added the `setspeed` command to the Tera Term macro lexer.

In addition, I removed the `analyse_text()` math and just return 0.01 or 0.0 directly. I've also expanded and alphabetized the regex strings so that it's easier to update the macro commands and variables if any additional commands are added in the future.

Please let me know if I've overlooked anything or need to make changes to this update.


kurtmckee:
Haha, neat! I got PR number '1337'. 😃

Anteru:
The default for `analyse_text` is to return 0 (well, `None`, and that gets converted to 0 ...). Unless there is a warning that it's missing due to ambiguity, I'd remove it.

Issue #1340, opened 2019-12-19
Python lexer problem with fstrings


nivaca:
The Python lexer doesn't always detect fstrings correctly.
For instance, the code:
```python
name = "Ann"
print(f"{name}, you have {len(name)} characters in your name.")
```
is rendered as: 
![Screenshot_20191218_205714](https://user-images.githubusercontent.com/2309005/71138915-e3ffca00-21da-11ea-9b30-619e9867b746.png)

Note that while the first variable (`{name}`) seems to be correctly understood, the second one (`{len(name}`) isn't. Either way, they are rendered differently.

I'm using v.2.5.2 of Pygments (but through the pygmentize CLI, as well as through the `minted` LaTeX package).

Prometheus3375:
The same problem occurs with formatting strings as well. For example, in [Python docs](https://docs.python.org/3/library/string.html#format-examples):

![Image](https://github.com/user-attachments/assets/50589f9e-ec78-416e-9a9d-3081d5a65974)
![Image](https://github.com/user-attachments/assets/a448c543-c339-4da3-861e-6c9b5dd37c34)
![Image](https://github.com/user-attachments/assets/6841b65b-0e5b-4c89-87c5-8f63511d1917)

PR #1342, opened 2019-12-20
change in regex to allow for parentheses, and so functions, within f-…

nivaca:
change in regex to allow for parentheses, and so functions, within f-strings

See Issue #1340


CrafterKolyan:
That's actually don't fix the problem. Example given:
```
print(f"{name}, you have {'some_string with \n character'} characters in your name.")
```
Can we do it so any valid python code would be parsed?

Anteru:
Can you please add an unit test for this?

birkenfeld:
This change seems to have gotten lost?

birkenfeld:
Ah, I see it was superseded by a more comprehensive f-string overhaul.

PR #1343, opened 2019-12-21
Add nowrap option to LatexFormatter

tk0miya:
At present, LatexFormatter always output `\begin{Verbatim}` macro. As a result, the generated content always becomes block level element. To support to generate inline one, I'd like to add `nowrap` option to LatexFormatter like HtmlFormatter.


Jean-Abou-Samra (commented):
Sounds good, could you rebase this on top of master?

tk0miya:
Updated!

PR #1347, opened 2019-12-29
haskell: Fix highlighting of promoted type operators

### LINKED ISSUE #527: Haskell: highlighting promoted types
(Original issue 820 created by None on 2012-11-09T07:29:22.158157+00:00)

GHC DataKinds extension introduces new syntax for data constructors promoted to type level. A type name can now include a single quote in the beginning, see: http://www.haskell.org/ghc/docs/latest/html/users_guide/promotion.html#promotion-syntax
Also, GHC TypeFamilies extension introduces "family" keyword which isn't highlighted yet.
example: http://pygments.org/demo/58336/

Lysxia:
Closes #527 

Are there tests I'm supposed to run/update after changing the examples?


Anteru:
There are none for Haskell currently. You could add a new unit test file for it, see https://github.com/pygments/pygments/blob/master/Contributing.md for how to create a testcase. That would be highly appreciated :)

Lysxia:
Thanks for the tip! Tell me if anything is missing.

Anteru:
Perfect, thanks a lot!

PR #1349, opened 2020-01-05
Adds support for C++20 keywords

HubertGruniaux:
The soon C++20 newcomer standard will introduce lots of new keywords like `constinit`, `co_yield` (for courotines), `import` or`module` (for modules, however these are special identifiers instead of real keywords), etc...


Anteru:
Thanks, queued for the next release!

PR #1360, opened 2020-01-16
Lacking keywords were added to the Idris lexer

buzden:


Anteru:

PR #1361, opened 2020-01-16
Add lexer for LLVM's MIR format

dsandersllvm:
MIR is a human readable serialization format that's used to represent LLVM's
machine specific intermediate representation. It allows LLVM's developers to
see the state of the compilation process at various points, as well as test
individual pieces of the compiler. Our documentation for the format can be
found at https://llvm.org/docs/MIRLangRef.html.

Adding a lexer for this format will allow the LLVM documentation to contain
syntax highlighted examples of LLVM-MIR. Two lexers are included in this
change. 'llvm-mir' lexes the overall document format and delegates to 'llvm' and
'llvm-mir-body' as appropriate. 'llvm-mir-body' lexes the contents of the 'body:'
attribute and can be used directly to syntax highlight code examples without
including the document boilerplate.

Since the 'llvm-mir' lexer delegates to the 'llvm' lexer at times, this change
also adds the 'immarg' and 'willreturn' keywords to the 'llvm' lexer as these
were missing.

### DIFF
--- a/pygments/lexers/asm.py
+++ b/pygments/lexers/asm.py
@@ -12,15 +12,16 @@
 import re
 
 from pygments.lexer import RegexLexer, include, bygroups, using, words, \
-    DelegatingLexer
+    DelegatingLexer, default
 from pygments.lexers.c_cpp import CppLexer, CLexer
 from pygments.lexers.d import DLexer
 from pygments.token import Text, Name, Number, String, Comment, Punctuation, \
-    Other, Keyword, Operator
+    Other, Keyword, Operator, Literal
 
 __all__ = ['GasLexer', 'ObjdumpLexer', 'DObjdumpLexer', 'CppObjdumpLexer',
-           'CObjdumpLexer', 'HsailLexer', 'LlvmLexer', 'NasmLexer',
-           'NasmObjdumpLexer', 'TasmLexer', 'Ca65Lexer', 'Dasm16Lexer']
+           'CObjdumpLexer', 'HsailLexer', 'LlvmLexer', 'LlvmMirBodyLexer',
+           'LlvmMirLexer', 'NasmLexer', 'NasmObjdumpLexer', 'TasmLexer',
+           'Ca65Lexer', 'Dasm16Lexer']
 
 
 class GasLexer(RegexLexer):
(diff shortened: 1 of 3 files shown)


pygments/lexers/asm.py
```diff
@@ -452,6 +453,201 @@ class LlvmLexer(RegexLexer):
...
+            (r'\), \(', Text),
+            (r'\)', Text, '#pop'),
+        ],
+        'integer': [(r'-?[0-9]+', Number.Integer),],
+        'float': [(r'-?[0-9]+\.[0-9]+(e[+-][0-9]+)?', Number.Float)],
+        'global': [(r'\@[0-9a-zA-Z_.]+', Name.Variable.Global)],
+    }
+
+class LlvmMirLexer(RegexLexer):
+    name = 'LLVM-MIR'
```
Anteru: Can you please add a short comment here about this lexer, ideally with a link to the language specification? The comment should also include a ``.. versionadded:: 2.6``.

pygments/lexers/asm.py
```diff
@@ -452,6 +453,201 @@ class LlvmLexer(RegexLexer):
         ]
     }
 
+class LlvmMirBodyLexer(RegexLexer):
+    """
+    For LLVM MIR examples without the YAML wrapper
+    """
```
Anteru: Missing ``.. versionadded:: 2.6``

dsandersllvm: Added it

pygments/lexers/asm.py
dsandersllvm: I've added the first paragraph of the commit message to the docstring with a few small modifications and added the `.. versionadded:: 2.6`

dsandersllvm:
Hmm, the test environment seems to be a bit fussier than pygmentize. I've reproduced those test failures locally and I'll investigate

dsandersllvm:
> Hmm, the test environment seems to be a bit fussier than pygmentize. I've reproduced those test failures locally and I'll investigate

The error token was easy enough to fix by adding:
`            (r'[-+]', Operator),`
to the 'mir' state. However, it then failed the round trip test. It turned out I'd managed to use `bygroups(using(...))` in a way that resulted in the same newline character generating multiple tokens. The round trip test would then fail on the doubled newlines. I've modified both patterns that use `bygroups(using(...))` to fix this

Anteru:
Thanks for adding the tests -- and the fixes. I'll give it a spin and will merge it then.

dsandersllvm:
Thanks

Anteru:

PR #1362, opened 2020-01-17
D: remove typedef from keywords

kubo39:
ref: https://dlang.org/deprecate.html#typedef

### DIFF
--- a/pygments/lexers/d.py
+++ b/pygments/lexers/d.py
@@ -47,7 +47,7 @@ class DLexer(RegexLexer):
                 'module', 'new', 'nothrow', 'out', 'override', 'package', 'pragma',
                 'private', 'protected', 'public', 'pure', 'ref', 'return', 'scope',
                 'shared', 'static', 'struct', 'super', 'switch', 'synchronized',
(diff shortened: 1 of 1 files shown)


Anteru:
Hi, can you please elaborate a bit? It's still a valid keyword, just deprecated, should we really stop highlighting it or maybe move it to a new class of deprecated keyword or something? This way you could highlight it in a different way if really needed, but all existing code would continue to get highlighted correctly.

kubo39:
Hi! `typdef` keyword was completely removed in dmd 2.072, please check the top of [this url].(https://dlang.org/deprecate.html)(I agree that the page title is confusing.)

Anteru:
I see. The reason I'm not super comfortable with removing it wholesale is that old code can use it, and it should remain highlighted correctly. That's why I think using a new class like `Keyword.Deprecated` and `Keyword.Removed` is better than just removing it. Any thoughts?

kubo39:
Looks good, should I close this issue?

Anteru:
No, I'll close it once I fix it :)

PR #1363, opened 2020-01-19
Correct lexing of Idris compiler directives

sysvinit:
See commit message.


Anteru:
Nice catch! Would you mind adding a test-case for this? There's a unit-test-case formatter you can use for that.

sysvinit:
I've pushed some tests in 2d4bb3d, which use the code fragments I previously saw failures with before the regex fix.

Anteru:
Merged, thanks a lot for the contribution and the quick update!

PR #1374, opened 2020-01-29
image formatter supports background color

hmedina:
Extend Image formatters in img.py to support background colors.

* New function to get the background color for a given token type, optionally None
* The self.drawables list used the fill keyword, stuffed into a **kwrgs, for the color of the token; since tokens can have foreground and background colors, that list now uses two explicit names, passing one to the PIL/Pillow text draw call, the other to the rectangle draw call.
* When iterating over the `drawables` list, the presence of a background color triggers the drawing of a rectangle beneath the text.
* Dimensions of background rectangle calculated by PIL based on the rendered text and font.
* If input string is multiline, PIL will use multiline for draw and size internally.

### DIFF
--- a/pygments/formatters/img.py
+++ b/pygments/formatters/img.py
@@ -434,6 +434,16 @@ class ImageFormatter(Formatter):
             fill = '#000'
         return fill
 
+    def _get_text_bg_color(self, style):
+        """
+        Get the correct background color for the token from the style.
+        """
+        if style['bgcolor'] is not None:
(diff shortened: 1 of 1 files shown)


Anteru:
Sorry for taking this long to get around to this :( One question on the `**kw` replacement: Did you audit all places to make sure there is nothing else passed via `**kw`? Otherwise this looks good.

birkenfeld (approved):
LGTM.

Anteru:
Merged, thanks!

PR #1378, opened 2020-02-01
Fix glob patterns when absolute paths are used

catatonicprime:
When lexing an apache config the following example illustrates the issue of lexing a glob pattern which is also an absolute path:
```
Include /etc/httpd/conf.d/*.conf
```

This lexs to:
```
(Token.Name.Builtin, 'Include')
(Token.Text, ' ')
(Token.Literal.String.Other, '/etc/httpd/conf.d/')
(Token.Text, '*.conf')
(Token.Text, '')
(Token.Text, '\n')
```

The ```(Token.Text, '*.conf')``` seems like it should really be part of the String.Other, like so ```(Token.Literal.String.Other, '/etc/httpd/conf.d/*.conf')```

This PR adds the '*' to the acceptable character list for these type of absolute file path/glob patterns. Tests have been included to address:
- one original absolute file path from the examplefiles/apache2.conf
- one basic glob pattern
- one multi-glob pattern
- one root level multi-glob pattern

### DIFF
--- a/pygments/lexers/configs.py
+++ b/pygments/lexers/configs.py
@@ -313,7 +313,7 @@ class ApacheConfLexer(RegexLexer):
             (r'[^\S\n]+', Text),
             (r'\d+\.\d+\.\d+\.\d+(?:/\d+)?', Number),
             (r'\d+', Number),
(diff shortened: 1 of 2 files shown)


Anteru:

PR #1382, opened 2020-02-03
Adds the walrus operator to the Python3 lexer.

### LINKED ISSUE #1381: Python 3.8 assignment expressions not treated as an operator
The `:=` operator was added into Python 3.8:

https://www.python.org/dev/peps/pep-0572/

The Pygments `PythonLexer` hasn't been updated to match:

https://github.com/pygments/pygments/blob/0cea7133313766e26edad46a2377eacbf7cbe459/pygments/lexers/python.py#L102

bslatkin:
- The walrus operator, also known as assignment expressions, was introduced in Python 3.8
- Moves the Token.Operator matching root above Token.Punctuation so the walrus operator takes precedence
- Includes a test to make sure this behavior doesn't regress since it's sensitive to the order of expressions
- Fixes #1381


birkenfeld:
Nice, thanks!

PR #1384, opened 2020-02-04
Python f-strings: highlight expressions in curly braces

### LINKED ISSUE #1228: f-strings: expressions in curly braces
(Original issue 1524 created by udioron on 2019-06-09T20:29:23.059458+00:00)

f-strings highlighting does not currently display expressions in curly braces correctly:

```python
s1 = "{1} {1+1} {a} {a*b}"
s2 = f"{1} {1+1} {a} {a*b}"
```

See: [http://pygments.org/demo/6837038/](http://pygments.org/demo/6837038) 

For `s1` \(regular string literal\) `{1}` and `{a}` should be in a different color.

For `s2` all four curly bracket expressions should be colored.

This is somewhat tricky, as the following expressions are allowed:

```python
f"{'}' * 10}"
```

However, if the expression inside the curly brackets can be completely syntax highlighted as well, as any python expression, this will be amazing!

tomerv:
Fixes #1228

### DIFF
--- a/pygments/lexers/python.py
+++ b/pygments/lexers/python.py
@@ -85,6 +85,20 @@ class PythonLexer(RegexLexer):
             # newlines are an error (use "nl" state)
         ]
 
+    def fstring_rules(ttype):
+        return [
+            # Assuming that a '}' is the closing brace after format specifier.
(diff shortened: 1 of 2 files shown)


tomerv:
I'll rebase a send another request

Issue #1355, opened 2020-01-11
request for reasonml(buckle script) language support.


geohuz:
Would like to have reasonml(buckle script) language syntax highlight.

PR #1386, opened 2020-02-05
Add ReasonML lexer.

fhammerschmidt:
This implements #1355 

I derived the ReasonML lexer from the OCaml one, since the former language is based on the latter, which turned out relatively good.

I added a testfile "test.re", too if you want to see for yourself.

Thanks in advance!

### DIFF
--- a/pygments/lexers/ml.py
+++ b/pygments/lexers/ml.py
@@ -15,7 +15,7 @@ from pygments.lexer import RegexLexer, include, bygroups, default, words
 from pygments.token import Text, Comment, Operator, Keyword, Name, String, \
     Number, Punctuation, Error
 
-__all__ = ['SMLLexer', 'OcamlLexer', 'OpaLexer']
+__all__ = ['SMLLexer', 'OcamlLexer', 'OpaLexer', 'ReasonLexer']
 
 
 class SMLLexer(RegexLexer):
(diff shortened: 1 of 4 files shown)


pygments/lexers/ml.py
```diff
@@ -767,3 +766,94 @@ class OpaLexer(RegexLexer):
             (r'[^\-]+|-', Comment),
         ],
     }
+
+class ReasonLexer(RegexLexer):
+    """
+    For the ReasonML language.
+
+    .. versionadded:: 2.5
```
Anteru: Please change to 2.6, the 2.5 ship has sailed.

pygments/lexers/ml.py
```diff
@@ -767,3 +766,94 @@ class OpaLexer(RegexLexer):
             (r'[^\-]+|-', Comment),
         ],
     }
+
+class ReasonLexer(RegexLexer):
+    """
+    For the ReasonML language.
```
Anteru: Could you please add a link to the language here?

jordwalke:
Thanks @fhammerschmidt !

montogeek:
@fhammerschmidt Do you have a plan to do the same for ReScript?

fhammerschmidt:
> @fhammerschmidt Do you have a plan to do the same for ReScript?
Yeah, I could do that. 

But it may be wiser to do it when more of the syntax has settled (like local opens, Infix, module imports).

PR #1388, opened 2020-02-08
Multiline flag does not work with re.match

### LINKED ISSUE #1056: re.match causes re.MULTILINE to be a noop
(Original issue 1352 created by alanhamlett on 2017-06-08T06:43:07.775020+00:00)

A lot of lexers use `re.match` with `re.M`/`re.MULTILINE`, but `re.M` doesn't actually match start of lines unless used with `re.search`.

This is causing poor Lexer accuracy and probably some edge-case bugs.

alanhamlett:
Replace `re.match` with `re.search` whenever the `re.MULTILINE` / `re.M` flag is present. Fixes #1056.

### DIFF
--- a/pygments/lexers/asm.py
+++ b/pygments/lexers/asm.py
@@ -101,9 +101,9 @@ class GasLexer(RegexLexer):
     }
 
     def analyse_text(text):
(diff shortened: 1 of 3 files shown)


Anteru:
Merged, thanks -- nice find!

Issue #321, opened 2019-08-31
inherance (of bold, italic, etc) not working as expected with LaTeX


Anteru:
(Original issue 614 created by None on 2010-11-16T21:00:59.827635+00:00)

Hi, I'm using Pygmentize for LaTeX output but I've got an annoying bug, I can't get italic comments while getting non italic preprocessor directives like a #include, while it works on HTML output. The problem lies in some wrong handling of style inherance  for LaTeX.
For example, in murphy style we have:

Comment:                   "#666 italic",
Comment.Preproc:           "#579 noitalic",
Comment.Special:           "#c00 bold",

and, even though we have noitalic for Preproc, it keeps italic in LaTeX output (I'm using pdfLatex on TeX Live 2010) while it outputs non italic on html.

Pygmentize is great thanks.

Anteru:
(Original comment by tshatch on 2010-12-21T08:35:00.132796+00:00)

Hi, can you provide full steps to reproduce?

Anteru:
(Original comment by None on 2010-12-21T12:36:18.155821+00:00)

Create a directory and inside it create a file named **main.cpp** with the following contents:
{{{

#include <iostream>

int main()
{
	// Saying hello to the world
	
	std::cout << "Hello World!" << std::endl;
	
	return 0;
}

}}}

Navigate to that directory through the command shell and execute the command

**pygmentize -O full,style=murphy -l c++ -f html -o main.html main.cpp**

then execute

**pygmentize -O full,style=murphy -l c++ -f latex -o main.tex main.cpp**

now execute

**pdflatex main.tex**

Now opening both pdf and html output you may see that the #include in the html file isn't italic, which is expected since the murphy style used here doens't inherent italics for the directives (directives are view as subsections of comments in styles), but it keeps italic in the pdf file, while in both, the comments are italic as expected.

Regards

Anteru:
(Original comment by None on 2011-03-02T16:41:29.829977+00:00)

Hello?

Anteru:
(Original comment by None on 2011-07-22T20:01:21.271185+00:00)

The issue is discussed here http://pbelmans.wordpress.com/2011/07/05/guest-blog-inheritance-in-pygments-latexformatter/ and a solution is presented.

Here's a patch against current pygments tip:
{{{
--- pygments/formatters/latex.py.orig	2011-07-22 22:55:29.000000000 +0300
[...]

PR #1391, opened 2020-02-11
Updated LaTeX formatter to explicitly define style

tabarnakos:
There are 6 macros that define the style of the text: \PY@it, \PY@bf, \PY@ul, \PY@tc, \PY@bc and \PY@ff.

The proposed change explicitly sets the style for each token, fixing #321

### DIFF
--- a/pygments/formatters/latex.py
+++ b/pygments/formatters/latex.py
@@ -287,19 +287,29 @@ class LatexFormatter(Formatter):
             cmndef = ''
             if ndef['bold']:
                 cmndef += r'\let\$$@bf=\textbf'
+            else:
+                cmndef += r'\let\$$@bf=\relax '
             if ndef['italic']:
                 cmndef += r'\let\$$@it=\textit'
+            else:
+                cmndef += r'\let\$$@it=\relax '
             if ndef['underline']:
                 cmndef += r'\let\$$@ul=\underline'
+            else:
+                cmndef += r'\let\$$@ul=\relax '
             if ndef['roman']:
                 cmndef += r'\let\$$@ff=\textrm'
-            if ndef['sans']:
+            elif ndef['sans']:
                 cmndef += r'\let\$$@ff=\textsf'
-            if ndef['mono']:
+            elif ndef['mono']:
                 cmndef += r'\let\$$@ff=\textsf'
+            else:
+                cmndef += r'\let\$$@ff=\relax '
             if ndef['color']:
                 cmndef += (r'\def\$$@tc##1{\textcolor[rgb]{%s}{##1}}' %
                            rgbcolor(ndef['color']))
+            else:
+                cmndef += r'\let\$$@tc=\relax '
             if ndef['border']:
                 cmndef += (r'\def\$$@bc##1{\setlength{\fboxsep}{0pt}'
                            r'\fcolorbox[rgb]{%s}{%s}{\strut ##1}}' %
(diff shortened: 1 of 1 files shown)


pygments/formatters/latex.py
```diff
@@ -287,19 +287,29 @@ def rgbcolor(col):
...
+            else:
+                cmndef += r'\let\$$@ul=\relax '
             if ndef['roman']:
                 cmndef += r'\let\$$@ff=\textrm'
-            if ndef['sans']:
+            elif ndef['sans']:
                 cmndef += r'\let\$$@ff=\textsf'
-            if ndef['mono']:
+            elif ndef['mono']:
                 cmndef += r'\let\$$@ff=\textsf'
```
Anteru: If this is mono, should this be `\texttt`?

tabarnakos: I didn't touch that piece of the code, simply changed if for elif because the font can't be both serif and sans-serif at the same time.



It would indeed make sense (I saw another issue on this), but I am not sure how this option is used. In my case the output is always monospaced. Is this a legacy option no longer in use?

Anteru: I have no idea to be honest, I haven't tried the LaTeX formatter myself :) It just looked wrong to me. Maybe @birkenfeld remembers how this came to be?

CoelacanthusHex:
Is it still WIP?

birkenfeld:
@jfbu  do you have time for a review?

pygments/formatters/latex.py
jfbu: I also wonder why the `'mono'` does not trigger `\texttt`.

birkenfeld: Must be a copy-paste error from `sans` above. But I don't see `mono` being supported anywhere else, so it would be best to simply remove it, along with `roman` and `sans`.

jfbu (commented):
The effect of this is that only the last specified styling class prevails. For example in #1989 an example is given where `k+kt` is passed to LaTeX formatter. With this PR, the effet of `k` (which in default style is `"bold #008000"`) will be entirely be shadowed by `kt` (`"nobold #B00040"`). So this PR would also fix #1989.

But is this desirable? looking quickly at the LaTeX formatter for the default style I see no token class generates say "bold italic". But currently the combination of two classes can achieve a bold italic. With this PR, this will not be possible.

Another remark is that the same effect as the one obtained here could be obtained by simply modifying only one part of the output of the LaTeX formatter for the highlighting: in place of 
```latex
\def\PY@toks#1+{\ifx\relax#1\empty\else%
    \PY@tok{#1}\expandafter\PYG@toks\fi}
```
one would use
```latex
\def\PY@toks#1+{\ifx\relax#1\empty\else%
    \PY@reset\PYG@tok{#1}\expandafter\PYG@toks\fi}
```
in effect cancelling at each new class via `\PYG@reset` what was configured by the prior one. This would make for a less bulky change to the LaTeX stylesheets for highlighting which one gets for example from
```
pygmentize -f latex -S default
```
To illustrate more, this PR will make `w` token class in the default style do
```latex
\let\PY@bf=\relax \let\PY@it=\relax \let\PY@ul=\relax \let\PY@ff=\relax \def\PY@tc##1{\textcolor[rgb]{0.73,0.73,0.73}{##1}}\let\PY@bc=\relax 
```
which is the same as doing first `\PY@reset` then the single 
```latex
\def\PY@tc##1{\textcolor[rgb]{0.73,0.73,0.73}{##1}}
```
hence my proposal to locate the `\PY@reset` externally to the macro associated to `w` class LaTeX macro implementation.

But again, is the effect of killing all inheritance the expected one? Only last class prevails.

PR #1393, opened 2020-02-12
Add execline lexer

sysvinit:
This introduces a lexer for [execline](https://skarnet.org/software/execline), a small scripting language. Execline isn't really a "shell" language per se (like bash, etc.), however the syntax is sufficiently similar that I thought it would be sensible to add it to `pygments.lexers.shell`.

### DIFF
--- a/pygments/lexers/shell.py
+++ b/pygments/lexers/shell.py
@@ -889,6 +889,7 @@ class ExeclineLexer(RegexLexer):
             (r'(?s)"(\\.|[^"\\$])*"', String.Double),
             (r'"', String.Double, 'string'),
             (r'\s+', Text),
+            (r'[^\s{}$"\\]+', Text)
         ],
         'string': [
             (r'"', String.Double, '#pop'),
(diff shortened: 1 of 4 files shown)


pygments/lexers/shell.py
```diff
@@ -847,3 +848,59 @@ class FishShellLexer(RegexLexer):
             include('root'),
         ],
     }
+
+class ExeclineLexer(RegexLexer):
+    """
+    Lexer for Laurent Bercot's execline language
+    """
```
Anteru: Needs a `versionadded` statement, this will go into 2.6.

pygments/lexers/shell.py
```diff
@@ -847,3 +848,59 @@ class FishShellLexer(RegexLexer):
             include('root'),
         ],
     }
+
+class ExeclineLexer(RegexLexer):
+    """
+    Lexer for Laurent Bercot's execline language
```
Anteru: Can you please provide a link to the language here?

sysvinit:
Okay, I'll take a look at that in a couple of days when I have some more free time again.

sysvinit:
I think I've covered all the requested changes -- does this all look okay?

pygments/lexers/shell.py
sysvinit: Ah, it looks like I missed the 2.6 release; I'll bump this to 2.7 instead.

Anteru:
Yes it does :) Merged for the next release, thanks for your contribution!

PR #1399, opened 2020-02-24
Matlabsession line continuation

bryanwweber:
There are a few changes wrapped up here:

1. Allow explicit line continuation in the `matlabsession` lexer. This allows this syntax:
   ```matlab
   set('T', 100.0, ...
   'P', 1.0e5);
   ```
   Previously, the second line was marked as generic output.
2. ~Keywords previously could only be ended by a word boundary, which does not include a newline. However, some keywords (`else`, `end`) are often on their own line. These keywords are now highlighted correctly.~
3. The ellipses used for line continuations are now highlighted as a Keyword, which better matches the built-in Matlab highlighter.
4. The functions and keywords which should be highlighted have been updated from Matlab R2018a. I removed the old keywords that did not appear in the output of the included bash script, but they could be retained for backwards compatibility. I could also update the lists from Matlab R2020a, but since I had R2018a handy, I just used that. Let me know!

Thank you!

### DIFF
--- a/pygments/lexers/matlab.py
+++ b/pygments/lexers/matlab.py
@@ -187,6 +187,7 @@ class MatlabSessionLexer(Lexer):
 
         curcode = ''
         insertions = []
+        continuation = False
 
         for match in line_re.finditer(text):
             line = match.group()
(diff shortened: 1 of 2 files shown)


bryanwweber:
Hi, Do I need to make any changes to this pull request? Thank you!

Anteru:
Hi, sorry, didn't get around to this earlier. Would you mind adding a unit test for the continuation case? There is a `testcase` formatter which should make this easy (hopefully :) ).

bryanwweber:
@Anteru Thanks for the feedback! I added some tests here. I removed the commit that handled point 2 above, because I couldn't reproduce it on any versions going back to late last year. I also found and fixed one other bug, which was that in syntax like `x  = 100;` the `= 100;` was being formatted as String due to the `commandargs` state. This is related to changes from #1328. With the changes here, if an `=` or an operator is encountered, the `commandargs` state is popped. I hope that's the right fix there, I couldn't see an easier way to do it.

Anteru:

bryanwweber:
Thank you!

PR #1400, opened 2020-02-26
Move Python 3-only exceptions into PythonLexer

verhovsky:


verhovsky:
If you go to https://docs.python.org/3/library/exceptions.html#exception-hierarchy you'll see that these exceptions aren't highlighted.

birkenfeld:
Thanks!

PR #1402, opened 2020-02-29
sql lexer: Add temporal support keywords

lonetwin:
SQL:2011 introduced support for [temporal databases](https://en.wikipedia.org/wiki/SQL:2011) via new keywords to create and manage temporal tables. This patchset adds these new keywords to the sql lexer.

One concrete example of these feature is the temporal tables support in [MariaDB](https://mariadb.com/kb/en/temporal-data-tables/). As evident from the code listing on that page, the syntax highlighting for these keywords is not as widespread as expected (I wouldn't be surprised if code listing on that page ultimately use pygments for syntax highlighting !)


Anteru:
Merged, thanks!

PR #1404, opened 2020-03-01
Add Windows CI support & fix tests on Windows

Anteru:
Adds Windows to the test matrix, and fixes the tests on Windows.


Anteru:
Previously we had a few failures on Windows, and with the recent changes to `img.py` which are Windows-specific I want to make sure we got it covered properly.

Anteru:
Thanks for the review!

PR #1405, opened 2020-03-02
Fix #1247 by looking up in both HKLM and HKCU.

### LINKED ISSUE #1247: FontManager windows font locations problem
The problem is that it searches the registry only for fonts installed to HKEY_LOCAL_MACHINE, and ignores HKEY_CURRENT_USER. On Windows there's no easy way for a user to know which fonts are installed in which, and it's easy to have fonts installed to one and not the other, so the missing font error this creates can be very confusing.

Easy to fix by checking HKEY_CURRENT_USER first, and then looking in HKEY_LOCAL_MACHINE only if the font is not found there. To get this working locally, this is what I've done to **formatters/img.py FontManager._create_win**:

``` python
    def _create_win(self):
        keyfound = False
        fontfound = False
        lookuperror = None
        keynames = [ (_winreg.HKEY_CURRENT_USER, r'Software\Microsoft\Windows NT\CurrentVersion\Fonts'),
                     (_winreg.HKEY_CURRENT_USER, r'Software\Microsoft\Windows\CurrentVersion\Fonts'),
                     (_winreg.HKEY_LOCAL_MACHINE, r'Software\Microsoft\Windows NT\CurrentVersion\Fonts'),
                     (_winreg.HKEY_LOCAL_MACHINE, r'Software\Microsoft\Windows\CurrentVersion\Fonts') ]
        for keyname in keynames:
            try:
                key = _winreg.OpenKey(*keyname)
                keyfound = True
                try:
                    path = self._lookup_win(key, self.font_name, STYLES['NORMAL'], True)
                    self.fonts['NORMAL'] = ImageFont.truetype(path, self.font_size)
                    for style in ('ITALIC', 'BOLD', 'BOLDITALIC'):
                        path = self._lookup_win(key, self.font_name, STYLES[style])
                        if path:
                            self.fonts[style] = ImageFont.truetype(path, self.font_size)
                        else:
                            if style == 'BOLDITALIC':
                                self.fonts[style] = self.fonts['BOLD']
                            else:
                                self.fonts[style] = self.fonts['NORMAL']
                    fontfound = True
                    _winreg.CloseKey(key)
                    break
                except FontNotFound as err:
                    lookuperror = err
                    _winreg.CloseKey(key)
            except EnvironmentError:
                pass
        if not fontfound:
            raise lookuperror
        if not keyfound:
            raise FontNotFound('Can\'t open Windows font registry key')
```

Anteru:
This is slightly different from the code posted inside the issue:
* Use finally to close the registry key no matter what
* Don't use break, but return directly on the first hit

### DIFF
--- a/pygments/formatters/img.py
+++ b/pygments/formatters/img.py
@@ -164,31 +164,40 @@ class FontManager:
             return None
 
     def _create_win(self):
-        try:
-            key = _winreg.OpenKey(
-                _winreg.HKEY_LOCAL_MACHINE,
-                r'Software\Microsoft\Windows NT\CurrentVersion\Fonts')
-        except EnvironmentError:
+        keyfound = False
+        fontfound = False
+        lookuperror = None
+        keynames = [ (_winreg.HKEY_CURRENT_USER, r'Software\Microsoft\Windows NT\CurrentVersion\Fonts'),
+                     (_winreg.HKEY_CURRENT_USER, r'Software\Microsoft\Windows\CurrentVersion\Fonts'),
+                     (_winreg.HKEY_LOCAL_MACHINE, r'Software\Microsoft\Windows NT\CurrentVersion\Fonts'),
+                     (_winreg.HKEY_LOCAL_MACHINE, r'Software\Microsoft\Windows\CurrentVersion\Fonts') ]
+        for keyname in keynames:
             try:
-                key = _winreg.OpenKey(
-                    _winreg.HKEY_LOCAL_MACHINE,
-                    r'Software\Microsoft\Windows\CurrentVersion\Fonts')
+                key = _winreg.OpenKey(*keyname)
+                keyfound = True
+                try:
+                    path = self._lookup_win(key, self.font_name, STYLES['NORMAL'], True)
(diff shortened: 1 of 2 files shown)


pygments/formatters/img.py
```diff
@@ -164,31 +164,40 @@ def _lookup_win(self, key, basename, styles, fail=False):
...
-                else:
-                    if style == 'BOLDITALIC':
-                        self.fonts[style] = self.fonts['BOLD']
-                    else:
-                        self.fonts[style] = self.fonts['NORMAL']
-        finally:
-            _winreg.CloseKey(key)
+                pass
+        if not fontfound:
+            raise lookuperror
```
birkenfeld: this could "raise None" if all the `OpenKey`s raise EnvironmentError. Should it go below `if not keyfound`?

Anteru: Hm, you're right. Maybe better to do:

```
else:
    raise FontNotFound(..)
if not fontfound:
    rause lookuperror
```

and remove the `keyfound` wholesale?

birkenfeld: sure, but then you need to also `break` the loop where `lookuperror` is assigned.

PR #1406, opened 2020-03-04
a filter for math symbols

lsf37:
This filer replaces math symbols from e.g. LaTeX and Isabelle with
corresponding unicode. It could be expanded to other math-heavy
languages.

(Ported from an existing bitbucket PR)


lsf37:
Oops, failing tests. I guess things have moved on from then, will look into it.

lsf37:
Ok, fixed.

Anteru:
Can you please add a unit test for this, just to make sure there is some code calling it running during CI/CD?

lsf37:
Sure, will probably have a look tomorrow.

lsf37:
Rebased and test added. I wasn't quite sure where exactly it should go. It's now together with the other filter tests, which seemed appropriate.

Anteru:
Yep, that seems fine. I want to make sure that CI is covering this. Thanks!

Issue #1407, opened 2020-03-04
Add lexer for YANG 1.1


gribok:
Add YANG lexer to pygments.

Already two pull requests #1305 and #1297 tried to add a lexer for YANG 1.1, based on [RFC7950](https://tools.ietf.org/html/rfc7950#page-184).
 
I reviewed the file changes. Based on both PRs, I will provide a new version of lexer with some enhancement.

gribok:
Pull request created. See #1408

PR #1408, opened 2020-03-04
Add lexer for YANG 1.1

gribok:
Based on issue #1407, I created a new lexer for YANG. 

### Features
* Found and added additional keywords based on [RFC7950](https://tools.ietf.org/html/rfc7950#page-184) to lexer
* Seperating keywords in groups, based on BNF definition of [RFC7950](https://tools.ietf.org/html/rfc7950#page-184) for maintenance
* Highlighting of constants, like `true`, `false` or `deprecated` with special token
* Highlighting of namespace-prefix, like `type namespace-prefix:counter64;`. `namespace-prefix` will have special token. 
For example `namespace urn:ieee:std:802.1AX:yang:ieee802-dot1ax;` will have no special token
* Prefix-highlighting should be the complex regex pattern of all pattern. Due to that, I created test cases to validate prefix-highlighting
* Create example yang file for testing
* No breaked rules in pylint


gribok:
Simplify ruleset for YANG lexer based on PR rouge-ruby/rouge#1458, which I have created.
If you accept my PR, please squash the commits.

Anteru:
Hi, sorry for not getting to this earlier. This is looking good, one question on the example file -- where is that from? Could there be a license incompatibilities with it?

gribok:
> Hi, sorry for not getting to this earlier. This is looking good, one question on the example file -- where is that from? Could there be a license incompatibilities with it?

Good point. I changed the example file to an own created file.

pygments/lexers/yang.py
```diff
@@ -0,0 +1,101 @@
...
+from pygments.lexer import (RegexLexer, bygroups, words)
+from pygments.token import (Text, Token, Name, String, Comment,
+                            Number)
+
+__all__ = ['YangLexer']
+
+class YangLexer(RegexLexer):
+    """
+    Lexer for `YANG <https://tools.ietf.org/html/rfc7950/>`_, based on RFC7950
+    """
```
Anteru, line 21: Please add a `.. versionadded:: 2.7` line here.

gribok: Done. Thanks for support.

Anteru:

PR #1413, opened 2020-03-10
Added support for C++14 separators in numbers

mpusz:
C++14 added optional separators with `'` in numbers (<https://en.cppreference.com/w/cpp/language/floating_literal>). The change above splits numbers definitions to C and C++. C is left with the original implementation while C++ gets `'` support for every place beside the first one.

This makes the code below highlight properly:

```cpp
struct kibi : units::prefix<kibi, prefix, "Ki", ratio<                    1'024>> {};
struct mebi : units::prefix<mebi, prefix, "Mi", ratio<                1'048'576>> {};
struct gibi : units::prefix<gibi, prefix, "Gi", ratio<            1'073'741'824>> {};
struct tebi : units::prefix<tebi, prefix, "Ti", ratio<        1'099'511'627'776>> {};
struct pebi : units::prefix<pebi, prefix, "Pi", ratio<    1'125'899'906'842'624>> {};
struct exbi : units::prefix<exbi, prefix, "Ei", ratio<1'152'921'504'606'846'976>> {};
```

### DIFF
--- a/pygments/lexers/c_cpp.py
+++ b/pygments/lexers/c_cpp.py
@@ -55,11 +55,6 @@ class CFamilyLexer(RegexLexer):
             (r'(L?)(")', bygroups(String.Affix, String), 'string'),
             (r"(L?)(')(\\.|\\[0-7]{1,3}|\\x[a-fA-F0-9]{1,2}|[^\\\'\n])(')",
              bygroups(String.Affix, String.Char, String.Char, String.Char)),
-            (r'(\d+\.\d*|\.\d+|\d+)[eE][+-]?\d+[LlUu]*', Number.Float),
-            (r'(\d+\.\d*|\.\d+|\d+[fF])[fF]?', Number.Float),
-            (r'0x[0-9a-fA-F]+[LlUu]*', Number.Hex),
-            (r'0[0-7]+[LlUu]*', Number.Oct),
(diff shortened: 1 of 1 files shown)


Anteru:
Looks like there's a conflict due to the large amount of updates recently done on the C++ lexer. Would you mind checking if this is still needed? If so, I'd appreciate if you could add a unit test to make sure this works as expected for C/C++. (And sorry it took me a while to get around to this one.)

mpusz:
I believe it is still needed as new changes did not provide support for a `'` character. Regarding the unit test, I might try to create something but would prefer not to do it as I really suck at Python and I do not know the environment that well. It would require a significant amount of time for me to learn and set up everything.

mpusz:
BTW, I applied the changes in a really hacky way here: https://github.com/mpusz/units/blob/5dd9eaac87a50adc62b170987ebac2d1344d63c5/docs/conf.py#L1-L234.

and the output is OK right now:

https://mpusz.github.io/units/framework/units.html#prefixed-unit

Anteru:
Looks like this has been already added in the meantime in one of the various C++ fixes, at least your example works fine now (sorry, there was a time with tons of C++ changes coming and your contribution fell through the cracks :/ ) Closing.

PR #1426, opened 2020-04-12
Update nim

wltsmrz:
https://nim-lang.org/docs/manual.html#lexical-analysis-identifiers-amp-keywords

* Removed: with, without, atomic, shared
* Added: using


Anteru:
Merged, thanks!

PR #1427, opened 2020-04-13
Remove Duplicate Keyword in YANG Lexer

gribok:
I have found a duplicate `identity` keyword definition and removed it.


Anteru:
Good catch! I'll merge this once CI finishes.

Anteru:
Thanks for your contribution!

PR #1432, opened 2020-04-21
Make HttpLexer compatible with RFC 7230 (section 3.1.2).

Josholith:
Specifically this addresses for the case where only a numeric HTTP status code is returned (eg. 200) and no textual reason phrase (eg. OK).  Strictly according to RFC 7230, the white space just after the status code number is NOT optional, and in fact Tomcat 8.5 behaves this way, emitting status lines like "HTTP/1.1 200 \n" (note the white space after the 200).

### DIFF
--- a/pygments/lexers/textfmts.py
+++ b/pygments/lexers/textfmts.py
@@ -179,7 +179,7 @@ class HttpLexer(RegexLexer):
              bygroups(Name.Function, Text, Name.Namespace, Text,
                       Keyword.Reserved, Operator, Number, Text),
              'headers'),
-            (r'(HTTP)(/)(1\.[01]|2|3)( +)(\d{3})(?:( +)([^\r\n]+))?(\r?\n|\Z)',
+            (r'(HTTP)(/)(1\.[01]|2|3)( +)(\d{3})(?:( +)([^\r\n]*))?(\r?\n|\Z)',
              bygroups(Keyword.Reserved, Operator, Number, Text, Number, Text,
                       Name.Exception, Text),
              'headers'),
--- a/tests/test_textfmts.py
+++ b/tests/test_textfmts.py
@@ -46,6 +46,20 @@ def test_http_status_line_without_reason_phrase(lexer):
     assert list(lexer.get_tokens(fragment)) == tokens
 
 
+def test_http_status_line_without_reason_phrase_rfc_7230(lexer):
+    fragment = u'HTTP/1.1 200 \n'
+    tokens = [
+        (Token.Keyword.Reserved, u'HTTP'),
+        (Token.Operator, u'/'),
+        (Token.Number, u'1.1'),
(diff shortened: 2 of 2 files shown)


Josholith:
This is my first PR to pygments, please be patient and nice!

Context for this was that I was using [HTTPie](https://github.com/jakubroztocil/httpie) (which uses pygments for formatting and highlighting), and getting all red HTTP headers displayed when returning a successful request which happened to be served from a Tomcat server.

After digging in, I had found it unusual that the HTTP status line looked like `HTTP/1.1 200` and not `HTTP/1.1 200 OK`, as I am accustomed to seeing the text status codes (like `Not Found`, `Forbidden`, et al.).

But when I tried `HTTP/1.1 200` directly through pygments HttpLexer it was interpreted just fine, so I was puzzled.  On deeper inspection (via `curl | hexdump`), I saw the status line had a trailing whitespace after the status code: `"HTTP/1.1 200 "` (without the quotes)

Apparently in [RFC 7230 Section 3.1.2](https://tools.ietf.org/html/rfc7230#section-3.1.2), the text status is are now optional but the whitespace between the numeric code and text description is NOT optional and must be retained, even if the server chooses to write an empty status text.

The RFC reads ...
> The first line of a response message is the status-line, consisting of the protocol version, a space (SP), the status code, another space, a possibly empty textual phrase describing the status code, and ending with CRLF.
>
>     status-line = HTTP-version SP status-code SP reason-phrase CRLF

Emphasis here is that the reference to "another space" is not optional or empty.

This PR makes HttpLexer compatible with any of these status line variants:
- `"HTTP/1.1 200 OK"`
- `"HTTP/1.1 200"`
- `"HTTP/1.1 200 "`  (new)

The first two variants other two were already supported by HttpLexer.  Tangent: the second variant is not strictly compliant with RFC 7230 (as it lacks the required whitespace after the status code), but I would imagine encountering it in the wild enough I wouldn't remove support for it.

The third variant above is the one I have added support for in this PR, which matches the strict specification of RFC 7230 (and Tomcat's implementation) with its required whitespace after the status code.

Josholith:
Easy method to reproduce the issue ...
```python
from pygments.lexers import HttpLexer
from pygments.formatters import TerminalFormatter

print(highlight("HTTP/1.1 200 OK", HttpLexer(), TerminalFormatter()))
print(highlight("HTTP/1.1 200"   , HttpLexer(), TerminalFormatter()))
print(highlight("HTTP/1.1 200 "  , HttpLexer(), TerminalFormatter()))
```

Josholith:
@jakubroztocil, this may be familiar to you

jakubroztocil:
Good catch. I missed this aspect when I was working on the PR that added support for phrase-lesses statuses (#1322).

Anteru:
Merged, thanks a lot for the nice commit including an unit test!

PR #1436, opened 2020-05-01
Add SystemVerilog (IEEE 1800-2012) keywords

cjdrake:
See Table 22-6 -- IEEE 1800-2012 additional reserved keywords

* implements
* interconnect
* nettype
* soft


Anteru:

PR #1439, opened 2020-05-02
Several small changes to the SystemVerilog lexer

cjdrake:
Also one new unit test that points out a few issues.


cjdrake:
This isn't as good as a unit test, but I have been testing this branch here: http://www.cjdrake.com/testing123.html#testing123

PR #1441, opened 2020-05-03
Add support for (System)Verilog escaped identifiers

whitequark:
The escaped identifiers start with a backslash, end with whitespace, and can include any character that is not whitespace.

See IEEE 1364-2005 §3.7.1 and IEEE 1800-2017 §5.6.1.

### DIFF
--- a/pygments/lexers/hdl.py
+++ b/pygments/lexers/hdl.py
@@ -110,6 +110,7 @@ class VerilogLexer(RegexLexer):
              Keyword.Type),
             (r'[a-zA-Z_]\w*:(?!:)', Name.Label),
             (r'\$?[a-zA-Z_]\w*', Name),
+            (r'\\(\S+)', Name),
         ],
         'string': [
             (r'"', String, '#pop'),
@@ -316,6 +317,7 @@ class SystemVerilogLexer(RegexLexer):
              Keyword.Type),
             (r'[a-zA-Z_]\w*:(?!:)', Name.Label),
             (r'\$?[a-zA-Z_]\w*', Name),
+            (r'\\(\S+)', Name),
         ],
         'classname': [
             (r'[a-zA-Z_]\w*', Name.Class, '#pop'),


pygments/lexers/hdl.py
```diff
@@ -316,6 +317,7 @@ class SystemVerilogLexer(RegexLexer):
              Keyword.Type),
             (r'[a-zA-Z_]\w*:(?!:)', Name.Label),
             (r'\$?[a-zA-Z_]\w*', Name),
+            (r'\\(\S+)\s', Name),
```
cjdrake, line 320: The '\\(\S+)' looks mostly right to me (Verilog only defines behavior on ASCII characters, and I think we will also capture Unicode with this, but that's not important). Is it correct to include the '\s' whitespace character with the identifier? Does it work correctly if you leave this part off and let the whitespace capture rules chomp that part?

whitequark: > Is it correct to include the '\s' whitespace character with the identifier?

No, and neither is the backslash part of the identifier. The identifiers `foo` and `\foo ` (note the space) are completely equivalent in every regard.

> Does it work correctly if you leave this part off and let the whitespace capture rules chomp that part?

I can check that, but what about the backslash?

birkenfeld:
IMO it's perfectly fine to have the backslash included in the token.  It might not be part of the resulting identifier name, but it's part of how the identifier is spelled in the source.  It's the same principle as including the quotes for string literals in the String tokens.

The whitespace should be fine to leave out.

whitequark:
Updated to address review.

cjdrake:
Maybe you can use `bygroups` to mark the r'\\' as punctuation?

eg:

```
r'(\\)(\S+)', bygroups(Punctuation, Name))
```

If the trailing whitespace character is necessary for some reason, you could still include it like this:

```
r'(\\)(\S+)\s', bygroups(Punctuation, Name, Text))
```

The spec says: "Escaped identifiers shall start with the backslash character (\) and end with white space (space, tab, newline). ... Neither the leading backslash character nor the terminating white space is considered to be part of the identifier."

So having the leading \\ and trailing characters as part of the capture group seems legit :).

birkenfeld:
LGTM as is.

birkenfeld:
Not sure what's up with CI, I tested locally and this is small anyway.

Anteru:
I kicked off CI manually again, must have been sleeping at night.

PR #1442, opened 2020-05-05
fixed fortran literal regex pattern

amanotk:
This fix let the fortran lexer properly handle literal expressions with kind parameters specified as follows:
```
  integer :: n = 128_4
  real(8) :: pi = 3.14_8
```

I believe this is part of the fortran standard.


birkenfeld:
Looks good, thanks!

PR #1456, opened 2020-05-19
Reflect Perl6 to Raku language rename

sumanstats:
For uniformity with language rename 

Perl6Lexer => RakuLexer
PERL6 => RAKU
Remove compilers of historical interests niecza, pugs

### DIFF
--- a/pygments/lexers/perl.py
+++ b/pygments/lexers/perl.py
@@ -17,7 +17,7 @@ from pygments.token import Text, Comment, Operator, Keyword, Name, String, \
     Number, Punctuation
 from pygments.util import shebang_matches
 
-__all__ = ['PerlLexer', 'Perl6Lexer']
+__all__ = ['PerlLexer', 'RakuLexer']
 
 
 class PerlLexer(RegexLexer):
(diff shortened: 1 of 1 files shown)


sumanstats:
As Perl and Raku are different languages, it would be better to separate `RakuLexer` from `perl.py` and have separate file for `Raku` like `raku.py`.

birkenfeld:
In general I'm fine with changing the name, but we need a compatibility alias for the lexer class in the `perl` module.

The `agile` module is for compatibility only, so the name there should remain `Perl6Lexer`, and there is no need to import the new one.

The entry in `doc/languages` could still state "formerly known as Perl 6".

Anteru:
Looking at [the Raku website](https://raku.org/), they no longer claim to be Perl6. It makes sense to me to rename `Perl6` to `Raku` in this context. I guess we can resurrect this by moving the `Perl6Lexer` to `raku.py`, then add an alias in `perl.py` similar as we did when we transitioned Python2 to 3. `agile.py` is one of the deprecated "summary" models we'd want to remove in Pygments 3, right?

birkenfeld:
Yep.

PR #1458, opened 2020-05-20
Configs lexer: Allow Kconfig derivate names

MaximeChretien:
In the linux source code a lot of files are named Kconfig.debug and
there are also other derivatives like Kconfig.x86 or Kconfig-nommu


Anteru:
Merged, thanks, nice find!

PR #1460, opened 2020-05-23
Update SystemVerilog literal constants

cjdrake:
The original implementation was missing some of the more arcane features
such as underbars, the character 's' for signed/unsigned, support for
spaces before/after the base specifier, capital letter base specifiers
(ie 'B 'D 'H), and the 4-state 'xXzZ?' characters.

For regular integers, the 'l' and 'L' suffixes are not valid.
That is, unlike C, in Verilog '42L' is not a valid int literal.

Create a new test that exercises most of the interesting kinds of
SystemVerilog numbers.

This fixes a couple minor issues with what type of number the lexer
returns. For example, Numbers like '42' used to return Integer.Hex,
but now return Integer.Decimal.


Anteru:
Thanks a lot, looks like SystemVerilog is getting a lot of love here :)

Anteru:
Merged, thanks again for your contributions! It's great to get them complete with a nice description, comments and unit tests!

PR #1462, opened 2020-05-27
Change Verilog 'wo' to 'wor'

cjdrake:
There is no 'wo' keyword.
This was probably supposed to be 'wor', short for "wired OR".


birkenfeld:
Thanks!

PR #1463, opened 2020-05-27
Refactor SystemVerilog unit tests

cjdrake:
Most of the contents of these two unit tests are static.
Move things around so the entire test fits on a single page,
for better readability/maintainability.

Name the code part <TEST_NAME>_TEXT,
and the tokens part <TEST_NAME>_TOKENS.

Choosing "text" b/c it's the parameter name to the
lexer.get_tokens(text) method.


birkenfeld:
Looks good!

PR #1466, opened 2020-05-29
Matlab class properties

drmoose:
Given the following Matlab code,

```matlab
classdef Name < dynamicprops
    properties
        % i am a comment
        name1
        name2
    end
    methods
        % i am also a comment
        function self = Name()
            % i am a comment inside a constructor
        end
    end
end
```

pygments' matlab lexer gets very confused indeed. This attempts to improve the situation. Although it does not capture all of [the properties syntax](https://www.mathworks.com/help/matlab/matlab_oop/defining-properties.html), it special-cases `properties` blocks to prevent the list of names from being misinterpreted as the "command syntax." It also prevents the comments from being mistaken for string literals.

Let me know anything I can do to improve the PR. In particular, if you'd like support for the full properties syntax, I can do that, but I wasn't sure of the best way to re-use token rules from the `'root'` in `'defprops'` while maintaining readability.

### DIFF
--- a/pygments/lexers/matlab.py
+++ b/pygments/lexers/matlab.py
@@ -20,6 +20,7 @@ from pygments.lexers import _scilab_builtins
 __all__ = ['MatlabLexer', 'MatlabSessionLexer', 'OctaveLexer', 'ScilabLexer']
 
 
+
 class MatlabLexer(RegexLexer):
     """
     For Matlab source code.
@@ -107,7 +108,11 @@ class MatlabLexer(RegexLexer):
             (r'%\{\s*\n', Comment.Multiline, 'blockcomment'),
             (r'%.*$', Comment),
             (r'(\s*^\s*)(function)\b', bygroups(Whitespace, Keyword), 'deffunc'),
-            (r'(\s*^\s*)(properties)\b', bygroups(Whitespace, Keyword), 'defprops'),
+            (r'(\s*^\s*)(properties)(\s+)(\()',
+             bygroups(Whitespace, Keyword, Whitespace, Punctuation),
+             ('defprops', 'propattrs')),
+            (r'(\s*^\s*)(properties)\b',
+             bygroups(Whitespace, Keyword), 'defprops'),
 
             # from 'iskeyword' on version 9.4 (R2018a):
             # Check that there is no preceding dot, as keywords are valid field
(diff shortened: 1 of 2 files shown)


Anteru:
You can separate out rules and include them, which makes it easy to reuse them in various places. See https://github.com/pygments/pygments/blob/master/doc/docs/lexerdevelopment.rst, the example at "You can include the rules of a state in the definition of another. "

drmoose:
Using that construct to try to get closer to full syntax support, I find there are some inconsistencies with whitespace that I'm not sure how best to address.

Should spaces in the code be `Token.Text` or `Token.Text.Whitespace`? And, should contiguous whitespace be a single token (`(Token.Text, '\n    ')`) or one token per character (`(Token.Text, '\n'), (Token.Text, ' '), ...`)?

apjanke:
Ping. @drmoose, are you still working on this?

drmoose:
@apjanke Oh whoops. This thread completely slipped my mind. I've got a bunch of uncommitted changes and I was fighting with the inconsistent treatment of whitespace when I got distracted by other things .... aaaand that was apparently four months ago. Linear time is annoying.

Let's see how close I can get this to commit-worthy today...

apjanke:
That'd be great; thanks! It would be nice to have this to build on for my work on #1684.

drmoose:
@apjanke Do you have the code snippet you posted in #1684 available in a form other than a screenshot? I'd like to run through my local branch as a test case. I've made no effort to support the `arguments` syntax but I'm pretty sure I've got property constraints working.

apjanke:
Oh, sure! Sorry for not posting code as text too, originally. I should know better.

```matlab
classdef SomeClass < SomeOtherClass

  properties
    x (1,1) double = 42
    y
  end

  methods
    function this = SomeClass()
    end
  end

end

function anExampleFunction(foo, bar)
  arguments
    foo (1,1) double
    bar string = "whatever"
  end

  fprintf('Hello, world!\n')
end
```

It's from https://github.com/janklab/MatlabProjectTemplate/blob/main/doc-src-jekyll/index.md.

drmoose:
image

Not bad. As expected, `arguments` could be better.

apjanke:
Here's a more thorough version:

```matlab
classdef SomeClass < SomeOtherClass

  properties
    x (1,1) double = 42
    y
  end

  methods
    function this = SomeClass()
    end
  end

end

function anExampleFunction(foo, bar, baz, qux)
  arguments
    foo
    bar (1,1) double
    baz string = "whatever"
    qux string = "foo" {mustBeMember(qux, ["foo" "bar" "baz"])}
  end

  fprintf('Hello, world!\n')
end
```

apjanke:
Cool. If you want to leave off here, I can pick up on `arguments` and type annotations.

drmoose:
That's probably a good idea as I'm not terribly familiar with either syntax. Fork me if you want, but you now also have write access on my branch. :)

apjanke:
Hopefully you can just get this PR merged soon, and then I can work off of that. I'd rather not do separate PRs or pushes to your branch/PR – that can get messy, and I'm going to be doing a _lot_ of screwing around and testing with this code.

Anteru (approved):

Anteru:
@apjanke Please go ahead :)

PR #1467, opened 2020-05-29
Added separate lifetime lexing for Rust

### LINKED ISSUE #1459: Rust highlighting changes based on lifetime
I have this code sample:
```rust
impl MyStruct {
    fn foo(
      a: &AThing, 
      b: &'static BThing, 
      c: CThing
    ) -> usize {
        5
    }
}
```
Using pygmentize, `AThing` and `CThing` are bold, while `BThing` is not. They are all types, so they should have the same formatting.

I also realized that Github's formatting is also wrong in this case, the `&` symbol should not change color (this is correct in Pygments).

EmreAtes:
Fixes #1459

### DIFF
--- a/pygments/lexers/rust.py
+++ b/pygments/lexers/rust.py
@@ -131,8 +131,7 @@ class RustLexer(RegexLexer):
             (r'b?r(#*)".*?"\1', String),
 
             # Lifetime names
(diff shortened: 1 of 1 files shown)


birkenfeld:
Nice, thanks!

PR #1472, opened 2020-06-02
Updated languages and templates list with more links

rys:
I tried to comprehensively fill out the list of language and template links in the documentation.

For each one I verified that every link works at the time of submission, and used my best judgement on whether the link points the best place to go and read about whatever the link represents, be it a programming language, programming system, language family of some kind, ISO or IEEE standard, etc.

If I couldn't find a good live link that I felt described the thing well, I fell back to English language Wikipedia. In some cases Wikipedia is all that's left for a language or system. One or two were just http -> https promotions. If there's a better name now for a language or system than was in the list, I changed it (at least Nimrod -> Nim, maybe others).

### DIFF
--- a/doc/languages.rst
+++ b/doc/languages.rst
@@ -8,205 +8,205 @@ Pygments supports an ever-growing range of languages. Watch this space...
 Programming languages
 ---------------------
 
-* ActionScript
-* Ada
-* Agda (incl. literate)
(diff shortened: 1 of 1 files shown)


Anteru:
That is ... quite impressive, thanks a lot for working through the list and leaving no stone unturned!

birkenfeld:
Indeed. This is great :fireworks:

PR #1477, opened 2020-06-07
Fix Solarized line number colors

### LINKED ISSUE #1356: Solarized line number wrong colors
The line number column has wrong colors (both font and background color).
[Solarized homepage](https://ethanschoonover.com/solarized/)

### Solarized dark
Current font color for line numbers: `base0` `#839496`
Font color should be for line numbers: `base01` `#586e75`
Current background color for line numbers: _not a Solarized color_ `#f0f0f0`
Background color should be for line numbers: `base02` `#073642`

### Solarized light
Current font color for line numbers: `base00` `#657b83`
Font color should be for line numbers: `base01` `#586e75`
Current background color for line numbers: _not a Solarized color_ `#f0f0f0`
Background color should be for line numbers: `base2` `#eee8d5`

pfertyk:
Fixes #1356.

Changes:
* the structure of line number elements and styles was changed to improve styling
* all style definitions were moved to `get_style_defs` (and removed from `CSSFILE_TEMPLATE`)
* documentation for `noclasses` option was updated
* `Style` class received new attributes for:
  * line numbers font color
  * line numbers background color
  * special line numbers font color
  * special line numbers background color
* line number font and background colors were defined for Solarized styles
* tests were added to check the HTML output for different combinations of line number related options


tests/test_cmdline.py
```diff
@@ -200,8 +200,9 @@ def test_H_opt():
 def test_S_opt():
     o = check_success('-S', 'default', '-f', 'html', '-O', 'linenos=1')
     lines = o.splitlines()
-    for line in lines:
-        # every line is for a token class
+    for line in lines[5:]:
```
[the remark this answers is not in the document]
pfertyk: I've refactored that part of HTML formatter and split the styles into 3 groups:
* linenos
* background
* tokens

This test now only check the token styles and therefore doesn't need to start with line 5. Another test, in `test_cmdline.py`, still starts with line 5, but there is a comment about that.

I've also noticed that multiple CSS prefixes don't work correctly with `.hll` class, so I've fixed that and added a test for that scenario.

pfertyk:
@Anteru are there any other fixes you would like me to include?

Anteru:
Nope, looks good, I'll merge as soon as I have some more time :)

pygments/formatters/html.py
```diff
@@ -510,23 +568,25 @@ def prefix(cls):
...
-                         (prefix(''), self.style.background_color, text_style))
-        if self.style.highlight_color is not None:
-            lines.insert(0, '%s.hll { background-color: %s }' %
-                         (prefix(''), self.style.highlight_color))
-        return '\n'.join(lines)
+        return prefix
+
+    @property
+    def _pre_style(self):
+        return 'line-height: 125%; margin: 0;'
```
HyukjinKwon, line 575: Just curious - in which way `margin: 0` makes it look better? In my case, it looks less pretty because there is no margin after the code. This looks more obvious when you write some explanation on the top and code after that few times, for example:

Oct-27-2020 18-30-12

HyukjinKwon: Another example:

**With `margin: 0`**

Screen Shot 2020-10-27 at 6 43 35 PM


**Without `margin: 0`**

Screen Shot 2020-10-27 at 6 44 02 PM

I trying to override the CSS now and I was curious why this change was made.

Anteru: You're seeing (I guess) because your theme doesn't have margin/padding on the paragraph either. I'd try to override the style if you care about specific padding/margin. That said, the `margin: 0` was not intentional (I think) so I'll revert that for 2.7.3.

Anteru: Removed in 4dede4010180c8a5a231628deedcf393565cf785

HyukjinKwon: Thanks @Anteru :).

pfertyk: Hi! I'm sorry if I'm a bit too late, but I'm pretty I've introduced the `margin: 0` rule for a reason. Since I've worked on this one some time ago, I'm not sure what exactly was it, but I'd guess it had something to do with making the table and non-table styles look more similar. An alternative might be that it had something to do with highlighting lines/special lines. There was a lot of different combinations of styles and highlight rules etc. and I believe `margin: 0` was introduced to make everything work better in regard to some of these combinations.

Having said that, if it actually looks better without `margin: 0` and it doesn't break anything else, I'm totally OK with removing this rule :)

HyukjinKwon:
@Anteru, please allow me a quick question here: when is expected release date of 2.7.3?

Anteru:
Today.

HyukjinKwon:
Thanks!

pygments/formatters/html.py
Anteru: I can't find the case where it would matter, and it does break various themes -- if you can point me to a problematic case then we can revisit alternative solutions. Ideally theme authors should have a way to override Pygments CSS rules, but for now this seems to be causing too much fallout :(

pfertyk: I tried to add all problematic cases to the test suite :P But while working on this a lot of dependencies came up and it's possible I've changed one thing too many ;) If you could just add a proper test with an explanation/justification, I would be very grateful. Right now, unfortunately, I'm not able to spend any time fixing this or checking why exactly was this change made. Reverting it seems like a good solution, as long as it's documented ;)

Anteru: Fair enough. I ran a bunch of manual tests and also spot-checked the test cases, and couldn't find a problematic case. Was hoping you'd remember something particularly problematic, but given the feedback we got, it seems that any rule changing the margin will cause more harm than good :(

PR #1482, opened 2020-06-20
Fix exception causes all over the codebase

cool-RR:
@birkenfeld This is a continuation of #1478

### DIFF
--- a/pygments/filters/__init__.py
+++ b/pygments/filters/__init__.py
@@ -777,8 +777,10 @@ class RaiseOnErrorTokenFilter(Filter):
             # issubclass() will raise TypeError if first argument is not a class
             if not issubclass(self.exception, Exception):
                 raise TypeError
-        except TypeError:
-            raise OptionError('excclass option is not an exception class')
+        except TypeError as e:
+            raise OptionError(
+                'excclass option is not an exception class'
+            ) from e
 
     def filter(self, lexer, stream):
         for ttype, value in stream:
--- a/pygments/filters/__init__.py
+++ b/pygments/filters/__init__.py
@@ -777,10 +777,8 @@ class RaiseOnErrorTokenFilter(Filter):
             # issubclass() will raise TypeError if first argument is not a class
             if not issubclass(self.exception, Exception):
                 raise TypeError
-        except TypeError as e:
-            raise OptionError(
-                'excclass option is not an exception class'
-            ) from e
+        except TypeError:
+            raise OptionError('excclass option is not an exception class')
 
     def filter(self, lexer, stream):
         for ttype, value in stream:
--- a/pygments/formatters/other.py
+++ b/pygments/formatters/other.py
@@ -74,16 +74,16 @@ class RawTokenFormatter(Formatter):
         if self.error_color is not None:
             try:
                 colorize(self.error_color, '')
-            except KeyError as e:
+            except KeyError:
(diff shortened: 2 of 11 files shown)


pygments/filters/__init__.py
```diff
@@ -777,8 +777,10 @@ def __init__(self, **options):
             # issubclass() will raise TypeError if first argument is not a class
             if not issubclass(self.exception, Exception):
                 raise TypeError
-        except TypeError:
-            raise OptionError('excclass option is not an exception class')
+        except TypeError as e:
+            raise OptionError(
+                'excclass option is not an exception class'
+            ) from e
```
birkenfeld, line 783: should use `from None`

pygments/formatters/other.py
```diff
@@ -74,16 +74,16 @@ def __init__(self, **options):
         if self.error_color is not None:
             try:
                 colorize(self.error_color, '')
-            except KeyError:
+            except KeyError as e:
                 raise ValueError("Invalid color %r specified" %
-                                 self.error_color)
+                                 self.error_color) from e
```
birkenfeld, line 79: `from None`

pygments/util.py
```diff
@@ -72,14 +72,14 @@ def get_int_opt(options, optname, default=None):
     string = options.get(optname, default)
     try:
         return int(string)
-    except TypeError:
+    except TypeError as e:
         raise OptionError('Invalid type %r for option %s; you '
                           'must give an integer value' % (
-                              string, optname))
-    except ValueError:
+                              string, optname)) from e
```
birkenfeld, line 78: `from None`

pygments/util.py
```diff
@@ -72,14 +72,14 @@ def get_int_opt(options, optname, default=None):
...
+    except ValueError as e:
         raise OptionError('Invalid value %r for option %s; you '
                           'must give an integer value' % (
-                              string, optname))
+                              string, optname)) from e
```
birkenfeld, line 82: `from None`

tests/test_basic_api.py
```diff
@@ -66,10 +66,10 @@ def test_random_input(cls):
     inst = cls()
     try:
         tokens = list(inst.get_tokens(test_content))
-    except KeyboardInterrupt:
+    except KeyboardInterrupt as e:
         raise KeyboardInterrupt(
             'interrupted %s.get_tokens(): test_content=%r' %
-            (cls.__name__, test_content))
+            (cls.__name__, test_content)) from e
```
birkenfeld, line 72: `from None`

cool-RR:
I'm strongly opposed to `from None`. I think that when people are debugging they have a very different mindset than when they are writing code. When you're debugging, and you get a very limited amount of information that may have been copy-pasted in an email by a non-technical person, you're very desperate for the information. Even if that information is a dead end for exploration, I at least want to know its content before I rule it out as a dead end.

Another issue is that some error-reporting systems show not only the traceback, but all the local variables on each of these levels. These supposedly irrelevant levels could have variables that make it clearer what this exception is about.

Obviously it's your project and your call, but I don't want to be the person responsible for another developer not getting exception information. I went over the places you marked and removed them from the PR so you could add these in a separate PR.

cool-RR:
@birkenfeld Got a decision?

cool-RR:
@birkenfeld Nudge

Issue #1485, opened 2020-06-23
Escape in strings


cykerway:
Currently pygments doesn't support escape in strings, which makes typesetting with *minted* (which uses pygments) in latex very limited. Actually there is a way to improve the latex formatter so that raw latex code can be embedded in strings, which can greatly enhance its ability. As an example, this shows how to render in different styles and languages, much similar to plain `fancyvrb`, but with auto code highlighting:

```tex
\begin{minted}[escapeinside=||,escapeinstr]{java}
public class HelloWorld {
    public static void main(String[] args) {
        System.out.println("abc.|{\rmfamily{|rmfamily|}}|.xyz");
        System.out.println("abc.|{\sffamily{|sffamily|}}|.xyz");
        System.out.println("abc.|{\ttfamily{|ttfamily|}}|.xyz");
        System.out.println("abc.|{\mdseries{|mdseries|}}|.xyz");
        System.out.println("abc.|{\bfseries{|bfseries|}}|.xyz");
        System.out.println("abc.|{\upshape{|upshape|}}|.xyz");
        System.out.println("abc.|{\itshape{|itshape|}}|.xyz");
        System.out.println("abc.|{\slshape{|slshape|}}|.xyz");
        System.out.println("abc.|{\scshape{|scshape|}}|.xyz");
        System.out.println("abc.|{\large{|large|}}|.xyz");
        System.out.println("abc.|{\small{|small|}}|.xyz");
        System.out.println("abc.|{\tiny{|tiny|}}|.xyz");
        System.out.println("abc.|{\japanesefont{|日本語|}}|.xyz");
        System.out.println("abc.|\RL{\arabicfont{|اللغة العربية|}}|.xyz");
        System.out.println("abc.|{\color{cyan}{cyan}}|.xyz");
        System.out.println("abc.|{\color{magenta}{magenta}}|.xyz");
        System.out.println("abc.|{\color{yellow}{yellow}}|.xyz");
        System.out.println("abc.|{\color{black}{black}}|.xyz");
    }
}
\end{minted}
```

I understand this is a python program not a latex one, but the popular latex package `minted` uses this program as an underlying tool. When pygments is patched, it is very easy to patch `minted`:

```
\minted@def@optcl@switch{escapeinstr}{-P escapeinstr}
```

![a](https://user-images.githubusercontent.com/392675/85452866-e13cc200-b5cd-11ea-9625-27e8dee37022.jpg)

cykerway:
[...]

PR #1486, opened 2020-06-23
Allows escape in strings using the latex formatter.

cykerway:
This patch adds code to break strings into substrings and escapes in
latex formatter. This allows users to embed raw latex formatting code in
strings, using command line options `escapeinside` and `escapeinstr`.

Escape in strings is off by default, and is enabled with command line
option `escapeinstr`.

Solves <https://github.com/pygments/pygments/issues/1485>.

### DIFF
--- a/pygments/formatters/latex.py
+++ b/pygments/formatters/latex.py
@@ -227,12 +227,18 @@ class LatexFormatter(Formatter):
 
         .. versionadded:: 1.2
 
+    `escapeinstr`
+        If set to ``True``, enables escaping to LaTeX in string literals, using
+        delimiter characters provided by `escapeinside`.  (default: ``False``).
+
+        .. versionadded:: x.y
+
     `escapeinside`
         If set to a string of length 2, enables escaping to LaTeX. Text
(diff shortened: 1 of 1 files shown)


eric-wieser:
I think this should be fixed in `LatexEmbeddedLexer` - see gh-1517 for an idea of what's going on there.

Anteru:
Given #1517 has been merged, is that still an issue?

cykerway:
Looks like it's still an issue unless you can produce correct result for this example:

~~~latex
\documentclass[10pt,a4paper]{article}
\usepackage{color}
\usepackage{minted}
\begin{document}
\begin{minted}[escapeinside=||]{java}
public class HelloWorld {
    public static void main(String[] args) {
        System.out.println("abc.|{\color{cyan}{cyan}}|.xyz");
        System.out.println("abc.|{\color{magenta}{magenta}}|.xyz");
        System.out.println("abc.|{\color{yellow}{yellow}}|.xyz");
        System.out.println("abc.|{\color{black}{black}}|.xyz");
    }
}
\end{minted}
\end{document}
~~~

d

PR #1497, opened 2020-07-25
Update ps1con prompt

vsalvino:
This change allows a more flexible PowerShell session prompt. The prompt now must start with `PS` and end with `>` but the inner contents are not important. This allow simpler prompts such as `PS>` which are more useful in documentation and more comparable to the Bash `$`.

Also adding documentation showing examples of how to use the prompts for each supported shell session. This was very difficult to look up previously, as one simply had to know the magic combination of characters that would be parsed as a prompt.


Anteru:
Does this work with remote sessions? There have been a few changes related to that recently, specifically adding the space before `>` which this change removes again -- did you check that those continue to work? Could you please add a few tests for the new syntax to make sure this doesn't break in the future (unit tests, not just documentation, as we don't check the documentation output.)

vsalvino:
The current unit tests do pass, however I will check and add new tests. The idea was to support a prompt which does not contain a path. We could probably add the space back in if necessary.

Anteru:
Ok -- I'm mostly interested in a test for this, as I can't tell you if the space should be there or not, but if someone ever touches that that piece again, there should be a test which makes sure that today's command line don't break.

vsalvino:
For reference, the default prompt is `PS>` https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_prompts?view=powershell-7#the-default-prompt

I will write some tests. This is my first time contributing to this project - could you link me to a good example test to model after?

Anteru:
Yes :) Take a look here: https://github.com/pygments/pygments/blob/master/tests/test_shell.py#L172 -- there's also a special formatter (`testcase`) which produces test cases directly, so you can provide a text snippet and get the tokens to test for directly.

Anteru:
Thanks, looks good!

PR #1500, opened 2020-07-31
Improve HTML formatter output.

Anteru:
With the previous changes, we started to emit one <pre> per line for
line numbers. This breaks for instance the Sphinx-RTD-Theme, which
expects the line numbers to be formatted the same way as the normal
content. This commit makes the following changes:

* Emit a single <pre> inside the linenos div
* Wrap individual lines into <span> as needed
* Update all tests
* Don't yield empty <span> elements when no style is specified

This also makes the .html test files look correct when looked at with a
browser, as there is no extra whitespace in them which needs stripping.

### DIFF
--- a/pygments/formatters/html.py
+++ b/pygments/formatters/html.py
@@ -677,7 +677,8 @@ class HtmlFormatter(Formatter):
                 else:
                     style = ''
 
-            line = '<pre%s>%s</pre>' % (style, line)
+            if style:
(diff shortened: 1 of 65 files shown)


Anteru:
The failure is unrelated to this PR; I've opened a separate branch to investigate that (b6d1d68de705e1cfc28a4188f792b29c545bf7ed). This does change the formatting for `linenos` quite a bit, but I believe this is the right fix.

PR #1507, opened 2020-08-13
Add async and await keywords to CythonLexer

mariushegele:


Anteru:
Thanks for the contribution!

PR #1511, opened 2020-08-15
fix punctuation not labeled in function definition in TypeScript

### LINKED ISSUE #1510: Typescript: Name.Other over extends to punctuation in function definition
```typescript
async function main() {
}
```

The above code produces this in Pygments 2.6.1:

image

The issue is caused by https://github.com/pygments/pygments/blob/master/pygments/lexers/javascript.py#L511 

```python
            # Match stuff like: function() {...}
            (r'([a-zA-Z_?.$][\w?.$]*)\(\) \{', Name.Other, 'slashstartsregex'),
```

which tags the `() {` as `Name.Other`. So I suggest change it to

```python
            # Match stuff like: function() {...}
            (r'([a-zA-Z_?.$][\w?.$]*)(?=\(\) \{)', Name.Other, 'slashstartsregex'),
```

felixhao28:
Fix #1510

### DIFF
--- a/tests/test_javascript.py
+++ b/tests/test_javascript.py
@@ -9,7 +9,7 @@
 
 import pytest
 
-from pygments.lexers import CoffeeScriptLexer
+from pygments.lexers import CoffeeScriptLexer, TypeScriptLexer
 from pygments.token import Token
 
 COFFEE_SLASH_GOLDEN = [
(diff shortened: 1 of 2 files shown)


Anteru:
Thanks for the PR! Would you mind adding a small unit test for that?

felixhao28:
The [tests/test_javascript.py](https://github.com/pygments/pygments/blob/master/tests/test_javascript.py) seems limited to CoffeeScript. I don't know if I should add another file for typescript tests.

felixhao28:
Test case added.

tests/test_javascript.py
```diff
@@ -83,3 +83,25 @@ def test_beware_infinite_loop(lexer):
...
     assert list(lexer.get_tokens(fragment)) == tokens
+
+@pytest.fixture(scope='module')
+def ts_lexer():
+    yield TypeScriptLexer()
+
+def test_function_definition(ts_lexer):
+    fragment = u'async function main() {\n}'
+    tokens = [
+        (Token.Name.Other, u'async'),
```
Mesteery, line 94:
```suggestion
        (Token.Keyword, u'async'),
```
The `async` should be a keyword (https://github.com/pygments/pygments/pull/1651)

PR #1513, opened 2020-08-18
Update PostgreSQL builtins

kurtmckee:
The existing auto-update code no longer worked with current Postgres source code (or with Python 3.8), so I've updated that. I also encountered newline problems when running this on Windows and adjusted to fix that.

Note that the email lexer's filename prevents urllib from importing, so I had to rename `email.py` to get the code running. It may be worthwhile to rename the email lexer in the future.

Please let me know if there's anything I'm missing with this. I read the contributing docs but didn't see anything about multiple commits so I tried to keep the history clean. I also ran the branch through tox but only on Python 3.8.

Thanks for any feedback you have!


Anteru:
Thanks for the contribution. I may be blind, but where was the `email.py` file renamed? Or was this a one-time thing to get the script to run?

kurtmckee:
Just a manual, local change while working on this. The error message was something like "email.parser is not a package", and that took a little bit of time to figure out.

I didn't rename the file in these commits but I wanted to point out that it may hamper update efforts in the future. =)

Issue #1516, opened 2020-08-20
The escapeinside argument to the latex formatter breaks pycon input


eric-wieser:
With `.\tmp.txt` containing
```python
>>> where_did_this_go = str('e_x e_y e_z')
```
the command
```bash
pygments -l pycon -f latex -O escapeinside=QQ .\tmp.txt
```
gives
```latex
\begin{Verbatim}[commandchars=\\\{\},codes={\catcode`\$=3\catcode`\^=7\catcode`\_=8}]
\PY{l+s+s1}{\PYZsq{}}\PY{l+s+s1}{e\PYZus{}x e\PYZus{}y e\PYZus{}z}\PY{l+s+s1}{\PYZsq{}}\PY{g+go}{)}
\end{Verbatim}
```

`Q` doesn't even appear anywhere in my input - so why is the output getting mangled?

PR #1517, opened 2020-08-20
Fix LatexEmbeddedLexer

eric-wieser:
Closes gh-1516


pygments/formatters/latex.py
```diff
@@ -446,30 +446,62 @@ def __init__(self, left, right, lang, **options):
...
-    def get_tokens_aux(self, index, text):
+    def _find_escape_tokens(self, text):
+        """ Find escape tokens within text, give token=None otherwise """
+        index = 0
         while text:
             a, sep1, text = text.partition(self.left)
             if a:
-                for i, t, v in self.lang.get_tokens_unprocessed(a):
-                    yield index + i, t, v
-                    index += len(a)
```
eric-wieser, line 472: This line was a bug, it should have been outside the `for` loop

pygments/formatters/latex.py
```diff
@@ -446,30 +446,62 @@ def __init__(self, left, right, lang, **options):
...
+
 
-    def get_tokens_aux(self, index, text):
+    def _find_escape_tokens(self, text):
+        """ Find escape tokens within text, give token=None otherwise """
+        index = 0
         while text:
             a, sep1, text = text.partition(self.left)
             if a:
-                for i, t, v in self.lang.get_tokens_unprocessed(a):
```
eric-wieser, line 470: This wasn't safe - the nested tokenizer needs to see all the text at once, not just be served chunks of it at a time

Anteru (changes requested):
Thanks! We have some merge conflicts due to the latest `yield from` improvements -- could you please take a look and update this as needed?

eric-wieser:
Conflicts resolved, the changes in #1537 were irrelevant

Anteru:
I did look at this and it all seems fine, but I don't quite understand the original code. Please bear with me -- I just need to convince myself it works (I'm fairly certain it does no harm :) ).

eric-wieser:
I wouldn't spend too much time understanding the old code - it had both indentation mistakes and a bad design...

Anteru:
Fair enough -- it does have tests now so I doubt it can be any worse than the original one; sorry for taking that long to merge.

PR #1519, opened 2020-08-23
Added lexer for Janet language.

uvtc:


uvtc:
[Janet syntax](https://janet-lang.org/docs/syntax.html) is lisp-like, and also has similarities with Clojure (one main difference from both is that Janet uses `#` for comments instead of `;`).

pygments/lexers/lisp.py
```diff
@@ -2692,3 +2692,148 @@ class FennelLexer(RegexLexer):
...
+
+            (r'\s+', Text),
+
+            # number
+            (r'-?\d+\.\d+', Number.Float),
+            (r'-?\d+', Number.Integer),
+            (r'0[xX][a-fA-F0-9]+', Number.Hex),
+
+            # double-quoted string
+            (r'@?"(\\\\|\\"|[^"])*"', String),
```
Anteru, line 2815: This is going to result in catastrophic backtracking (similar for the case below.) Please do two things to make sure we won't suffer from this:

* Add a test case with a long string with backslashes in a string. That should take a looong time to execute, potentially timeout.
* Change it to `[^\\"]` to avoid the backtracking (same below), the test should pass just fine now.

pygments/lexers/lisp.py
```diff
@@ -2692,3 +2692,148 @@ class FennelLexer(RegexLexer):
             (r'(\{|\})', Punctuation),
         ]
     }
+
+
+class JanetLexer(RegexLexer):
+    """Lexer for the `Janet programming language <https://janet-lang.org/>`_.
+
+    .. versionadded:: 2.7.0
```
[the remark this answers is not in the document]
uvtc: Thanks for looking at this, Anteru.

Ok. 86d2d20 .

pygments/lexers/lisp.py
uvtc: Hi Anteru. Sorry for the delay. Thank you for investigating this. Can you please tell me where to add this test case? I'm not familiar with adding tests to pygments, and also don't see a tests/test_lisp.py file.

Anteru: Just create a new file then :) The way to write a test is to use the `testcase` formatter (see also the docs), here's an example of the exact same bug just in the JsonLexer which you can use as a blueprint: https://github.com/gerner/pygments/commit/ec777784e381f29968705a7a3df9db83a2c7736f

Anteru (changes requested):
I'd like to see an unit test for the problematic string regex, because that's prone to return if left unchecked. Other than that looks good to me.

uvtc:
Hi Anteru, thanks so much for your review work, and patience here. My free time as almost completely dried up in the last few weeks, and I'm looking up the git docs now on how to proceed (my git knowledge is currently very thin!).

uvtc:
Anteru, does this apply to the regexes for both double-quoted *and* for backtick-quoted strings? Note that, unlike double-quoted strings, backtick-quoted strings may utilize multiple backticks, as in

      `some str`
     ``some str``
    ```some str```

Incidentally, both double-quoted and backtick-quoted strings can be multi-line.

Anteru:
I only applied it to double-quoted strings. Do backtick-quoted strings allow for escape characters inside? At some point it may be easier to mimic what the Python lexer does and have separate rules for handling the various escapes.

Also, no need to hurry, we just released 2.7 and 2.8 is not coming that soon :)

uvtc:
No, backtick-quoted strings do not allow for escape sequences.

uvtc:
But backtick quoted strings *do* allow for:

    ``this is `something` here``
    ```this is ``something`` here```

The language allows for even more backticks (as long as there's the same number of them starting as ending the string), but so far the most I've seen people use is

    ```three```

Anteru:
My advice would be to look at the Python lexer family and copy that, using three states -- for single, double, triple quoted backticks. Same as the string handling there -- that'll be much less hassle than any other approach IMHO.

uvtc:
Sorry I dropped the ball on this. Unable to make time to look more at this right now, and my git knowledge needs to be better so I can get this current before continuing. I'd like to return to it when I can, but if someone wants to pick up the ball and run with it that's fine with me.

PR #1525, opened 2020-08-25
Replace harcoded year in check_sources script by current year

mondeja:
The script `check_sources.py` has the current year (2020) hardcoded checking for copyrights. We can use `datetime` to create the regex dinamically.

### DIFF
--- a/scripts/check_sources.py
+++ b/scripts/check_sources.py
@@ -11,11 +11,12 @@
     :license: BSD, see LICENSE for details.
 """
 
+import datetime
+import getopt
 import io
 import os
 import re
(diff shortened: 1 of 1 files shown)


Anteru:
While this is a good idea, it also means that checking out a specific commit may not pass the checks depending on when you check it out. While it's not too comfortable to have to update this once a year, the ability to verify a commit only based on itself and not relying on external state like the date is a valuable goal IMHO.

Sorry, but I'm not going to merge this :/

mondeja:
Ah OK, it's part of the release process... don't worry.

Issue #1063, opened 2019-08-31
MySQL Variable support


Anteru:
(Original issue 1359 created by None on 2017-06-26T16:12:03.533437+00:00)

It appears that the mysql lexer doesn't include support for variables in SQL / MySQL. The '@' prefix is highlighted as an operator, rather than as part of the name.

Example: http://pygments.org/demo/1808073/

Doc: https://dev.mysql.com/doc/refman/5.7/en/user-variables.html

kurtmckee:
I'm overhauling the MySQL lexer, and improved variable support (including all variants of quoting) is already handled in my branch. PR coming after I've written automatic keyword update code a-la the Postgres builtins, as well as unit tests.

Issue #1453, opened 2020-05-18
mysql unicode 


Freed-Wu:
mysql allow user to use unicode character to create table, but in minted

```
\documentclass{ctexbook}
\usepackage{minted}
\begin{document}
\begin{minted}{mysql}
create table basic(
	股票编号 int not null,
	上市年限 int not null,
	所属行业 varchar(20) not null,
	所属概念板块 varchar(200) not null
);
\end{minted}
\end{document}
```

the result:

![image](https://user-images.githubusercontent.com/32936898/82216088-009b5c00-994b-11ea-954b-310f9dd5d85e.png)

the unicode character be considered as `PYG@token@err` which will be added a red box. is it true?

marijnschraagen:
Interestingly, it does work if there is a non-space latin character before the unicode characters:
```
\documentclass{ctexbook}
\usepackage{minted}
\begin{document}
\begin{minted}{mysql}
create table basic(
   a股票编号 int not null,
   b上市年限 int not null,
   c所属行业 varchar(20) not null,
   d所属概念板块 varchar(200) not null
);
\end{minted}
\end{document}
```
![Schermafdruk van 2020-06-23 15-58-29](https://user-images.githubusercontent.com/14089378/85412920-6abaa880-b56a-11ea-9616-a550f9379dc8.png)

Freed-Wu:
amazing!

kurtmckee:
I'm working on an overhaul of the MySQL lexer. I will _try_ to make this work, at least to the point that it doesn't render as an error.

PR #1527, opened 2020-08-29
Overhaul the MySQL lexer

### LINKED ISSUE #975: CHARACTER SET isn't recognized on MySQL lexer
(Original issue 1271 created by dereckson on 2016-07-29T00:20:08.320023+00:00)

As a sample, here the instructions for Etherpad encoding:

```
#!mysql

ALTER DATABASE `etherpad` CHARACTER SET utf8mb4 COLLATE utf8mb4_bin;
ALTER TABLE `etherpad.store` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_bin;
```

Expected behavior: CHARACTER SET and CONVERT TO CHARACTER SET receive the same highlighting

Actual behavior: CHARACTER received the class k, SET the class kt

kurtmckee:
Fixes #975, #1063, #1453

Changes include:

Documentation
-------------

* Note in the lexer docstring that Oracle MySQL is the target syntax.
  MariaDB syntax is not a target (though there is significant overlap).

Unit tests
----------

* Add 140 unit tests for MySQL.

Literals
--------

* Hexadecimal/binary/date/time/timestamp literals are supported.
* Integer mantissas are supported for scientific notation.
* In-string escapes are now tokenized properly.
* Support the "unknown" constant.

Comments
--------

* Optimizer hints are now supported, and keywords are
  recognized and tokenized as preprocessor instructions.
* Remove nested multi-line comment support, which is no
  longer supported in MySQL.

Variables
---------

* Support the '@' prefix for variable names.
* Lift restrictions on characters in unquoted variable names.
  (MySQL does not impose a restriction on lead characters.)
* Support single/double/backtick-quoted variable names, including escapes.
* Support the '@@' prefix for system variable names.
* Support '?' as a variable so people can demonstrate prepared statements.

Keywords
--------

* Keyword / data type / function are now in a separate, auto-updating file.
* Support 25 additional data types (including spatial and JSON types).
* Support 460 additional MySQL keywords.
* Support 372 MySQL functions.
  Explicit function support resolves a bug that causes non-function
  items to be treated as functions simply because they have a trailing
  opening parenthesis.
* Support exceptions for the 'SET' keyword, which is both a datatype and
  a keyword depending on context.

Schema object names
-------------------

* Support Unicode in MySQL schema object names.
* Support parsing of backtick-quoted schema object name escapes.
  (Escapes do not produce a distinct token type at this time.)

Operators
---------

* Remove non-operator characters from the list of operators.
* Remove non-punctuation characters from the list of punctuation.


Anteru:
This is a massive overhaul -- thanks a lot for the PR and the very detailed PR message. Please give us some time to review this -- looks mostly good to me, but I have only skimmed the regexes so far.

kurtmckee:
No problem.

I included an example file that demonstrates each of the changes. You should definitely look at the code, but would it be helpful to run the sample file through master and PR versions of the lexer and attach some screenshots?

Also, I references three issues in the commit message using commas to separate them but GitHub only claims one issue will be closed. Do I need separate "Fixes #xyz" lines for each issue?

Anteru:
No need to attach screenshots -- I will run it locally; and I'm not too concerned about incorrect/missing syntax as I am about a regex with catastrophic backtracking.

Re issues: You don't have to reference the issues manually, I will close them when I merge the PR and update the changelog.

Thanks again!

kurtmckee:
Ah, gotcha. If you encounter catastrophic backtracking (or have tips for what to watch out for) please let me know so I can (a) fix it, and (b) learn from it. =)

Thanks!

pygments/lexers/sql.py
```diff
@@ -595,63 +606,149 @@ class MySqlLexer(RegexLexer):
     tokens = {
         'root': [
             (r'\s+', Text),
-            (r'(#|--\s+).*\n?', Comment.Single),
-            (r'/\*', Comment.Multiline, 'multiline-comments'),
+
+            # Comments
+            (r'(?:#|--\s+).*(?=\n)?', Comment.Single),
```
birkenfeld, line 611: Does the optional lookahead assertion do anything?

pygments/lexers/sql.py
```diff
@@ -595,63 +606,149 @@ class MySqlLexer(RegexLexer):
...
+            (r'/\*\+', Comment.Special, 'optimizer-hints'),
+            (r'/\*', Comment.Multiline, 'multiline-comment'),
+
+            # Hexadecimal literals
+            (r"x'([0-9a-f]{2})+'", Number.Hex),  # MySQL requires paired hex characters in this form.
+            (r'0x[0-9a-f]+', Number.Hex),
+
+            # Binary literals
+            (r"b'[01]+'", Number.Bin),
+            (r'0b[01]+', Number.Bin),  # MySQL requires the 'b' to be lowercase.
```
birkenfeld, line 621: The case is not enforced (due to `flags = re.IGNORECASE`), but I don't think that matters...

pygments/lexers/sql.py
```diff
@@ -595,63 +606,149 @@ class MySqlLexer(RegexLexer):
...
-             r'usage|use|using|utc_date|utc_time|utc_timestamp|values|'
-             r'varying|when|where|while|with|write|x509|xor|year_month|'
-             r'zerofill)\b', Keyword),
-            # TODO: this list is not complete
-            (r'\b(auto_increment|engine|charset|tables)\b', Keyword.Pseudo),
-            (r'(true|false|null)', Name.Constant),
-            (r'([a-z_]\w*)(\s*)(\()',
+
+            # Date literals
+            (r"\{\s*d\s*(?P<quote>['\"])\s*\d{2}(\d{2})?.?\d{2}.?\d{2}\s*(?P=quote)\s*\}", Literal.Date),
```
birkenfeld, line 630: Do you need to escape the `.`s in there?  (Same for time and timestamp below.)

pygments/lexers/sql.py
```diff
@@ -595,63 +606,149 @@ class MySqlLexer(RegexLexer):
...
+                Literal.Date
+            ),
+
+            # String literals
+            (r"'", String.Single, 'single-quoted-string'),
+            (r'"', String.Double, 'double-quoted-string'),
+
+            # Variables
+            (r'@@(?:global\.|persist\.|persist_only\.|session\.)?[a-z_]+', Name.Variable),
+            (r'@[a-z0-9_$.]+', Name.Variable),
```
birkenfeld, line 651: are Unicode names allowed here?

pygments/lexers/sql.py
```diff
@@ -595,63 +606,149 @@ class MySqlLexer(RegexLexer):
...
+
+        'optimizer-hints': [
+            (r'[^*a-z]+', Comment.Special),
+            (r'\*/', Comment.Special, '#pop'),
+            (words(MYSQL_OPTIMIZER_HINTS, suffix=r'\b'), Comment.Preproc),
+            ('[a-z]+', Comment.Special),
+            (r'\*', Comment.Special),
+        ],
+
+        'multiline-comment': [
```
birkenfeld, line 695: The old version allowed nested `/* */` comments, is that not actually legal?

birkenfeld:
Thanks for the huge overhaul! Not much for me to pick on, just a few clarifications...

pygments/lexers/_mysql_builtins.py
```diff
@@ -0,0 +1,1286 @@
...
+)
+
+
+if __name__ == '__main__':  # pragma: no cover
+    import re
+
+    try:
+        from urllib import urlopen
+    except ImportError:
+        from urllib.request import urlopen
```
birkenfeld, line 1187: since we removed py2 compatibility, this can be simplified to just `urllib.request`.

pygments/lexers/_mysql_builtins.py
```diff
@@ -0,0 +1,1286 @@
...
+        for m in re.finditer(r'{"(?P<function>[^"]+?)",\s*SQL_F[^(]+?\(', f, flags=re.I):
+            results.add(m.group('function').lower())
+
+        if not results:
+            raise ValueError('No item_create functions found')
+
+        return results
+
+
+    def update_content(field_name, content):
```
birkenfeld, line 1268: (note to self: we really should have a util function that does the main bulk of this self-updating...)

pygments/lexers/sql.py
kurtmckee: I don't think so. According to [the MySQL date literal documentation](https://dev.mysql.com/doc/refman/8.0/en/date-and-time-literals.html), "Any punctuation character may be used as the delimiter between date parts." The docs don't specify the allowed characters, and testing against MySQL 8 showed that mixed punctuation like "2020@08^27" is legal.

pygments/lexers/sql.py
kurtmckee: Variables using the `@@` syntax are dynamic system variables defined by MySQL Server. [The current list of dynamic system variables](https://dev.mysql.com/doc/refman/8.0/en/dynamic-system-variables.html) appears to show that only ASCII letters and underscores are used in these types of variables.

For unquoted user variable names (those using a single `@` with no single/double/backtick quoting), [the MySQL docs](https://dev.mysql.com/doc/refman/8.0/en/user-variables.html) state that "the variable name *var_name* consists of alphanumeric characters, `.`, `_`, and `$`".

Quoted user variables should support arbitrary characters, though.

pygments/lexers/sql.py
kurtmckee: No, although it apparently was in the past and may be valid in other SQL dialects. [The MySQL docs regarding comments](https://dev.mysql.com/doc/refman/8.0/en/comments.html) state that "Nested comments are not supported, are deprecated, and will be removed in a future MySQL release."

When I tested nested comments in MySQL 8.0 I found that they are indeed unsupported. For example, typing something like this is highlighted as an error in MySQL Workbench:

image

and when I attempt to run it:

image

pygments/lexers/_mysql_builtins.py
kurtmckee: Whoops! I forgot to strip that out.

By the way, tox told me that it was skipping testing against Python 2.7, but when I installed Python 2.7 it said Pygments only supports Python >= 3.5. If I get a chance I'll submit a PR to remove Python 2.7 from the tox environments.

pygments/lexers/_mysql_builtins.py
kurtmckee: Heh, I'll let you resolve this one at your leisure. 😃

pygments/lexers/sql.py
kurtmckee: That's correct. I tried using syntax to temporarily disable `re.IGNORECASE` for the letter `b` (the syntax was something like `(?-i:b)` but this syntax is only available in Python 3.6 or higher. Hex, bin, and temporal literals all have this lowercase requirement but I forgot to remove this one comment. I'll get that fixed.

Note that I'll be using a history-rewriting force push to my branch.

pygments/lexers/sql.py
kurtmckee: It might not, but during testing it seemed to be the only thing that worked.

The original lexer code consumed a newline if it existed, and I didn't want the newline tokenized as a comment. I tried using `$` to indicate end-of-line but the `#` comment syntax started rendering as an error token. I found that the lookahead gave me the behavior I wanted without introducing errors, but there may still be a better solution for this (or perhaps I made a mistake during testing).

I'm open to suggestions!

pygments/lexers/sql.py
kurtmckee: Okay, no history rewriting, GitHub appears to only update the PR if new commits are pushed.

pygments/lexers/sql.py
kurtmckee: I've got a nagging feeling that the lookahead isn't necessary, Let me double-check this later today.

pygments/lexers/sql.py
birkenfeld: Hm, I never had problems with github picking it up when I force-pushed.

kurtmckee:
I don't recall that being an issue in the past either, but this time GitHub ignored it and suggested that I create a new pull request. When I reset to the previous SHA and then added a second commit GitHub instantly refreshed this pull request and launched the automated checks.

Maybe history rewrites are handled with a slight delay, but after a few minutes of refreshes I just went with the path of least resistance.

pygments/lexers/sql.py
kurtmckee: Yep, this was unnecessary. I removed the lookahead and expanded the single line comment tests as well.

kurtmckee:
@Anteru please close #1063 and #1453, too. These are also fixed by this patch. Thanks!

Issue #627, opened 2019-08-31
cpp-objcopy highlighting problem with arm.


Anteru:
(Original issue 920 created by john_kearney on 2013-10-19T15:03:24.162662+00:00)

The highlighter is is not working with arm disassembler.

Arm asm example.

```
#!asm

  fe028210:	e2433001 	sub	r3, r3, #1
  fe028214:	e6ef3073 	uxtb	r3, r3
  fe028218:	e3530001 	cmp	r3, #1
  fe02821c:	8a00004d 	bhi	fe028358 <procfs_read+0x210>
```

vs x86

```
#!asm

  f00a3586:	8b 58 20             	mov    0x20(%eax),%ebx
  f00a3589:	8b 2d 84 3b 0b f0    	mov    0xf00b3b84,%ebp
  f00a358f:	8b 9b 54 08 00 00    	mov    0x854(%ebx),%ebx
  f00a3595:	89 5c 24 64          	mov    %ebx,0x64(%esp)

```

changing this solves the main problem 

```
#!python

            # Code line with disassembled instructions
            ('( *)('+hex+r'+:)(\t)((?:'+hex+hex+' |'+hex+'{8,8} )+)( *\t)([a-zA-Z].*?)$',
                bygroups(Text, Name.Label, Text, Number.Hex, Text,
                         using(GasLexer))),
```
 
i.e. `((?:'+hex+hex+' |'+hex+'{8,8} )+)` instead of `((?:'+hex+hex+' ')+)`

Second issue is in the gas Lexer
    `number = r'(?:0[xX][a-fA-F0-9]+|#?\d+)'`
Added an optional #? as this is the asm syntax
e.g. #3

PR #1530, opened 2020-09-02
Add support for ARM constant notation in GAS lexer

btashton:
GNU supports ARM syntax which uses notation like `#1` for constants.
Currently this is handled like a comment which is incorrect.

Take the case of:
ldr     r0, [r1, #0]

Right now the #0] would be treated like a comment and looks very poor.
image

After this change:
image

Was referenced as part of issue #627

### DIFF
--- a/pygments/lexers/asm.py
+++ b/pygments/lexers/asm.py
@@ -37,7 +37,7 @@ class GasLexer(RegexLexer):
     string = r'"(\\"|[^"])*"'
     char = r'[\w$.@-]'
     identifier = r'(?:[a-zA-Z$_]' + char + r'*|\.' + char + '+)'
-    number = r'(?:0[xX][a-zA-Z0-9]+|\d+)'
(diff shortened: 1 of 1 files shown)


birkenfeld:
Hm, the lexer is explicitly for "AT&T" syntax, which doesn't apply to ARM assembly.  I'm not against making it more tolerant, but for example register names won't be highlighted specially either...

BTW, on ARM apparently `@` introduces a comment.

btashton:
I'm open to try to find some time to further adapt it. Or if you think a better direction would make more sense.  It is GAS syntax, which is ATT plus GNU extensions to kind of make it look like the ARM syntax so I felt like it was appropriate.

The # can have three different meanings... I think I fully capture at least you the common immediate notation, but it does not handle relocation.

https://sourceware.org/binutils/docs-2.22/as/ARM_002dChars.html#ARM_002dChars

birkenfeld:
Let's merge this now, the lexers can be improved further later.  CI failure is unrelated.

PR #1531, opened 2020-09-04
Support Ruby method names and operator overrides, including Unicode

### LINKED ISSUE #253: Ruby: Non-ASCII Method Names Not Recognised
(Original issue 474 created by None on 2010-02-07T23:12:41+00:00)

Ruby 1.9 allows method names to include non-ASCII characters with the following caveats:

* The characters must be valid in the file's source encoding.

* A legal method name that does not end with '!', '?', or '=' may have one of these characters appended.

* The ASCII punctuation characters of which operator methods consist (e.g. {{{[*%&^`~+-/\\[<>=]}}}) must not appear in any other permutation, with the exception of the above case.

Pygments does not recognise such method names, lexing the first non-ASCII character as an error. Examples of unrecognised method names are given in http://pygments.org/demo/3147/ .

Reported by guest

kurtmckee:
Fixes #253

This patch contains these changes:

* Add 50+ new Ruby tests for method name/operator override matching.
* Unicode method names are now supported (#253).
* `=` method name postfixes are now supported.
* These operator overrides are now supported:
  `<`, `<=`, `!` `!@`, `~@`, `!`, `!=`, `!~`, `=~`
* The `<>` "operator" override is removed. It appears that this was
  a typo in the regular expression (`<=?>` should have been `<=>?`).

Syntax verified with https://docs.ruby-lang.org/en/2.7.0/syntax/methods_rdoc.html

### DIFF
--- a/pygments/lexers/ruby.py
+++ b/pygments/lexers/ruby.py
@@ -328,11 +328,11 @@ class RubyLexer(ExtendedRegexLexer):
         ],
         'funcname': [
             (r'\(', Punctuation, 'defexpr'),
-            (r'(?:([a-zA-Z_]\w*)(\.))?'  # optional scope name, like "self."
+            ('(?:([a-zA-Z_]\\w*)(\\.))?'  # optional scope name, like "self."
              '('
                 '[a-zA-Z\u0080-\uffff][a-zA-Z0-9_\u0080-\uffff]*[!?=]?'  # method name
                 '|!=|!~|=~|\\*\\*?|[-+!~]@?|[/%&|^]|<=>|<[<=]?|>[>=]?|===?'  # or operator override
-                r'|\[\]=?'  # or element reference/assignment override
+                '|\\[\\]=?'  # or element reference/assignment override
                 '|`'  # or the undocumented backtick override
              ')',
              bygroups(Name.Class, Operator, Name.Function), '#pop'),
--- a/pygments/lexers/ruby.py
+++ b/pygments/lexers/ruby.py
@@ -328,9 +328,13 @@ class RubyLexer(ExtendedRegexLexer):
         ],
         'funcname': [
             (r'\(', Punctuation, 'defexpr'),
-            (r'(?:([a-zA-Z_]\w*)(\.))?'
-             r'([a-zA-Z_]\w*[!?]?|\*\*?|[-+]@?|'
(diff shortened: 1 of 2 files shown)


pygments/lexers/ruby.py
```diff
@@ -328,9 +328,13 @@ def intp_string_callback(self, match, ctx):
...
             (r'\(', Punctuation, 'defexpr'),
-            (r'(?:([a-zA-Z_]\w*)(\.))?'
-             r'([a-zA-Z_]\w*[!?]?|\*\*?|[-+]@?|'
-             r'[/%&|^`~]|\[\]=?|<<|>>|<=?>|>=?|===?)',
+            (r'(?:([a-zA-Z_]\w*)(\.))?'  # optional scope name, like "self."
+             '('
+                '[a-zA-Z\u0080-\uffff][a-zA-Z0-9_\u0080-\uffff]*[!?=]?'  # method name
+                '|!=|!~|=~|\\*\\*?|[-+!~]@?|[/%&|^]|<=>|<[<=]?|>[>=]?|===?'  # or operator override
+                r'|\[\]=?'  # or element reference/assignment override
+                '|`'  # or the undocumented backtick override
```
Anteru, line 336: Any particular reason for mixing raw and non-raw strings here? I'd prefer to have everything consistent (i.e. `r'...'`) because it makes editing less error prone.

Anteru (changes requested):

kurtmckee:
I needed to use the \u Unicode escapes, so the r prefix wasn't desirable on that line...and then I overlooked standardizing the rest of the lines.

Is it acceptable to remove the r prefix from all of the lines in the multi-line string literal block?

pygments/lexers/ruby.py
kurtmckee: I standardized on plain strings for the string block so I could still use the Unicode escapes.

birkenfeld: That's perfectly fine, but for my curiosity: Aren't Unicode escapes supported by the regex engine nowadays?

kurtmckee: You're right! Support was added in Python 3.3. I've added a commit so that the block isn't just self-consistent; it's now consistent with the rest of the surrounding raw strings. Thanks for pointing this out! 😃

Anteru:
Thanks a lot, merged!

PR #1532, opened 2020-09-04
Remove Python 2 environments from the tox environment list (py27, pypy)

kurtmckee:
This patch removes Python 2.7 and PyPY 2.7 from the tox environments list.

This appears to bring the tox env list in line with Pygments' list of target Python environments in `setup.py`.


Anteru:
Good catch, we don't build on those platforms during CI either.

PR #1533, opened 2020-09-04
Fix a Windows/PyPy3 test failure

kurtmckee:
PyPy3 on Windows has a test failure in `test_cmdline:test_outfile()`
when trying to unlink the temporary output file. The root cause is that
`cmdline:inner_main()` does not explicitly close the file that it opens,
and PyPy3 isn't auto-closing the file when `inner_main()` returns.
This prevents the file from being unlinked, and the test case fails.

### DIFF
--- a/pygments/cmdline.py
+++ b/pygments/cmdline.py
@@ -515,7 +515,11 @@ def main_inner(popts, args, usage):
     # ... and do it!
     if '-s' not in opts:
         # process whole input as per normal...
-        highlight(code, lexer, fmter, outfile)
+        try:
+            highlight(code, lexer, fmter, outfile)
+        finally:
+            if outfn:
+                outfile.close()
         return 0
     else:
         # line by line processing of stdin (eg: for 'tail -f')...
@@ -532,6 +536,9 @@ def main_inner(popts, args, usage):
             return 0
         except KeyboardInterrupt:  # pragma: no cover
             return 0
+        finally:
+            if outfn:
+                outfile.close()
 
 
 def main(args=sys.argv):


Anteru:
Can you please shed some light on when this triggers? What do I need to reproduce this error?

kurtmckee:
Here are the commands to run to reproduce this. Application versions:

* Windows 10
* Python 3.8.5 (installed to `C:\Program Files\Python38`)
* PyPy 7.3.1 (extracted to `C:\Program Files (x86)\pypy36`)

Your paths to Python 3.8 and PyPy3 may vary.

```batch
git clone https://github.com/pygments/pygments.git demo-1533
cd demo-1533
"C:\Program Files\Python38\python.exe" -m venv venv
venv\Scripts\activate.bat
python -m pip install tox
set "PATH=C:\Program Files (x86)\pypy36;%PATH%"
tox -e pypy3
```

Everything runs to completion except for `test_cmdline:test_outfile()`:

```
================================== FAILURES ===================================
________________________________ test_outfile _________________________________

    def test_outfile():
        # test that output file works with and without encoding
        fd, name = tempfile.mkstemp()
        os.close(fd)
        for opts in [['-fhtml', '-o', name, TESTFILE],
                     ['-flatex', '-o', name, TESTFILE],
                     ['-fhtml', '-o', name, '-O', 'encoding=utf-8', TESTFILE]]:
            try:
                check_success(*opts)
            finally:
>               os.unlink(name)
E               PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\Kurt\\AppData\\Local\\Temp\\tmp2vsf0gyx'

tests\test_cmdline.py:115: PermissionError
=========================== short test summary info ===========================
FAILED tests/test_cmdline.py::test_outfile - PermissionError: [WinError 32] T...
============ 1 failed, 3099 passed, 9 skipped in 167.43s (0:02:47) ============
ERROR: InvocationError for command 'C:\Users\Kurt\Documents\dev\demo-1533\.tox\pypy3\Scripts\py.test.EXE' (exited with code 1)
___________________________________ summary ___________________________________
ERROR:   pypy3: commands failed
```

Closing the files as shown in the patch resolves the problem.

birkenfeld:
LGTM for now.  The `cmdline` code is in desperate need of a rewrite, though.  (Together with switching to a modern argument parser lib.)

Anteru:
> 
> 
> LGTM for now. The `cmdline` code is in desperate need of a rewrite, though. (Together with switching to a modern argument parser lib.)

Yes, I'm aware :) Let's get 2.7 out of the door and then I can look into that.

PR #1534, opened 2020-09-05
Overhaul Javascript numeric literals

### LINKED ISSUE #307: Javascript lexer missparse floating point numbers without dot
(Original issue 528 created by None on 2010-08-30T17:10:29+00:00)

Numbers, which are floating point, but do not include a dot, will not be parsed correctly by the Javascript lexer.

Example:

1e5, 1e+5, 12E-2

Fix:

change

   (r'[0-9][0-9]*\\.[0-9]+([eE][0-9]+)?[fd]?', Number.Float),

to

   (r'[0-9][0-9]*\\.[0-9]+([eE][+-]?[0-9]+)?[fd]?|[0-9][0-9]*([eE][+-]?[0-9]+)[fd]?', Number.Float),

or something similar

in lexers/web.py

Reported by thoka

kurtmckee:
This patch contains the following changes:

* Adds 50+ unit tests for Javascript numeric literals
* Forces ASCII numbers for float literals (so, now reject `.୪`)
* Adds support for Javascript's BigInt notation (`100n`)
* Adds support for leading-zero-only octal notation (`0777`)
* Adds support for scientific notation with no significand (`1e10`, fixes #307)

### DIFF
--- a/pygments/lexers/javascript.py
+++ b/pygments/lexers/javascript.py
@@ -64,11 +64,14 @@ class JavascriptLexer(RegexLexer):
             (r'\A#! ?/.*?\n', Comment.Hashbang),  # recognized by node.js
             (r'^(?=\s|/|<!--)', Text, 'slashstartsregex'),
             include('commentsandwhitespace'),
-            (r'(\.\d+|[0-9]+\.[0-9]*)([eE][-+]?[0-9]+)?', Number.Float),
-            (r'0[bB][01]+', Number.Bin),
-            (r'0[oO][0-7]+', Number.Oct),
-            (r'0[xX][0-9a-fA-F]+', Number.Hex),
-            (r'[0-9]+', Number.Integer),
+
+            # Numeric literals
+            (r'0[bB][01]+n?', Number.Bin),
+            (r'0[oO]?[0-7]+n?', Number.Oct),  # Browsers support "0o7" and "07" notations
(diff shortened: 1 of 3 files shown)


pygments/lexers/javascript.py
```diff
@@ -64,11 +64,14 @@ class JavascriptLexer(RegexLexer):
...
-            (r'0[bB][01]+', Number.Bin),
-            (r'0[oO][0-7]+', Number.Oct),
-            (r'0[xX][0-9a-fA-F]+', Number.Hex),
-            (r'[0-9]+', Number.Integer),
+
+            # Numeric literals
+            (r'0[bB][01]+n?', Number.Bin),
+            (r'0[oO]?[0-7]+n?', Number.Oct),  # Browsers support "0o7" and "07" notations
+            (r'0[xX][0-9a-fA-F]+n?', Number.Hex),
+            (r'[0-9]+n', Number.Integer),  # Javascript BigInt requires an "n" postfix
```
birkenfeld, line 72: Maybe add a comment that other integer-looking numbers are covered by the float regex below? It threw me off until I remembered that JavaScript doesn't really have a simple integer type.

birkenfeld (approved):
Just a small nit.

pygments/lexers/javascript.py
Anteru: I'll add a comment after merging this.

Anteru:
Thanks for the contribution!

PR #1535, opened 2020-09-05
Update Fennel keywords to catch up to version 0.6.0.

technomancy:
The keywords in the current Fennel lexer are a bit behind; this catches it up to the latest release.

Also removes some disallowed identifier characters we accidentally left in previously.

### DIFF
--- a/pygments/lexers/lisp.py
+++ b/pygments/lexers/lisp.py
@@ -2638,27 +2638,27 @@ class FennelLexer(RegexLexer):
     # https://gitlab.com/technomancy/fennel-mode
     # this list is current as of Fennel version 0.6.0.
     special_forms = (
-        u'require-macros', u'eval-compiler', u'doc', u'lua', u'hashfn',
-        u'macro', u'macros', u'import-macros', u'pick-args', u'pick-values',
-        u'macroexpand', u'macrodebug', u'do', u'values', u'if', u'when',
-        u'each', u'for', u'fn', u'lambda', u'λ', u'partial', u'while',
-        u'set', u'global', u'var', u'local', u'let', u'tset', u'set-forcibly!',
-        u'doto', u'match', u'or', u'and', u'true', u'false', u'nil', u'not',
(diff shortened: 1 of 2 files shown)


birkenfeld (approved):
JFYI, the `u` prefix can be omitted now that we're py3-only.

birkenfeld:
However, the example file doesn't seem to like your changes?  Can you verify that https://github.com/pygments/pygments/blob/master/tests/examplefiles/fennelview.fnl is still valid?

technomancy:
Thanks! Sorry for the noise; I think this should take care of it now. I had forgotten the previous patch was from such an old version, so the example needed to be brought up to date.

pygments/lexers/lisp.py
```diff
@@ -2670,7 +2675,6 @@ class FennelLexer(RegexLexer):
             (r'-?\d+', Number.Integer),
 
             (r'"(\\\\|\\"|[^"])*"', String),
```
birkenfeld, line 2677: Sorry, I just saw this right now: this regex is prone to catastrophic backtracking, we should fix it while at it.  It should look like 
```
"(\\\\|\\"|[^"\\])*"
```
with alterations as needed for other allowed escape sequences.  If they are complex, a separate state can be used as well.

birkenfeld (approved):
The changes themselves look good now, we just should resolve the string regex issue.

birkenfeld (approved):

birkenfeld:
I made another small fix to account for other allowed escapes.

technomancy:
Thanks! As you can probably tell, I've never written Python before so I appreciate the cleanup.

birkenfeld:
Thanks even more for the contribution then!

pygments/lexers/lisp.py
Anteru: That's literally the third time in a week or so we got the same regex of doom :)

technomancy: Oh, well really that regex was added years ago; copied from an existing lexer IIRC. But I'm glad to see it gone.

PR #1536, opened 2020-09-05
all: remove "u" string prefix

birkenfeld:


birkenfeld:
Also removes support for "narrow unicode builds" which are not relevant after Py3.3 / PEP 393.

Anteru (approved):
I presume this was done using some regex. I glanced over the diff, I didn't spot anything concerning so good to merge.

birkenfeld:
Hand-assisted regex :)  wouldn't be sure about consequences otherwise.

PR #1537, opened 2020-09-05
all: use yield from

birkenfeld:


pygments/lexers/scripting.py
```diff
@@ -1253,13 +1253,13 @@ class MiniScriptLexer(RegexLexer):
...
-            	'sort', 'shuffle', 'push', 'pop', 'pull', 'range',
-            	'print', 'input', 'time', 'wait', 'locals', 'globals', 'outer',
-            	'yield'), suffix=r'\b'),
+                'abs', 'acos', 'asin', 'atan', 'ceil', 'char', 'cos', 'floor',
+                'log', 'round', 'rnd', 'pi', 'sign', 'sin', 'sqrt', 'str', 'tan',
+                'hasIndex', 'indexOf', 'len', 'val', 'code', 'remove', 'lower',
+                'upper', 'replace', 'split', 'indexes', 'values', 'join', 'sum',
+                'sort', 'shuffle', 'push', 'pop', 'pull', 'range',
+                'print', 'input', 'time', 'wait', 'locals', 'globals', 'outer',
+                'yield'), suffix=r'\b'),
```
Anteru, line 1262: I assume this is a drive-by shooting that the formatting changed here?

pygments/lexers/sql.py
```diff
@@ -117,9 +117,7 @@ class PostgresBase:
     def get_tokens_unprocessed(self, text, *args):
         # Have a copy of the entire text to be used by `language_callback`.
         self.text = text
-        for x in super(PostgresBase, self).get_tokens_unprocessed(
-                text, *args):
-            yield x
+        yield from super(PostgresBase, self).get_tokens_unprocessed(text, *args)
```
Anteru, line 120: `yield from super().`?

pygments/lexers/scripting.py
birkenfeld: yes

pygments/lexers/sql.py
birkenfeld: next PR :)

pygments/lexers/zig.py
```diff
@@ -122,8 +122,3 @@ class ZigLexer(RegexLexer):
             (r'"', String, '#pop')
         ]
     }
-
-    def get_tokens_unprocessed(self, text):
-        for index, token, value in \
-                RegexLexer.get_tokens_unprocessed(self, text):
-            yield index, token, value
```
Anteru, line 129: Why was this removed wholesale? I presume because that's equivalent to the parent class implementation?

birkenfeld: Since the parent class is RegexLexer, AFACIS it just delegates everything to it.

Anteru (approved):
Looks good to me, but I do have questions.

PR #1538, opened 2020-09-05
all: use argumentless super()

birkenfeld:


pygments/lexers/parsers.py
```diff
@@ -534,7 +529,7 @@ def analyse_text(text):
 #    filenames = ['*.G', '*.g']
 #
 #    def __init__(self, **options):
-#        super(AntlrCLexer, self).__init__(CLexer, AntlrLexer, **options)
+#        super().__init__(CLexer, AntlrLexer, **options)
```
Anteru, line 532: Given this is all commented out, maybe better get rid of this wholesale?

Anteru (changes requested):
One minor remark only, otherwise fine.

pygments/lexers/parsers.py
birkenfeld: Yeah, sure.

PR #1541, opened 2020-09-08
Support for RITA language

zaibacu:
Introduces Lexer for [RITA](https://github.com/zaibacu/rita-dsl)

### DIFF
--- a/pygments/lexers/rita.py
+++ b/pygments/lexers/rita.py
@@ -19,6 +19,10 @@ __all__ = ['RitaLexer']
 
 
 class RitaLexer(RegexLexer):
+    """
+    Lexer for `RITA <https://github.com/zaibacu/rita-dsl>`_
+     .. versionadded:: 2.7
+    """
     name = 'Rita'
     filenames = ['*.rita']
     aliases = ['rita']
@@ -31,11 +35,11 @@ class RitaLexer(RegexLexer):
             (r'\n', Text),
             (r'\s+', Text),
             (r'#(.*?)\n', Comment.Single),
-            (r'@(.*?)\n', Comment.Single),
+            (r'@(.*?)\n', Operator),  # Yes, whole line as an operator
             (r'(\"|\')(\w|\d|[^\"\']|(\\\")|(\\\'))+?(\"|\')', Literal),
             (r'([A-Z_]+)', Keyword),
             (r'([a-z_]+)', Name),
-            (r'((->)|[!?+*|])', Operator),
+            (r'((->)|[!?+*|=])', Operator),
             (r'[\(\),\{\}]', Punctuation)
         ]
     }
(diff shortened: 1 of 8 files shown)


pygments/lexers/rita.py
```diff
@@ -0,0 +1,46 @@
...
+
+    flags = re.MULTILINE | re.UNICODE
+
+    tokens = {
+        'root': [
+            (r'\n', Text),
+            (r'\s+', Text),
+            (r'#(.*?)\n', Comment.Single),
+            (r'@(.*?)\n', Operator),  # Yes, whole line as an operator
+            (r'(\"|\')(\w|\d|[^\"\']|(\\\")|(\\\'))+?(\"|\')', Literal),
```
birkenfeld, line 39: It's probably better to split this into separate rules for single and double quoted strings; right now you can do `"blah'`...

Also, no empty strings?

pygments/lexers/rita.py
```diff
@@ -0,0 +1,46 @@
...
+    inherit, words
+from pygments.token import Comment, Operator, Keyword, Name, Literal, Punctuation, Text
+
+__all__ = ['RitaLexer']
+
+
+class RitaLexer(RegexLexer):
+    """
+    Lexer for `RITA <https://github.com/zaibacu/rita-dsl>`_
+     .. versionadded:: 2.7
```
birkenfeld, line 24: rst requires an empty line before the versionadded directive.

pygments/lexers/rita.py
```diff
@@ -0,0 +1,46 @@
...
+
+    tokens = {
+        'root': [
+            (r'\n', Text),
+            (r'\s+', Text),
+            (r'#(.*?)\n', Comment.Single),
+            (r'@(.*?)\n', Operator),  # Yes, whole line as an operator
+            (r'(\"|\')(\w|\d|[^\"\']|(\\\")|(\\\'))+?(\"|\')', Literal),
+            (r'([A-Z_]+)', Keyword),
+            (r'([a-z_]+)', Name),
```
birkenfeld, line 41: No digits allowed in names?

birkenfeld (commented):
Thanks for the PR! I've commented a few things to discuss.

pygments/lexers/rita.py
zaibacu: Thank you, that's a fair point and a nice way to solve it

zaibacu:
Some times one of the builds freezes. Any ideas?

Anteru:
I have to guess but I believe it's a randomized test hitting a regex which causes catastrophic backtracking.

zaibacu:
> I have to guess but I believe it's a randomized test hitting a regex which causes catastrophic backtracking. Can you try to run it locally with the same random seed: `--randomly-seed=2266720041` and see if you get the same hang?

Yes, thank you, can reproduce it with this seed and indeed it crashes on `RitaLexer`. Looking how to fix regex so it would not hang

zaibacu:
Bump :) this one is still not merged

PR #1546, opened 2020-09-17
:pencil2: : Added much more python versions

ThatXliner:
Your original workflow only covered a tiny amount of python versions. My proposal is to cover *all* the versions you support, including the minor versions. It may seem like a waste of time, but it does give people more confidence.


Anteru:
@birkenfeld That seems like serious overkill to me -- what do you think?

ThatXliner:
Hmm, maybe.

birkenfeld:
No, sorry, latest of each minor is fine.  We don't need the additional work to keep track of each patch version, and neither the false-negatives, see Windows on 3.6.12 in this run.

ThatXliner:
👌

PR #1547, opened 2020-09-18
Add a check for CR/LF in files.

Anteru:
This can occur when checking out things on Windows, and it breaks the
tarball. This adds a script to check for the presence of CR/LF which
exits early if anything gets found.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ export PYTHONPATH = $(shell echo "$$PYTHONPATH"):$(shell python -c 'import os; p
 all: clean-pyc check test
 
 check:
-	@$(PYTHON) scripts/check_crlf.py pygments build external || true
+	@$(PYTHON) scripts/check_crlf.py pygments build external
 	@$(PYTHON) scripts/detect_missing_analyse_text.py || true
 	@pyflakes pygments | grep -v 'but unused' || true
 	@$(PYTHON) scripts/check_sources.py -i build -i dist -i pygments/lexers/_mapping.py \
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,7 @@ export PYTHONPATH = $(shell echo "$$PYTHONPATH"):$(shell python -c 'import os; p
 all: clean-pyc check test
 
 check:
+	@$(PYTHON) scripts/check_crlf.py pygments dist || true
 	@$(PYTHON) scripts/detect_missing_analyse_text.py || true
 	@pyflakes pygments | grep -v 'but unused' || true
 	@$(PYTHON) scripts/check_sources.py -i build -i dist -i pygments/lexers/_mapping.py \
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ export PYTHONPATH = $(shell echo "$$PYTHONPATH"):$(shell python -c 'import os; p
 all: clean-pyc check test
 
 check:
-	@$(PYTHON) scripts/check_crlf.py pygments dist || true
+	@$(PYTHON) scripts/check_crlf.py pygments build external || true
 	@$(PYTHON) scripts/detect_missing_analyse_text.py || true
 	@pyflakes pygments | grep -v 'but unused' || true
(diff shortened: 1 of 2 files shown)


scripts/check_crlf.py
```diff
@@ -0,0 +1,30 @@
...
+import os
+
+if __name__ == '__main__':
+    for directory in sys.argv[1:]:
+        if not os.path.exists(directory):
+            continue
+
+        for root, dirs, files in os.walk(directory):
+            for filename in files:
+                if not filename.endswith('.py'):
```
birkenfeld, line 23: hm, this is only checking .py files but the problem was in the bashcomp file...

Makefile
```diff
@@ -19,6 +19,7 @@ export PYTHONPATH = $(shell echo "$$PYTHONPATH"):$(shell python -c 'import os; p
 all: clean-pyc check test
 
 check:
+	@$(PYTHON) scripts/check_crlf.py pygments dist || true
```
birkenfeld, line 22: why `dist`? It contains only tarballs/wheels...

birkenfeld (changes requested):

scripts/check_crlf.py
Anteru: As far as I can tell if I mess up the CR/LF I mess it up everywhere. I can exclude `*.pyc`, I thought this was fairly safe for the particular issue at hand.

Anteru: Added `.bashcomp`.

Makefile
Anteru: Good catch, that should have been build. I've also added `external`.

scripts/check_crlf.py
birkenfeld: Makes sense.

Anteru:
Good to go? I'd like to merge this followed by #1549 , then I just need to fix too long lines and we could turn on `make check` for CI.

scripts/check_crlf.py
```diff
@@ -0,0 +1,30 @@
...
+            continue
+
+        for root, dirs, files in os.walk(directory):
+            for filename in files:
+                if not filename.endswith('.py') and not filename.endswith('.bashcomp'):
+                    continue
+
+                with open(os.path.join(root, filename), 'rb') as f:
+                    if b'\r\n' in f.read():
+                        sys.exit(1)
```
birkenfeld, line 28: do you want to output the name of the offending file?

Makefile
```diff
@@ -19,6 +19,7 @@ export PYTHONPATH = $(shell echo "$$PYTHONPATH"):$(shell python -c 'import os; p
 all: clean-pyc check test
 
 check:
+	@$(PYTHON) scripts/check_crlf.py pygments build external || true
```
birkenfeld, line 22: `|| true` can be omitted since this shouldn't complain...

scripts/check_crlf.py
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+    Checker for line endings
+    ~~~~~~~~~~~~~~~~~~~~~~~~
+
+    Make sure each Python does not use Windows-style file endings.
```
birkenfeld, line 7: `each Python file` probably?

scripts/check_crlf.py
Anteru: Yes, good call!

scripts/check_crlf.py
Anteru: So someone does read the docs! Thanks, good catch.

Makefile
Anteru: I'm not familiar with that Makefile mechanism to start with, will remove it :)

birkenfeld: It's shell syntax - `|| true` will ignore the exit status of the script and exit with success.  It's used for the other invocations because they are (at the moment) producing errors.

Once you fix all the warnings you can remove this for all the other lines so that they actually do complain if something is wrong :)

Anteru: Ah, thanks, that makes sense!

birkenfeld (approved):

PR #1549, opened 2020-09-19
Add analyze_text to make make check happy.

Anteru:
This also fixes a few small bugs:

* Slash uses *.sla as the file ending, not *.sl
* IDL has endelse, not elseelse

### DIFF
--- a/pygments/lexers/configs.py
+++ b/pygments/lexers/configs.py
@@ -974,5 +974,11 @@ class SingularityLexer(RegexLexer):
     def analyse_text(text):
         """This is a quite simple script file, but there are a few keywords
         which seem unique to this language."""
-        if re.match('osversion|includecmd', text, re.IGNORECASE):
-            return 0.1
+        result = 0
+        if re.search(r'\b(?:osversion|includecmd|mirrorurl)\b', text, re.IGNORECASE):
+            result += 0.5
+
+        if re.search(SingularityLexer._section[1:], text):
+            result += 0.49
+
+        return result
--- a/pygments/lexers/configs.py
+++ b/pygments/lexers/configs.py
@@ -970,3 +970,9 @@ class SingularityLexer(RegexLexer):
             (r'(.+?(?=^\s*%))|(.*)', using(BashLexer), '#pop'),
         ],
     }
+
+    def analyse_text(text):
+        """This is a quite simple script file, but there are a few keywords
+        which seem unique to this language."""
+        if re.match('osversion|includecmd', text, re.IGNORECASE):
+            return 0.1
--- a/pygments/lexers/dotnet.py
+++ b/pygments/lexers/dotnet.py
@@ -696,16 +696,12 @@ class FSharpLexer(RegexLexer):
     }
 
     def analyse_text(text):
-        """F# doesn't have that many unique features -- |> for matching
-        seems quite common though, in addition to let/match."""
+        """F# doesn't have that many unique features -- |> and <| are weak
+        indicators."""
         result = 0
-        if 'let' in text:
-            result += 0.05
-        if 'match' in text:
-            result += 0.05
-        if '->' in text:
-            result += 0.01
         if '|>' in text:
+            result += 0.05
+        if '<|' in text:
             result += 0.05
 
         return result
--- a/pygments/lexers/dotnet.py
+++ b/pygments/lexers/dotnet.py
@@ -295,6 +295,16 @@ class NemerleLexer(RegexLexer):
 
         RegexLexer.__init__(self, **options)
 
+    def analyse_text(text):
+        """Nemerle is quite similar to Python, but @if is relatively uncommon
+        elsewhere."""
+        result = 0
+
+        if '@if' in text:
+            result += 0.1
+
+        return result
+
 
 class BooLexer(RegexLexer):
     """
@@ -684,3 +694,18 @@ class FSharpLexer(RegexLexer):
             (r'"', String),
         ],
     }
+
+    def analyse_text(text):
+        """F# doesn't have that many unique features -- |> for matching
+        seems quite common though, in addition to let/match."""
+        result = 0
+        if 'let' in text:
+            result += 0.05
+        if 'match' in text:
+            result += 0.05
+        if '->' in text:
+            result += 0.01
+        if '|>' in text:
+            result += 0.05
+
+        return result
--- a/pygments/lexers/esoteric.py
+++ b/pygments/lexers/esoteric.py
@@ -55,16 +55,25 @@ class BrainfuckLexer(RegexLexer):
         and < > is brainfuck."""
         plus_minus_count = 0
         greater_less_count = 0
-        for c in text:
+
+        range_to_check = max(256, len(text))
+
+        for c in text[:range_to_check]:
             if c == '+' or c == '-':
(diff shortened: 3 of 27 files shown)


pygments/lexers/dotnet.py
```diff
@@ -684,3 +694,18 @@ class FSharpLexer(RegexLexer):
             (r'"', String),
         ],
     }
+
+    def analyse_text(text):
+        """F# doesn't have that many unique features -- |> for matching
+        seems quite common though, in addition to let/match."""
```
birkenfeld, line 700: let/match/-> are quite common in functional and derived languages, e.g. Haskell or Rust.  I wouldn't use them.

pygments/lexers/esoteric.py
```diff
@@ -48,6 +48,15 @@ class BrainfuckLexer(RegexLexer):
         ]
     }
 
+    def analyse_text(text):
+        """It's safe to assume that any reasonable brainfuck program will have
+        more than two consecutive + signs, which is very uncommon in any
+        other programming language."""
+        if '+++++' in text or '------' in text:
+            return 0.5
+        if '+++' in text:
```
birkenfeld, line 57: `+++` and `---` are common in diff/patch files, which are probably more common than brainfuck :)

The longer `+++++++` and `------` sequences are typical for underlines in readable markup like rst/markdown.

Another signature of brainfuck which has less conflicts is probably `[-]`, i.e. "clear this cell".

pygments/lexers/int_fiction.py
```diff
@@ -516,6 +516,15 @@ def get_tokens_unprocessed(self, text):
         while objectloop_queue:
             yield objectloop_queue.pop(0)
 
+    def analyse_text(text):
+        """We try to find a keyword which seem relatively common, unfortunately
+        there is a decent overlap with Smalltalk keywords otherwise here.."""
+        result = 0
+        if re.match('origsource', text, re.IGNORECASE):
```
birkenfeld, line 523: `match` vs `search` again?

pygments/lexers/configs.py
```diff
@@ -970,3 +970,9 @@ class SingularityLexer(RegexLexer):
             (r'(.+?(?=^\s*%))|(.*)', using(BashLexer), '#pop'),
         ],
     }
+
+    def analyse_text(text):
+        """This is a quite simple script file, but there are a few keywords
+        which seem unique to this language."""
+        if re.match('osversion|includecmd', text, re.IGNORECASE):
```
birkenfeld, line 977: `match` is anchored at the beginning of the text, is that intended?

pygments/lexers/matlab.py
```diff
@@ -647,6 +647,15 @@ class OctaveLexer(RegexLexer):
         ],
     }
 
+    def analyse_text(text):
+        """Octave is quite hard to spot from a casual glance, but we'll assume
+        that most examples will have an integer vector or matrix somewhere."""
+        # A \ B is actually quite uncommon outside of Matlab/Octave
+        if re.match(r'\w+\s*\\\s*\w+',text):
+            return 0.05
+        if re.match(r'\[\s*(?:(?:\d+\s*)+;\s*)+\s*\]', text):
```
birkenfeld, line 656: This is a quite complicated regex that might backtrack (not catastrophically, but still a bit much to do for analyse_text...)

pygments/lexers/modula2.py
```diff
@@ -1559,3 +1559,13 @@ def get_tokens_unprocessed(self, text):
 
             # return result
             yield index, token, value
+
+    def analyse_text(text):
+        """Not much we can go by. (* for comments is our best guess."""
```
birkenfeld, line 1564: that's the same for all pascal-like languages, unfortunately.

pygments/lexers/scripting.py
```diff
@@ -944,6 +944,15 @@ class HybrisLexer(RegexLexer):
         ],
     }
 
+    def analyse_text(text):
+        """public method and private method don't seem to be quite common
+        elsewhere."""
+        result = 0
+        if re.match(r'(?:public|private)\s+method', text):
```
birkenfeld, line 951: `match`

birkenfeld (changes requested):
In general I'd prefer for searched-for keywords to be bounded by a `\b` word delimiter.

pygments/lexers/configs.py
Anteru: Nope, and I'll fix all remaining instances as well.

pygments/lexers/matlab.py
Anteru: I just tried those out and both aren't really good indicators either. I'm inclined to `return 0` until someone has a better idea.

pygments/lexers/esoteric.py
Anteru: I've changed it to check if 25% of the source consist of +- or <>, that seems fairly unlikely in any other source (even HTML is not that <> heavy.)

birkenfeld: While this is probably a good test, I'm worrying about the performance implications of looping over the string on the Python level... the other tests at least use the C-based regex/stringlib routines.

Anteru: Given it's statistics, maybe trying it on the first 256 characters or so is enough? The character distribution for brainfuck should be similar across any subset of the program.

pygments/lexers/scripting.py
Anteru: Done

pygments/lexers/modula2.py
Anteru: Ok. New plan: Figure out if it's Pascal-Like first, then check if `PROCEDURE` is present (not valid Pascal), and back out if `FUNCTION` is found (not valid Modula2)

pygments/lexers/ezhil.py
```diff
@@ -69,7 +69,7 @@ def analyse_text(text):
         decent amount of Tamil-characters, it's this language. This assumption
         is obviously horribly off if someone uses string literals in tamil
         in another language."""
-        if len(re.findall('[\u0b80-\u0bff]')) > 10:
+        if len(re.findall(r'[\u0b80-\u0bff]', text)) > 10:
```
birkenfeld, line 72: Oops, I missed that.

pygments/lexers/modula2.py
```diff
@@ -1563,9 +1563,20 @@ def get_tokens_unprocessed(self, text):
...
+
+        is_pascal_like = 0
         if '(*' in text and '*)' in text:
-            result += 0.01
+            is_pascal_like += 0.5
         if ':=' in text:
-            result += 0.01
+            is_pascal_like += 0.5
+
+        if is_pascal_like == 1:
```
birkenfeld, line 1573: Looks good, but can you make this a boolean flag (`pascal = x in text and y in text and z in text`)?  The float-adding-and-comparing seems to be unneeded...

birkenfeld (approved):

pygments/lexers/ezhil.py
Anteru: Me too, but that's what ~friends~ tests are for :)

pygments/lexers/modula2.py
Anteru: Sure!

PR #1554, opened 2020-09-27
Add Python 3.9 as a supported version

kurtmckee:
This patch adds Python 3.9 as a tested tox version and lists it as a trove classifier in `setup.py`.

I didn't try to add it as a GitHub workflow because I'm not sure if Python 3.9 is supported yet. That piece still needs to be done.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,6 +55,7 @@ setup(
         'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: 3.8',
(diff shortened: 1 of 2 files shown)


Anteru:
Thanks! Python 3.9 is not out officially, is it?

kurtmckee:
Not yet, but they've released a final release candidate and tox is able to find and test against it when it's installed.

Anteru:
Python 3.9 was released, time to merge this :)

PR #1555, opened 2020-09-27
MySQL: Tokenize quoted schema object names, and escape characters, uniquely

### LINKED ISSUE #1551: MySQL ` (backtick) highlighting
Hello,
I am trying to highlighting everything between backticks exactly like in MySQL Workbench:
mysql workbench

# Code
This is my style and example sql
```python
# -*- coding: utf-8 -*-

from pygments.style import Style
from pygments.token import Keyword, Name, Comment, String, Error, \
    Number, Operator, Punctuation, Generic, Whitespace, Text, Literal, Punctuation


class WorkbenchStyle(Style):
    styles = {
        Whitespace: '#a89028',

        Text: '#000000',
        Punctuation: '#000000',

        Comment.Single: '#0987cb',
        Comment.Special: '#0987cb',
        Comment.Multiline: '#0987cb',
        Comment.Preproc: '#0987cb',

        Number.Hex: '#cc6c00',
        Number.Bin: '#cc6c00',
        Number.Float: '#cc6c00',
        Number.Integer: '#cc6c00',

        Literal.Date: '#cc6c00',

        String.Single: '#dd7a00',
        String.Double: '#dd7a00',
        String.Escape: '#dd7a00',

        Name: '#993a3e', # <<<<<<<<<<<<<
        Name.Variable: '#000000',
        Name.Constant: 'bold #007FBF',
        Name.Function: '#7d7d63',

        Operator: '#000000',

        Keyword: 'bold #007FBF',
        Keyword.Type: 'bold #007FBF',
    }
```
```mysql
SELECT MIN(date) FROM medw
WHERE afd = (SELECT anr FROM afd WHERE name = 'Verkoop\%');
/**
abc
\%
*/
-- abc
# lollll lol
create user bob@localhost identified by 'Secure1pass!';
SELECT * FROM abc GROUP BY a;
use bobdb;
PREPARE stmt1 FROM 'SELECT SQRT(POW(?,2) + POW(?,2)) AS hypotenuse';
SET @a = 3;
SET @b = 4; ?
EXECUTE stmt1 USING @a, @b;
SELECT * FROM abc WHERE x = ' %s a' OR a = 0xA111 or c = 0b0011
AND x IS NULL OR c = true
AND f = '1983-09-05 13:28:00' OR `xbc`.`a` >= 567;
'x'
SELECT /*+ MAX_EXECUTION_TIME(1000) */ * FROM t1
PROCEDURE
MAX_EXECUTION_TIME(3)
CREATE TABLE IF NOT EXISTS tasks (
    task_id INT AUTO_INCREMENT PRIMARY KEY,
    title VARCHAR(255) NOT NULL,
    start_date DATE,
    due_date DATE,
    status TINYINT NOT NULL,
    priority TINYINT NOT NULL,
    description TEXT,
    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP
)  ENGINE=INNODB;
```

I compile everything using
```
pygmentize -f html -l mysql -O style=workbench -O noclasses=True -o abc.html test.sql
```

# My attempts at solving the issue
The problem is that `Name: '#993a3e'` makes everything red:
comparison

I Have tried solving it with `noinherit`, but alas

Backtick identification seems to be happening [here](https://github.com/pygments/pygments/blob/master/pygments/lexers/sql.py#L676), [here](https://github.com/pygments/pygments/blob/master/pygments/lexers/sql.py#L731) or [here](https://github.com/pygments/pygments/blob/master/pygments/lexers/sql.py#L745). A solution would be appreciated

kurtmckee:
Changes in this patch:

* Name.Quoted and Name.Quoted.Escape are introduced as non-standard tokens
* HTML and LaTeX formatters were confirmed to provide default formatting
  if they encounter these two non-standard tokens. They also add style
  classes based on the token name, like "n-Quoted" (HTML) or "nQuoted"
  (LaTeX) so that users can add custom styles for these.
* Removed "\`" and "\\" as schema object name escapes. These are relics
  of the previous regular expression for backtick-quoted names and are
  not treated as escape sequences. The behavior was confirmed in the
  MySQL documentation as well as by running queries in MySQL Workbench.
* Prevent "123abc" from being treated as an integer followed by a schema
  object name. MySQL allows leading numbers in schema object names as long
  as 0-9 are not the only characters in the schema object name.
* Add ~10 more unit tests to validate behavior.

Closes #1551

### DIFF
--- a/pygments/lexers/sql.py
+++ b/pygments/lexers/sql.py
@@ -620,7 +620,7 @@ class MySqlLexer(RegexLexer):
             (r'[0-9]+\.[0-9]*(e[+-]?[0-9]+)?', Number.Float),  # Mandatory integer, optional fraction and exponent
             (r'[0-9]*\.[0-9]+(e[+-]?[0-9]+)?', Number.Float),  # Mandatory fraction, optional integer and exponent
             (r'[0-9]+e[+-]?[0-9]+', Number.Float),  # Exponents with integer significands are still floats
-            (r'[0-9]+(?=[^0-9a-z$_\u0080-\uffff]|$)', Number.Integer),  # Integers that are not in a schema object name
+            (r'[0-9]+(?=[^0-9a-z$_\u0080-\uffff])', Number.Integer),  # Integers that are not in a schema object name
 
             # Date literals
             (r"\{\s*d\s*(?P<quote>['\"])\s*\d{2}(\d{2})?.?\d{2}.?\d{2}\s*(?P=quote)\s*\}",
--- a/pygments/lexers/sql.py
+++ b/pygments/lexers/sql.py
@@ -620,7 +620,7 @@ class MySqlLexer(RegexLexer):
             (r'[0-9]+\.[0-9]*(e[+-]?[0-9]+)?', Number.Float),  # Mandatory integer, optional fraction and exponent
             (r'[0-9]*\.[0-9]+(e[+-]?[0-9]+)?', Number.Float),  # Mandatory fraction, optional integer and exponent
             (r'[0-9]+e[+-]?[0-9]+', Number.Float),  # Exponents with integer significands are still floats
-            (r'[0-9]+', Number.Integer),
+            (r'[0-9]+(?=[^0-9a-z$_\u0080-\uffff]|$)', Number.Integer),  # Integers that are not in a schema object name
 
             # Date literals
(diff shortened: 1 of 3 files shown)


kurtmckee:
These checks are not identical to what's in tox. I'll fix the regex linter warning and open a ticket about the tox/GitHub checks disparity.

Anteru (approved):

Anteru:
As usual, thanks for the good work!

PR #1559, opened 2020-09-30
Add missing tokens to SPARQL lexer

lucaswerkmeister:
@belett noticed that [VALUES was missing][1]; I found the other ones by running the following snippet on the SPARQL 1.1 Query Language spec:

```js
new Set(Array.from(document.querySelectorAll('.grammarTable'))
    .reduce((text, elem) => text + elem.textContent)
    .match(/'[a-z0-9-_ ]*'/ig))
```

I don’t know why a few keywords were missing; the docstring linked to the SPARQL 1.0 Query Language spec (also fixed here), but the lexer already contained other tokens which were only added in SPARQL 1.1, such as the aggregate functions (`MIN`, `MAX` etc.), which have already been in Pygments since the initial commit of the current history (6ded9db394).

[1]: https://phabricator.wikimedia.org/T264175

### DIFF
--- a/pygments/lexers/rdf.py
+++ b/pygments/lexers/rdf.py
@@ -20,7 +20,7 @@ __all__ = ['SparqlLexer', 'TurtleLexer', 'ShExCLexer']
 
 class SparqlLexer(RegexLexer):
     """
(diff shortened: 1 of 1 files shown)


Anteru:
Merged, thanks! No idea what went wrong originally, that was before my time.

PR #1565, opened 2020-10-06
llvm lexer: add freeze instruction and bfloat type

nunoplopes:
Adds:
 - new freeze instruction ([LLVM 10](https://releases.llvm.org/10.0.0/docs/ReleaseNotes.html))
 - new bfloat type ([Upcoming in LLVM 11](https://reviews.llvm.org/rG8c24f33158d81d5f4b0c5d27c2f07396f0f1484b))

Plus:
 - cleanup of types being duplicated in the keywords list
 - mark `i\d+` as type instead of generic keyword


Anteru:
Merged, thanks!

Issue #1425, opened 2020-04-10
Slow processing of JSON


kellyjonbrazil:
Can I do anything to optimize the performance of highlighting JSON in my app, or is this working as designed?
```
class JelloStyle(Style):
    styles = {
        Name.Tag: 'bold ansiblue',     # key names
        Keyword: 'ansibrightblack',    # true, false, null
        Number: 'ansimagenta',         # int, float
        String: 'ansigreen'            # string
    }

print(highlight(output, JsonLexer(), Terminal256Formatter(style=JelloStyle))[0:-1])
```
(I remove the last character because it seems for some reason `highlight` adds a newline character, but that doesn't seem to affect performance)

When I use the highlight function on a relatively large (6MB) JSON file, it adds 13 seconds to the output time:
```
time cat ls-R.out | jc --ls | jello -m
real    0m3.101s
user    0m1.727s
sys 0m0.537s
```
vs.
```
time cat ls-R.out | jc --ls | jello
real    0m16.916s
user    0m14.624s
sys 0m0.721s
```
Using the plain `TerminalFormatter` didn't seem to improve performance.

kellyjonbrazil:
Could there be any performance gains by compiling some of the core functionality via Cython or perhaps with mypyc when/if it is ready?

birkenfeld:
It might, although I suspect that most time is spent matching the regexes anyway, and of course the regex engine is already C code.

So either we optimize the regexes, or we switch to a faster regex engine...

kurtmckee:
@kellyjonbrazil could you link to the file you're using, or suggest a way to locally produce a representative file?

For example, if it's a very very very nested structure, perhaps there are some opportunities to improve the internal stack processing of the JSON lexer. Or perhaps the formatter is doing some calls/lookups/whatever that, at scale, are adding some time. Or perhaps the lexer's regular expressions are truly the issue and it would be worthwhile to explore a faster regex engine like @birkenfeld suggested.

Whatever the root cause actually is, it will be helpful to have a representative (if not identical) sample file.

By the way, is your app exclusively writing 6MB of JSON to console? Are there any other use cases (other languages, other output formats, etc) that should be accounted for while investigating?

kellyjonbrazil:
Hi @kurtmckee,

[...]

PR #1569, opened 2020-10-11
Speed up JSON and reduce HTML formatter consumption

kurtmckee:
In #1425, the author of an app that depends on Pygments reported slowness when running Pygments against large JSON files.

I investigated by generating a 118MB JSON file as input, using the inputs reported by the author of that app. I found that the regex parser's `.get_tokens_unprocessed()` took ~63 seconds to lex the entire file. I then rewrote the parser in Python and found that the time was reduced to only ~22 seconds.

I also found that Pygments was consuming ~3GB of memory when formatting the 118MB JSON file in HTML. It appears that the buffered file I/O is caching everything in memory before finally writing to the file all at once as Pygments exits. While I wasn't able to stop the buffering from waiting until the entire file was in memory, I was able to shave off an entire gigabyte of memory consumption by caching the opening span classes that are generated per-token (like `<span class="s2">` or `<span style="whatever">`). I didn't find similar memory consumption issues in the Terminal256 formatter.

For the Terminal256 formatter, this patch cuts the total runtime almost in half, dropping from 2:02 to 1:09 total processing time (as measured by Powershell's Measure-Command).

For the HTML formatter, the gains are more significant:

* Memory consumption drops by ~33%
* The HTML formatter with default options drops from 1:56 to 1:04 total processing time.
* The HTML formatter with the `noclasses` option drops from 2:59 to 1:03 total processing time.

All output from the HTML and Terminal256 formatters is 100% byte-for-byte identical between master branch and this branch.

### DIFF
--- a/pygments/lexers/data.py
+++ b/pygments/lexers/data.py
@@ -541,7 +541,7 @@ class JsonBareObjectLexer(JsonLexer):
 
 class JsonLdLexer(JsonLexer):
     """
-    For `JSON-LD <http://json-ld.org/>`_ linked data.
+    For `JSON-LD <https://json-ld.org/>`_ linked data.
 
     .. versionadded:: 2.0
     """
@@ -551,11 +551,38 @@ class JsonLdLexer(JsonLexer):
     filenames = ['*.jsonld']
     mimetypes = ['application/ld+json']
 
-    tokens = {
-        'objectvalue': [
-            (r'"@(context|id|value|language|type|container|list|set|'
-             r'reverse|index|base|vocab|graph)"', Name.Decorator,
-             'objectattribute'),
-            inherit,
-        ],
+    json_ld_keywords = {
+        '"@%s"' % keyword
+        for keyword in (
+            'base',
+            'container',
+            'context',
+            'direction',
+            'graph',
+            'id',
+            'import',
+            'included',
+            'index',
+            'json',
+            'language',
+            'list',
+            'nest',
+            'none',
+            'prefix',
+            'propagate',
+            'protected',
+            'reverse',
+            'set',
+            'type',
+            'value',
+            'version',
+            'vocab',
+        )
     }
+
+    def get_tokens_unprocessed(self, text, stack=('root',)):
+        for start, token, value in super(JsonLdLexer, self).get_tokens_unprocessed(text, stack):
+            if token is Name.Tag and value in self.json_ld_keywords:
+                yield start, Name.Decorator, value
+            else:
+                yield start, token, value
--- a/pygments/lexers/data.py
+++ b/pygments/lexers/data.py
@@ -448,6 +448,14 @@ class JsonLexer(Lexer):
     filenames = ['*.json', 'Pipfile.lock']
     mimetypes = ['application/json']
 
+    # No validation of integers, floats, or constants is done.
+    # As long as the characters are members of the following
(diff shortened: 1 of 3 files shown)


kurtmckee:
I added an LRU cache that further reduces the HTML formatter's memory consumption and processing time. The total run-time for the 118MB JSON test has dropped to ~51 seconds and the memory consumption has dropped to ~1.7GB (down from ~3.2GB).

tests/test_data.py
```diff
@@ -62,33 +223,6 @@ def test_basic_json(lexer_json):
     assert list(lexer_json.get_tokens(fragment)) == tokens
 
 
-def test_json_escape_backtracking(lexer_json):
```
Anteru, line 65: Any particular reason for removing this? The new lexer is no longer prone to this issue, as it's not using a regex, but keeping the test around doesn't hurt, and it prevents a regression. I suspect it's probably enough to check that the test terminates quickly, not the specific tokens.

Anteru (changes requested):
Ok, that's a quite impressive JSON parser there :) Well done -- my only concern is reduced test coverage. Yes, the parser is much less prone to some problems, and running those tests won't yield any new insights, but it ensures we don't regress. Unless a test is invalid, it should be kept around.

tests/test_data.py
kurtmckee: I removed it because it didn't apply. However, it won't be difficult to add back in.

I don't anticipate having time to add the pathological backtracking test back over the next few days but will as soon as I can.

kurtmckee: I've added the test back in, but I've asserted that it should take less than a second to run, rather than testing what the output is.

It may be worthwhile to use the pytest-timeout plugin to test catastrophic backtracking. Then, all such tests can be decorated with something like this:

```python
@pytest.mark.timeout(1)
def test_backtracking(lexer):
    list(lexer.get_tokens("input_text_with_backtracking_test"))
```

This would force all such tests to fail if they take too long, rather than waiting for someone to notice that a test is taking a long time.

Anteru (approved):
I wasn't aware of `pytest-timeout`, but it does indeed sound very useful, given backtracking is typically going to timeout. @birkenfeld Any concerns with adding `pytest-timeout` as a dependency?

pygments/lexers/data.py
```diff
@@ -448,71 +448,177 @@ class JsonLexer(RegexLexer):
     filenames = ['*.json', 'Pipfile.lock']
     mimetypes = ['application/json']
 
-    flags = re.DOTALL
+    integers = set('-0123456789')
```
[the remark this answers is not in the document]
kurtmckee: I've captured this in the comments in the latest patch. I also caught a comment that was no longer accurate regarding the internal queue that's created when double-quoted strings are encountered.

Anteru:
Thanks a lot, excellent work!

kurtmckee:
Thanks! I really had a blast working on this. =)

birkenfeld:
@kurtmckee thanks a lot!  I am surprised that the lru_cache made such a big difference. I guess it's kind of specific for some languages with lots of similar small tokens like braces etc.

gerner:
@kurtmckee nice work. I'm looking forward to using the parser and I'm glad that the backtracking test is still in there :) I can't wait for this to get released.

PR #1576, opened 2020-10-16
Change regex order for FortranLexer

td-mpcdf:
Punctuation check should be after the check for literal operators, otherwise the . is absorbed in a Token.Punctuation. Also added the => operator and corrected the order to check for the operators.


Anteru:
Good catch! Can you add an unit test for this, please?

Anteru (changes requested):
Looks good, but please add an unit test so it doesn't come back.

PR #1581, opened 2020-10-23
Add 'some' Ada reserved word

leogermond:
The some Ada reserved word is available since Ada 2012, it is used in the same context as the any keyword.
See RM 2.9 - Reserved Words https://www.adaic.org/resources/add_content/standards/12rm/html/RM-2-9.html for a list of keywords (with this inclusion, all are covered if I'm not mistaken)
and for usage example
See RM 4.5.8 - Quantified expressions  https://www.adaic.org/resources/add_content/standards/12rm/html/RM-4-5-8.html


Anteru:

PR #1589, opened 2020-10-30
Djangojavascriptlexerfix

### LINKED ISSUE #1588: Doxygen comments in django/javascript code creating weird behaviour from lexer
With the following code as an example
```
/**
  * @param {*} cool
  */
func = function(cool) {
};

/**
 * @param {*} stuff
 */
fun = function(stuff) {
};
```

For some reason the first '*' in '{\*}' causes everything after it to be marked a Token.Comment until you hit the second '{\*}'. The weirder thing is the behaviour only happens if both '{\*}' are present. Remove either one of the '{\*}' and it works as expected. Is there a formatter, filter, or another lexer I'm supposed to use to help with doxygen comments?

I confirmed this by running it through the demo you have on your website.

Just to clarify I'm using the javascriptDjangoLexer and I see the behaviour already happen before my filters run so its not any of my custom filters or formatters making something weird happen.

For reference of the correctness of the commenting you can look at JSDoc [here](https://jsdoc.app/tags-param.html)

boanergies:
Fixes #1588. Only needed to change the one line of code the Django lexer to remove `{* .. *}` from being thought as a Django comment.

### DIFF
--- a/pygments/lexers/templates.py
+++ b/pygments/lexers/templates.py
@@ -338,7 +338,7 @@ class DjangoLexer(RegexLexer):
             (r'[^{]+', Other),
             (r'\{\{', Comment.Preproc, 'var'),
             # jinja/django comments
-            (r'\{[*#].*?[*#]\}', Comment),
+            (r'\{#.*?#\}', Comment),
             # django comments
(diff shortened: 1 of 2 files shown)


birkenfeld:
Very nice, thanks!

PR #1590, opened 2020-11-02
testing turtle prefix names where reference starts with number

### LINKED ISSUE #1553: turtle CURIE highlight differs if reference starts with number
This screenshot should show the difference

image

It seems that references starting with number result in incorrect highlight

elf-pavlik:
closes #1553

I don't work with python myself but I hope changes here are reasonable. ShExC lexer handles prefixed names where part after `:` starts with number correctly. This change should only copy that part of ShExC lexer to Turtle lexer.

### DIFF
--- a/pygments/lexers/rdf.py
+++ b/pygments/lexers/rdf.py
@@ -187,6 +187,53 @@ class TurtleLexer(RegexLexer):
     filenames = ['*.ttl']
     mimetypes = ['text/turtle', 'application/x-turtle']
 
+    # character group definitions ::
+    PN_CHARS_BASE_GRP = ('a-zA-Z'
+                         '\u00c0-\u00d6'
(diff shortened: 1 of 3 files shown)


Anteru:
Thanks for the contribution! Can you please take a look at the regexlint failures? It looks like some of your expression can be simplified.

elf-pavlik:
I did my best trying to align Turtle lexer with SPARQL and ShExC, not the other way. I also posted on RDFlib mailing list if some python developer familiar with RDF would consider helping with those RDF lexers: https://groups.google.com/g/rdflib-dev/c/hSGFT9jV61k

In my last commit I needed to fix example.ttl which wasn't valid turtle. Now it passes validation with http://ttl.summerofcode.be/

elf-pavlik:
Please let me know if there's anything else I should do in this PR.

PR #1592, opened 2020-11-06
fix closing tag for unnamed blocks on MasonLexer

gandarez:
The regex for unnamed blocks seems to be wrong defined. According to the [official documentation](https://metacpan.org/pod/distribution/Mason/lib/Mason/Manual/Syntax.pod#%3C%class%3E) all unnamed blocks including class/doc/init must be ended by a </%class> and not by </%class%> and subsequently.

Example here:
https://metacpan.org/pod/distribution/Mason/lib/Mason/Manual/Intro.pod#Email-generator-(from-script)

### DIFF
--- a/pygments/lexers/templates.py
+++ b/pygments/lexers/templates.py
@@ -565,7 +565,7 @@ class MasonLexer(RegexLexer):
 
     def analyse_text(text):
         result = 0.0
-        if re.search(r'</%(class|doc|init)%>', text) is not None:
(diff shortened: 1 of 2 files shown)


Anteru:
Good catch and thanks for the contribution! Could you please add a small unit test for this to make sure it won't regress?

gandarez:
@Anteru unit test added as required.

Anteru:
Thanks!

PR #1593, opened 2020-11-08
Fix mysql analyze text logic

gandarez:
According to the documentation MySQL syntax does not allow the usage of square brackets in its queries. Saying it's a truth I removed all the logic comparison between square brackets and backticks.

https://stackoverflow.com/a/9719904/833531
https://dev.mysql.com/doc/refman/8.0/en/identifiers.html

### DIFF
--- a/pygments/formatters/latex.py
+++ b/pygments/formatters/latex.py
@@ -304,8 +304,8 @@ class LatexFormatter(Formatter):
                            (rgbcolor(ndef['border']),
                             rgbcolor(ndef['bgcolor'])))
             elif ndef['bgcolor']:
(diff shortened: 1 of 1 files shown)


Anteru:
I think the point here was to differentiate MySQL from MSSQL, which does use square brackets, and both will probably be trigger happy on SQL text. Was it incorrectly identifying code or what triggered this change?

gandarez:
The reason I opnened this PR is because we've been porting pygments lexers to chroma project (GoLang) and the MySQL lexer contains validations (the square brackets) for MSSQL as you already mentioned. What I'm not sure is which weight should I return.

gandarez:
@Anteru any update here?

gandarez:
Any update here?

PR #1594, opened 2020-11-09
Fix a catastrophic backtracking bug in JavaLexer

### LINKED ISSUE #1586: Denial of service with malformed file
When running the following code with the latest git version of pygments on the attached input results of in 100% CPU consumption for an arbitrary long time:

```python
import sys

import pygments
import pygments.formatters
import pygments.lexers

with open(sys.argv[1], 'rb') as f:
    data = f.read()
    lexer = pygments.lexers.guess_lexer(str(data))
    pygments.highlight(str(data), lexer, pygments.formatters.HtmlFormatter())
```

[timeout-9a00111e78b5cd0979a370fc9a5cd22e39a249e4.txt](https://github.com/pygments/pygments/files/5439983/timeout-9a00111e78b5cd0979a370fc9a5cd22e39a249e4.txt)

kurtmckee:
This closes #1586, which triggered the bug.

1. Pygments guessed from an input string that it should use SspLexer to lex the file.
2. The SspLexer is a delegating lexer that relies on XmlLexer and JspRootLexer.
3. JspRootLexer relies on regular expressions from the JavaLexer.
4. The JavaLexer has a catastrophic backtracking bug in its string literal regular expression.

This patch addresses the problem by replacing the string literal regex with a sub-state for string literal parsing.

### DIFF
--- a/pygments/lexers/jvm.py
+++ b/pygments/lexers/jvm.py
@@ -65,7 +65,7 @@ class JavaLexer(RegexLexer):
              'var'),
             (r'(import(?:\s+static)?)(\s+)', bygroups(Keyword.Namespace, Text),
              'import'),
-            (r'"(\\\\|\\"|[^"])*"', String),
+            (r'"', String, 'string'),
             (r"'\\.'|'[^\\]'|'\\u[0-9a-fA-F]{4}'", String.Char),
             (r'(\.)((?:[^\W\d]|\$)[\w$]*)', bygroups(Punctuation,
                                                      Name.Attribute)),
@@ -95,6 +95,13 @@ class JavaLexer(RegexLexer):
         ],
         'import': [
             (r'[\w.]+\*?', Name.Namespace, '#pop')
+        ],
+        'string': [
+            (r'[^\\"]+', String),
+            (r'\\\\', String),  # Escaped backslash
+            (r'\\"', String),  # Escaped quote
+            (r'\\', String),  # Bare backslash
+            (r'"', String, '#pop'),  # Closing quote
         ],
     }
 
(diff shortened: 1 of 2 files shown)


Anteru:
Thanks! High time we add `pytest-time`, I guess :)

kurtmckee:
> Thanks! High time we add `pytest-time`, I guess :)

If it's okay with you and Georg, I'll work on a patch to add `pytest-timeout` (which will attempt to hard-stop tests that run too long) and to update existing backtrack-checking tests to use the timeout.

Let me know if that's something you'd like me to tackle. =)

Anteru:
I have no objection, and given Georg didn't red flag it, I'd say please go ahead. It's fairly obvious to me by now that backtracking needs robust testing, and I don't see how we can do it without timeouts.

kurtmckee:
Okay, then I'll create a ticket for tracking and discussion and submit a patch for review.

birkenfeld:
Sure, I'd just like a graceful behavior if the timeout plugin isn't installed.

kurtmckee:
Understood. I anticipate that I'll add it as a dependency in tox so that it's guaranteed to be available when running the tests.

kurtmckee:
By the way, I just tested the pytest timeout plugin. It supports a thread-based and a signal-based method.

On Windows, which only supports thread-based timeouts, the plugin is unable to interrupt catastrophic backtracking like this:

```python
import re

def test_catastrophic_backtracking():
    re.search(r'\d+\d+a', '0' * 10_000)
```

The thread-based timeouts can successfully interrupt operations like `time.sleep(60)` but cannot successfully interrupt catastrophic backtracking. For this reason I'm not pursuing using the pytest-timeout plugin.

PR #1597, opened 2020-11-09
Fix coq analyze text logic

gandarez:
the current Coq analysis the actual value for completing a proof is `Qed` but also `qed` can appear on it. So I applied a regex to validate it.

https://coq.inria.fr/refman/proof-engine/proof-handling.html?highlight=qed#coq:cmd.qed
https://coq.inria.fr/refman/proof-engine/tactics.html#coq:tacn.tauto

### DIFF
--- a/pygments/lexers/theorem.py
+++ b/pygments/lexers/theorem.py
@@ -154,7 +154,7 @@ class CoqLexer(RegexLexer):
     }
 
     def analyse_text(text):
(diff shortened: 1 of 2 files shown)


gandarez:
@Anteru any update here?

birkenfeld:
Seems like this has been superseded by #1648.

PR #1602, opened 2020-11-16
RawToken{Formatter,Lexer}: support Python 3 and handle exceptions

andersk:
In Python 3, `RawTokenFormatter` would output non-ASCII for non-ASCII input, and `RawTokenLexer` would throw Unicode-related exceptions for ASCII or non-ASCII input; fix them. Also, handle all exceptions, so that callers who find `RawTokenLexer` via `get_lexer_by_name` on user input don’t unexpectedly get a lexer that throws exceptions.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -41,6 +41,7 @@ Version 2.8.0
 - Added markdown to MarkdownLexer aliases (#1687)
 - Removed "raw" alias from RawTokenLexer, so that it cannot be
   selected by alias.
+- Fixed RawTokenLexer to work in Python 3 and handle exceptions.
 - Added prompt colors to the Solarized theme (#1529)
 
 Version 2.7.4
--- a/pygments/formatters/other.py
+++ b/pygments/formatters/other.py
@@ -87,34 +87,32 @@ class RawTokenFormatter(Formatter):
             import gzip
             outfile = gzip.GzipFile('', 'wb', 9, outfile)
 
-            def write(text):
-                outfile.write(text.encode())
-            flush = outfile.flush
+            write = outfile.write
+            flush = outfile.close
         elif self.compress == 'bz2':
             import bz2
             compressor = bz2.BZ2Compressor(9)
 
             def write(text):
-                outfile.write(compressor.compress(text.encode()))
+                outfile.write(compressor.compress(text))
 
             def flush():
                 outfile.write(compressor.flush())
                 outfile.flush()
         else:
-            def write(text):
-                outfile.write(text.encode())
+            write = outfile.write
             flush = outfile.flush
 
         if self.error_color:
             for ttype, value in tokensource:
-                line = "%s\t%r\n" % (ttype, value)
+                line = b"%r\t%r\n" % (ttype, value)
                 if ttype is Token.Error:
                     write(colorize(self.error_color, line))
                 else:
                     write(line)
         else:
             for ttype, value in tokensource:
-                write("%s\t%r\n" % (ttype, value))
+                write(b"%r\t%r\n" % (ttype, value))
         flush()
 
 
--- a/pygments/lexers/special.py
+++ b/pygments/lexers/special.py
@@ -8,8 +8,8 @@
     :license: BSD, see LICENSE for details.
 """
 
+import ast
 import re
-from io import BytesIO
 
 from pygments.lexer import Lexer
 from pygments.token import Token, Error, Text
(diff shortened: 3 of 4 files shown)


birkenfeld:
The non-ASCII handling fixes are useful, but I don't think the lexer should swallow exceptions.

It is special in that it should faithfully restore a token stream, and if that stream is corrupted it needs to be visible, not just an "Error" token.

andersk:
Do you have a concrete reason this lexer should raise exceptions, or just a fuzzy feeling? Because I have a concrete reason why it shouldn’t.

```pytb
$ pip install markdown pygments
$ cat input.md
    :::raw
    δ
$ python3 -m markdown -x codehilite < input.md
  File "/nix/store/cqh3nsk4z8lc960g8yqvz5nf2f6qx9v0-python3-3.8.6/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/nix/store/cqh3nsk4z8lc960g8yqvz5nf2f6qx9v0-python3-3.8.6/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/tmp/ve/lib/python3.8/site-packages/markdown/__main__.py", line 151, in <module>
    run()
  File "/tmp/ve/lib/python3.8/site-packages/markdown/__main__.py", line 145, in run
    markdown.markdownFromFile(**options)
  File "/tmp/ve/lib/python3.8/site-packages/markdown/core.py", line 405, in markdownFromFile
    md.convertFile(kwargs.get('input', None),
  File "/tmp/ve/lib/python3.8/site-packages/markdown/core.py", line 334, in convertFile
    html = self.convert(text)
  File "/tmp/ve/lib/python3.8/site-packages/markdown/core.py", line 268, in convert
    newRoot = treeprocessor.run(root)
  File "/tmp/ve/lib/python3.8/site-packages/markdown/extensions/codehilite.py", line 246, in run
    placeholder = self.md.htmlStash.store(code.hilite())
  File "/tmp/ve/lib/python3.8/site-packages/markdown/extensions/codehilite.py", line 143, in hilite
    return highlight(self.src, lexer, formatter)
  File "/tmp/ve/lib/python3.8/site-packages/pygments/__init__.py", line 84, in highlight
    return format(lex(code, lexer), formatter, outfile)
  File "/tmp/ve/lib/python3.8/site-packages/pygments/__init__.py", line 63, in format
    formatter.format(tokens, realoutfile)
  File "/tmp/ve/lib/python3.8/site-packages/pygments/formatter.py", line 95, in format
    return self.format_unencoded(tokensource, outfile)
  File "/tmp/ve/lib/python3.8/site-packages/pygments/formatters/html.py", line 930, in format_unencoded
    for t, piece in source:
  File "/tmp/ve/lib/python3.8/site-packages/pygments/formatters/html.py", line 764, in _wrap_div
    yield from inner
  File "/tmp/ve/lib/python3.8/site-packages/pygments/formatters/html.py", line 781, in _wrap_pre
    yield from inner
  File "/tmp/ve/lib/python3.8/site-packages/pygments/formatters/html.py", line 786, in _wrap_code
    yield from inner
  File "/tmp/ve/lib/python3.8/site-packages/pygments/formatters/html.py", line 804, in _format_lines
    for ttype, value in tokensource:
  File "/tmp/ve/lib/python3.8/site-packages/pygments/lexers/special.py", line 70, in get_tokens
    text = text.encode('ascii')
UnicodeEncodeError: 'ascii' codec can't encode character '\u03b4' in position 0: ordinal not in range(128)
```

This caused internal server errors in the Zulip message sending codepath. There was no reason for the developers of Zulip, Python-Markdown, or the codehilite extension to expect that certain user input could trigger the selection of this weird lexer that raises exceptions unlike all the others, unless they had gone and read the documentation for every individual Pygments lexer!

If for some reason it’s really important for this lexer to throw some exceptions (but not others—why was there a partial exception handler to begin with, then?), can we remove this lexer from consideration by `get_lexer_by_name` so it needs to be specifically requested by the code?

birkenfeld:
If you re-read my reply, you'll find that I have given a reason beyond "fuzzy feelings" in the second sentence.

However, you are correct in that this is behavior, and the special-ness of the lexer, are not easily apparent, and there is a mix-up of both behaviors already, so I'm getting convinced to change it to using Error tokens throughout. @Anteru any thoughts?

Anteru:
@birkenfeld Sorry for being a bit late to the party ... Personally, I'd expect that no lexer will throw an exception and instead produce Error tokens where it can't make sense of the input. It seems like a burden for users to wrap syntax highlighting in a `try`, `catch`, and they can't fix incorrect input. Exceptions are fine if a lexer is used incorrectly (i.e. it's an app issue), but wrong input is in my opinion hardly exceptional :)

It should however produce error tokens which contain the original input in that case, so we can still round-trip things. That seems valuable to have.

birkenfeld:
@Anteru candidate for 2.8 I think.

Anteru:
Sounds good to me. Can anyone please draft something for the changelog here?

andersk:
I fixed some additional bugs and added tests and a changelog entry.

Anteru:
Thanks a lot! Merged for 2.8.

Issue #1604, opened 2020-11-17
Duplicate aliases for the language "Juttle"


sumanthvrao:
The aliases for the language Juttle occurs twice in the tuple resulting from get_all_lexers(). Here's a small snippet to reproduce 

``` py3
In [15]: from pygments.lexers import get_all_lexers

In [16]: lexers = get_all_lexers()

In [17]: {alias[0]: alias[1] for alias in lexers}.get('Juttle')
Out[17]: ('juttle', 'juttle')
```

Pygments PyPI version used: 2.7.2 (latest)

PR #1606, opened 2020-11-17
JuttleLexer: Fix duplicate 'juttle' occurance in lexer aliases.

sumanthvrao:
The output from pygments.lexers.get_all_lexers() contains 'juttle'
twice in the aliases section for the Juttle lexer entry.

This could be reproduced using:

```
>>> from pygments.lexers import get_all_lexers
>>> lexers = get_all_lexers()
>>> {alias[0]: alias[1] for alias in lexers}.get('Juttle')
('juttle', 'juttle')
```

This patch fixes the duplicate entry and generates the associated
_mapping.py file.

Fixes: #1604


birkenfeld:
Thanks!

PR #1607, opened 2020-11-17
Fix CBM BASIC V2 analyze text logic

gandarez:
This PR fixes **CBM BASIC V2** text analyser to match decimal at the beginning not at anywhere as it was written. Also in the comments you can see `if it starts with a line number...`. After fixing `analyse_text` function it started breaking the test `test_guess_lexer_forth` which was expecting to get `ForthLexer` but started getting `CbmBasicV2Lexer` so to fix it I changed the weight for ForthLexer to **0.3**. All tests are passing.


gandarez:
@Anteru any update here?

gandarez:
Any update here?

birkenfeld:
Thanks, I rebased it accordingly.

PR #1608, opened 2020-11-18
Rust lexer: bug fix with rust macros

klux2:
Rust macros would not be recognized/formatted, as they end with a '!', which does not count as word boundary. This was required by the suffix r'\b' (after the '!').
A typical macro call like `"println!("test");"` would not be matched by the associated regex `r"println!\b"` (it would instead be matched by `r"println\b!"`).

To fix this problem, the suffix is removed. As every macro ends with an '!' (which implicitely includes a word boundary before), it's not necessary anyway for the macros.

There were also several keywords which are not macros (as far as I know) listed in this category. They were moved to the correct section. As both categories used the same token anyway, it doesn't really matter.

Lastly, the token type for macros was changed. Rust macros seem to fit more into the *magic function* (`Name.Function.Magic`) category than into the *builtin* (`Name.Builtin`) one.


birkenfeld:
Thanks for the PR! The failure to highlight macros is indeed embarrassing.

Issue #1213, opened 2019-08-31
img formatter renders text at wrong position (on Windows?)


Anteru:
(Original issue 1509 created by hhsprings on 2019-04-26T01:11:55.667733+00:00)

Note that I have reproduced this problem only on Windows (Japanese edition).

In the case where the lexer is other than "text", the drawing position of tokens other than the first one appearing on the line is wrong.

If you change the passing text to getsize in the "get_char_size" method of class FontManager from "M" to "a", it will draw in the correct position, but I can't explain what's happened. Apart from this mysterious fix method, I think that in the first place the approach adopted does not work when it contains kanji characters.

Maybe I think this is the conclusion:

* Don't multiply the font size by the character position.
* Use the size of all text before the drawing target token in the line.

---
used versions:

* python 2.7, 3.5
* pygments 2.3.1
* pillow 4.2.1
* Windows 7

Anteru:
(Original comment by hhsprings on 2019-04-26T06:06:48.733712+00:00)

Sorry to attach the modified source code directly. I got lost about how to do PR with BitBucket + Mercurial.

15b3:
I faced a similar problem.

I outputed the image by using ImgFormatter from this sample code containing Japanese characters:
```python
# sample.py

print("Hello, World!")
print("こんにちは、世界！")
```
Outputed Image:
![sample_before](https://user-images.githubusercontent.com/30496251/100328836-4e86d000-3010-11eb-82bb-e1dc76eefd41.png)
The position of the `")` is wrong. And, size of the image is too small.

As hhsprings suggests, this may be due to calculating the start position as the number of characters multiplied by the number of widths of the letter "M", especially when the character widths are different. 
In this case, it is Japanese font, but I believe the same is possible in Chinese as well.  #1558 

This can be solved by using the `get_size` method of the `PIL.ImageFont.ImageFont` class and drawing the token from the calculated position for each token.

How about a fix like this one? ( 15b3@2bdcf50 )

the result image (fixed):
![sample_after](https://user-images.githubusercontent.com/30496251/100328851-52b2ed80-3010-11eb-9f44-24572e992039.png)

Anteru:
That does sound reasonable -- can you please file a PR for it? Does this have significant drawbacks in terms of performance, as it needs one call per line now?

15b3:
PR is OK.

The calculation cost will be greater.
To be honest, I'm not sure how much. (I don't have enough familiarity with Pillow's method.)

[...]

PR #1611, opened 2020-11-26
ImgFormatter: Use the start position based on the length of text

15b3:
Related #1213 

Changed from calculating the start position as the number of characters multiplied by the number of widths of the letter "M" to using the `get_size` method of the `PIL.ImageFont.ImageFont` class and drawing the token from the calculated position for each token.

### DIFF
--- a/pygments/formatters/img.py
+++ b/pygments/formatters/img.py
@@ -209,6 +209,12 @@ class FontManager:
         """
         return self.fonts['NORMAL'].getsize('M')
 
+    def get_text_size(self, text):
+        """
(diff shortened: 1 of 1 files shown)


Anteru:
Merged, thanks a lot of for the timeline PR!

15b3:
Could I write my name on AUTHORS?

Anteru:
@birkenfeld Any objections? I don't think we have established rules yet for which change warrants a mentioning in `AUTHORS`.

birkenfeld:
No objections.

PR #1612, opened 2020-11-26
llvm lexer: add poison keyword

nunoplopes:
New 'poison' keyword coming with LLVM 12: http://lists.llvm.org/pipermail/llvm-commits/Week-of-Mon-20201123/856253.html


Anteru:
Merged, thanks a lot!

PR #1613, opened 2020-11-26
Add new Lexer for OCL language, version 2.4

sunye:
Hi,

I needed a lexer for the OCL langiage from OMG and I wrote this one this afternoon. I hope it can help.

Regards,

Gerson

### DIFF
--- a/pygments/lexers/ocl.py
+++ b/pygments/lexers/ocl.py
@@ -1,0 +1,228 @@
+# -*- coding: utf-8 -*-
+"""
+    pygments.lexers.ocl
+    ~~~~~~~~~~~~~~~~~~~
+
+    Lexer for the Object Constraint language.
+
+    :copyright: Copyright 2006-2020 by the Pygments team, see AUTHORS.
+    :license: BSD, see LICENSE for details.
+"""
+
+from pygments.lexer import RegexLexer, bygroups, include, combined, default, words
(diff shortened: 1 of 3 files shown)


pygments/lexers/ocl.py
```diff
@@ -0,0 +1,207 @@
...
+            (r'--.*?\n', Comment.Single),
+            (r'[]{}:(),;[]', Punctuation),
+            (r'\\\n', Text),
+            (r'\\', Text),
+            (words(_operator, suffix=r'\b'), Operator.Word),
+            (r'<>|==|->|<=|>=|\?|[-~+/*%=<>&^|.!]', Operator),
+            include('keywords'),
+            include('builtins'),
+            include('name'),
+            (r'("(\\\\|\\"|[^"])*")|(\'(\\\\|\\\'|[^\'])*\')', String),
```
Anteru, line 183: This looks like you're trying to match strings and allow escapes inside. This is error prone (we've had a bunch of issues w.r.t. to catastrophic backtracking and strings) to put into a single regex. Can you please use state-based processing here? I.e. push/pop when entering/leaving a string?

Anteru (changes requested):
Please check the string handling again, and move the example file to /tests/examplefiles/ocl/... and generate the expected output using `pytest --update-goldens` (see `Contributing.md` for details.)

Issue #1617, opened 2020-11-28
extremely poor performance on certain markdown file


gerner:
This particular markdown takes a long time (21 seconds on my laptop) to parse:
https://github.com/date-fns/date-fns/blob/a9fc0c7b715883349555bfb94daa1059430eda52/src/locale/en-US/snapshot.md

```
$ time pygmentize -f terminal -l md -o /dev/null /tmp/snapshot.md

real    0m21.318s
user    0m21.306s
sys     0m0.012s
```

I've seen slow-ish parsing performance on markdown with tables before, however I'm not certain it's tables that's causing the issue.

gerner:
I do see Error tokens showing up for '\n' on its own line which I think is because the rule here doesn't match a newline:

https://github.com/pygments/pygments/blob/master/pygments/lexers/markup.py#L601

That seems like a separate issue. If I add a rule to match r'\n' (most) of the errors go away, but that doesn't speed up processing any.

gerner:
I think tables is a red herring. I pulled out the table rules and performance didn't change.

However, this rule looks expensive:

```
# strikethrough
(r'([^~]*)(~~[^~]+~~)', bygroups(Text, Generic.Deleted)),
```

I don't know why, but if I comment it out the file is processed in 400ms down from 21s. Note that this file doesn't have the character "~" anywhere in it.

Also, why the heavy use of bygroups throughout?

gerner:
Note, if I change the rule as suggested above the output of pygmentize doesn't change and all test cases still pass. It just happens in 400ms instead of 21s.

gerner:
This is fixed by #1623

PR #1618, opened 2020-12-01
Simplify markdown strikethrough regex and leave text parsing to lower cases

gerner:
the regex: `r'([^~]*)(~~[^~]+~~)'` ends up being expensive looking for essentially a string of any characters (not "~") of any length preceeding the strikethrough case. In the cases that this fails on a long string we might fall back to consuming a single character and then checking again this regex again, only to fail again. In the worst case we might repeat this O(n) regex for every character in the file leading to O(n^2) behavior.

#1617 illustrates how this can cause poor performance in practice on an otherwise "normal" markdown file.

This fix eliminates the `[^~]* ` group so it will only trigger when the next character is "~". I believe this is ok because other cases that will advance lexing are either specific to other cases with reasonable starting and terminating patterns (e.g. image tags or code spans), or are the non-space, non-escape string text case or the single character text matching case. In these cases we'll consume a minimal (in some sense) amount of content until we might possibly come back to the start of the strikethrough. In fact, other than the non-space text case, this is the only "inline" pattern that consumes nearly arbitrary strings prior to matching a specific pattern.

Existing tests pass and I added another simple, happy-path test case just to make sure that text processing isn't messed up. I also tested the old and new logic on the file from #1617 and got the same terminal formatted result, only got there in 400ms instead of 21s.

### DIFF
--- a/pygments/lexers/markup.py
+++ b/pygments/lexers/markup.py
@@ -582,7 +582,7 @@ class MarkdownLexer(RegexLexer):
             # italics fenced by '_'
             (r'(\_[^_ \n][^_\n]*\_)', bygroups(Generic.Emph)),
             # strikethrough
-            (r'([^~]*)(~~[^~]+~~)', bygroups(Text, Generic.Deleted)),
+            (r'(~~[^~]+~~)', bygroups(Generic.Deleted)),
             # mentions and topics (twitter and github stuff)
             (r'[@#][\w/:]+', Name.Entity),
             # (image?) links eg: ![Image of Yaktocat](https://octodex.github.com/images/yaktocat.png)
--- a/pygments/lexers/markup.py
+++ b/pygments/lexers/markup.py
@@ -597,7 +597,7 @@ class MarkdownLexer(RegexLexer):
              bygroups(Text, Name.Label, Text, Name.Attribute)),
 
             # general text, must come last!
-            (r'[^\\\s]+', Text),
+            (r'[^\\\s`~*]+', Text),
             (r'.', Text),
         ],
     }
--- a/tests/test_markdown_lexer.py
+++ b/tests/test_markdown_lexer.py
@@ -7,6 +7,8 @@
     :license: BSD, see LICENSE for details.
 """
 
+import time
+
 import pytest
 from pygments.token import Generic, Token, String, Keyword, Name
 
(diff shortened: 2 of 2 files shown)


gerner:
There's a slight problem with this in that the following won't get the right tokens:

```markdown
this is not striked~~but this should be striked but is not~~
```

which GitHub renders:

this is not striked~~but this should be striked but is not~~

I have another forthcoming change that fixes this.

gerner:
the key to handling inline formatting embedded in a word is by specializing the text rule to not match these inline formatting characters "\~\*\`" (notably excluding "_"). This allows these inline formatting to trigger when embedded within words without being delimited by spaces. There's still a fallback rule that matches any single character when this doesn't trigger formatting (e.g. foo*bar, bar should not be emphasized).

"_" doesn't trigger formatting when embedded in words, at least on GitHub. I have a GitHub test gist here:

https://gist.github.com/gerner/8351bd8abd709ef892808f041584fae2


This change is necessary to prevent a regression in this case for strikethrough, but at the same time I think it fixes previously unknown bugs for the other cases. I added test cases that checks for those cases as well as the strikethrough case that already worked because of how the strikethrough regex was written, but would have been broken by my change if I hadn't caught the issue.

I think this also produces more sensible stream of tokens too, that is, each word turns into one Text token instead of a string of words as a single Text token preceeding a strikethrough string.

gerner:
I added a test case that covers the original perf issue. pygmentizing a file with that fragment used to take 25 seconds. Now pygmentize takes 400ms (of which much is just spinning up the python process I think).

The fragment is 104KB on a single line "this is text " over and over, but the same pattern happens if there are newlines throughout (e.g. 104KB of "this is text\n").

Note that the original file from #1617 was 79KB. I believe that's above average, perhaps even in the top 20% or so of markdown files, but not crazy large. In fact, I converted doc/docs/lexerdevelopment.rst (about 28KB) to markdown using pandoc and ran pygmentize on it using the old code and the new. It used to take just over 1 second to pygmentize before and now takes 200ms. Because of the O(n^2) nature things get a lot worse, really quickly. So I think it's reasonable to expect this is at best annoying for lots of markdown docs and at worst totally blocking for some.

gerner:
Ouch, looks like there are other perf issues with pypy which make my perf test too aggressive. I've slimmed it up substantially. I don't think it actually provides much perf regression coverage now, but fixing the other perf issues is probably out of scope for this PR.

gerner:
Just to be clear, I think this change is good to go. The perf issue with pypy is not a regression. I'm just pointing out there are other perf issues in pypy that make a timed test hard across different python interpreters. This change doesn't make things worse, indeed I think it improves things. That improvement is very dramatic on cpython. On pypy it's more subtle.

Anteru:
Obsoleted by 1623.

PR #1619, opened 2020-12-02
Jasmin min score for analyse text

gandarez:
This PR adds `min` to make sure the highest value isn't greater than 1.0. Used the same logic as https://github.com/pygments/pygments/blob/master/pygments/lexers/gdscript.py#L346


gandarez:
Any update here?

Anteru:
Merged, sorry it took forever.

PR #1621, opened 2020-12-06
Move setup.py to a declarative setup.cfg

jdufresne:
Docs: https://setuptools.readthedocs.io/en/latest/userguide/declarative_config.html

Removes mixing code with configuration. No need to execute custom code
at the build step. The declarative configuration is also more friendly
to scripting.


birkenfeld:
Generally +1, but we should put the required setuptools version in the changelog.

@Anteru let's keep this for 2.8.

Anteru:
Sounds good to me. Speaking of the setuptools version -- which one _is_ the right one?

jdufresne:
Thanks for the reviews. I added a note to `CHANGES`, let me know if there is anything else I should include.

Anteru:
Looks good! Did you check if it produces the exact same package/wheel? Are there any other benefits in using the declarative variant that we might want to take advantage of?

jdufresne:
> Did you check if it produces the exact same package/wheel?

I extracted both the sdist and the wheel and compared the results. They look identical to me. If you catch something, please let me know, but I didn't find anything.

birkenfeld (approved):

PR #1622, opened 2020-12-07
Run pyupgrade across codebase to modernize syntax and patterns

jdufresne:
pyupgrade is a tool to automatically upgrade syntax for newer versions
of the Python language.

The project has been Python 3 only since
35544e2fc6eed0ce4a27ec7285aac71ff0ddc473, allowing for several cleanups:

- Remove unnecessary "-*- coding: utf-8 -*-" cookie. Python 3 reads all
  source files as utf-8 by default.

- Replace IOError/EnvironmentError with OSError. Python 3 unified these
  exceptions. The old names are aliases only.

- Use the Python 3 shorter super() syntax.

- Remove "utf8" argument form encode/decode. In Python 3, this value is
  the default.

- Remove "r" from open() calls. In Python 3, this value is the default.

- Remove u prefix from Unicode strings. In Python 3, all strings are
  Unicode.

- Replace io.open() with builtin open(). In Python 3, these functions
  are functionally equivalent.


birkenfeld:
Looks good, except that the files in `tests/examplefiles` should not be changed, they are test input.

jdufresne:
Thanks for catching that. Fixed in the latest revision.

Anteru:
@birkenfeld Ready to merge assuming you're not hacking on things right now?

birkenfeld:
Go ahead!

bl-ue:
I know this was a while ago, but I'd like to add that there was also a `yield` condensation effected, [here](https://github.com/pygments/pygments/pull/1622/files#diff-df5a7589bb821e11991511a8268f034ba60bc44f32b8b277617fc900f0d2638cL105-R104).

PR #1623, opened 2020-12-07
Markdown lexer improvements

Leistungsabfall:
This is a series of improvements to the Markdown Lexer:

* Fenced code was not recognized when it contained empty lines
* Inline code was not recognized after bulleted list entry (e.g. ```* `code` ```)
* Bold and italics were not recognized when surrounded with non-space char (e.g. `(**foo**)`)
* There is no way to recognize code indented by 4 spaces using regex only - so remove the error-prone detection


gerner:
I verified that the perf issue described in #1617 is fixed by this change with identical output from pygmentize on the test file. If this PR gets merged then my proposed fix in #1618 is unnecessary.

PR #1635, opened 2020-12-14
feat: added 'ONLY' to fortran.py Keywords

### LINKED ISSUE #1543: Fortran. "Only" clause used when importing modules variables is not highlighted.
The following Fortran code:

```
PROGRAM example

   USE testmodule, ONLY: var

   IMPLICIT NONE

   INTEGER :: n

   n = 1
   WRITE(*,*) n

END PROGRAM example
```

produces the following output:

only

The ```ONLY``` clause at line 3 should be highlighted. It can be uppercase and lowercase.

0xflotus:
maybe close #1543

Is it that simple?


birkenfeld:
Thanks for the PR!

PR #1642, opened 2020-12-23
Fixed highlighting template parameter without name

### LINKED ISSUE #1076: C++ lexer fails to parse default template argument
(Original issue 1372 created by twesterhout on 2017-08-28T23:42:07.373120+00:00)

In short, the following piece of C++ code is not parsed correcly
```cpp
template <class View, class Predicate, class = std::enable_if_t<
    boost::static_views::concepts::View::test<View>()>>
constexpr auto operator()(View const& xs, Predicate&& p) const;
```
For a longer version, see this discussion [#4025](https://github.com/sphinx-doc/sphinx/issues/4025).

AkibAzmain:
Pygments fails to highlight C++ code when it contains the keyword `class` without any name followed by. This pull request fixes it.

Closes #1076 and solves sphinx-doc/sphinx#4025.


birkenfeld:
This is not entirely correct, since it now allows prefixes of words like `classic` to be highlighted as a keyword.

Issue #678, opened 2019-08-31
unicode in coq files marked as error with "err"


Anteru:
(Original issue 971 created by DanGrayson on 2014-03-31T13:42:42.840883+00:00)

Github uses Pygments.  Look at

https://github.com/DanGrayson/UniMath/blob/370340a475eeafd22ac5fdde62c947a194812e0f/UniMath/Ktheory/AffineLine.v#L18

and you will see that some unicode characters are tagged as errors in the css, making them display badly.  The corresponding html is

<span class="err">ℤ</span>

The CSS comes from

https://github.global.ssl.fastly.net/assets/github-b4baae9c06716bcd0a8068e3030607851f19b980.css

and the particular CSS making the unpleasant color is

.err{color:#a61717;background-color:#e3d2d2}

Perhaps your coders think unicode characters in coq code are not allowed, and are helpfully flagging it as an error.  But unicode characters are allowed in coq, and I think it makes it rather pretty.

Could that be fixed?

Github staff assure me that a fix in pygments will make its way eventually to github.

Anteru:
(Original comment by tshatch on 2014-04-15T17:23:03.332005+00:00)

Can you confirm that this grammar is for the same Coq?

http://coq.inria.fr/refman/Reference-Manual003.html

If so, yes I see unicode allowed in identifiers, and this should be easy to fix.

Anteru:
(Original comment by DanGrayson on 2014-04-15T20:35:10.572881+00:00)

Yes, that's the same Coq?  Thanks!

Anteru:
(Original comment by DanGrayson on 2014-04-15T20:35:12.936625+00:00)

I meant: "the same Coq."  (no question mark)

Anteru:
(Original comment by tshatch on 2015-10-17T03:03:30.340448+00:00)

Will try to get to this.

Anteru:
(Original comment by JasonGross on 2019-05-06T00:55:02.834725+00:00)

I just ran into this in TeX’s minted package.  Is this getting fixed any time soon?  It seems like it should be a straightforward fix, and this issue has been open for over five years.

Anteru:
(Original comment by DanGrayson on 2019-05-06T18:21:04.618262+00:00)

I’d forgotten about this, and the link to our UniMath repository that I gave in my original post leads to a page with no visible coloring problem.  So maybe something was done??

Anteru:
[...]

PR #1648, opened 2020-12-26
Fix Coq-related bug #678

mwuttke97:


Anteru:
Is the `MULTILINE` flag intentional? I don't see rules requiring it, and I'd rather not have flags added with no effect.

mwuttke97:
Good point! I've updated the commit.

In fact, I just copied the flag from the Lean lexer, without knowing what it does... Now, I have looked up the doc for `MULTILINE`, and I tested the code. I don't know why `MULTILINE` was used in Lean.

Anteru:
Probably a bug in Lean as well. Would you mind adding an unit test to make sure one case you care about is working correctly now?

PR #1651, opened 2020-12-28
Update keywords for JS and its variants

Mesteery:
Some of the keywords are missing for TypeScript.
There are also non-existent builtins that I removed.

### DIFF
--- a/pygments/lexers/javascript.py
+++ b/pygments/lexers/javascript.py
@@ -67,7 +67,7 @@ class JavascriptLexer(RegexLexer):
 
             # Numeric literals
             (r'0[bB][01]+n?', Number.Bin),
-            (r'0[oO]?[0-7]+n?', Number.Oct),  # Browsers support "0o7" and "07" notations
+            (r'0[oO]?[0-7]+n?', Number.Oct),  # Browsers support "0o7" and "07" (< ES5) notations
             (r'0[xX][0-9a-fA-F]+n?', Number.Hex),
             (r'[0-9]+n', Number.Integer),  # Javascript BigInt requires an "n" postfix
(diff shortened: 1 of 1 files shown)


pygments/lexers/javascript.py
```diff
@@ -82,15 +82,15 @@ class JavascriptLexer(RegexLexer):
...
             (r'(for|in|while|do|break|return|continue|switch|case|default|if|else|'
              r'throw|try|catch|finally|new|delete|typeof|instanceof|void|yield|await|async|'
-             r'this|of)\b', Keyword, 'slashstartsregex'),
-            (r'(var|let|with|function)\b', Keyword.Declaration, 'slashstartsregex'),
-            (r'(abstract|boolean|byte|char|class|const|debugger|double|enum|export|'
-             r'extends|final|float|goto|implements|import|int|interface|long|native|'
-             r'package|private|protected|public|short|static|super|synchronized|throws|'
+             r'this|of|import|export|static|super|extends|debugger)\b', Keyword, 'slashstartsregex'),
+            (r'(var|let|const|with|function|class)\b', Keyword.Declaration, 'slashstartsregex'),
+            (r'(abstract|boolean|byte|char|double|enum|'
```
Anteru, line 87: This matches the empty string as you have `|` at the end of the line _and_ as the start of the next line. Please don't start lines with `|`, then this should be good to go.

Anteru (changes requested):

pygments/lexers/javascript.py
Mesteery: Yes I know, I was going to correct that, but I had gone out to eat, and so I couldn't.

PR #1658, opened 2020-12-30
Fix eiffel regex performance

brondsem:
I encountered a certain eiffel file that performed extremely poorly with this regex.  (I can share an example if desired, I'm not sure what proper etiquette is).

This change avoids the performance problem, and output is unchanged for the `example.e` file and a few other .e files I tested.  I am not very experienced with eiffel language syntax, but as far as I can tell this is an okay change to make.


birkenfeld (approved):
This change is totally fine, `[^\]%]` already includes `\n`.

PR #1659, opened 2021-01-03
Add Zenburn theme

sirex:
Zenburn theme home page: https://kippura.org/zenburnpage/

And Vim Zenburn theme repository, from which I took color codes: https://github.com/jnurmine/Zenburn

Screenshot from 2021-01-04 01-18-15

### DIFF
--- a/pygments/styles/zenburn.py
+++ b/pygments/styles/zenburn.py
@@ -28,6 +28,8 @@ class ZenburnStyle(Style):
     highlight_color = '#484848'
     line_number_color = '#5d6262'
     line_number_background_color = '#353535'
+    line_number_special_color = '#7a8080'
+    line_number_special_background_color = '#353535'
     styles = {
         Token: '#dcdccc',
         Error: '#e37170 bold',
--- a/pygments/styles/zenburn.py
+++ b/pygments/styles/zenburn.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 """
     pygments.styles.zenburn
     ~~~~~~~~~~~~~~~~~~~~~~~
--- a/pygments/styles/zenburn.py
+++ b/pygments/styles/zenburn.py
@@ -1,0 +1,73 @@
+# -*- coding: utf-8 -*-
+"""
+    pygments.styles.zenburn
+    ~~~~~~~~~~~~~~~~~~~~~~~
+
+    Low contrast color scheme Zenburn.
+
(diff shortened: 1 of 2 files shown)


pygments/styles/zenburn.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+
+    default_style = ""
+    background_color = '#3f3f3f'
+    styles = {
+        Token: '#dcdccc',
+        Error: '#e37170 bold',
+
+        Keyword: '#efdcbc',
+        Keyword.Type: '#dfdfbf bold',
```
Anteru, line 34: Can you please add line number styles here as well? There are four of those in total that should be defined.

Anteru (changes requested):

sirex:
pygmentize -f html -O style=zenburn,full,linenos=True,hl_lines=9 test.py -o test.html

Screenshot from 2021-02-06 21-27-52

Anteru:
Thanks! There are only two more colors missing, for special line numbers. You can test that with `lineospecial=2`, the color names are `line_number_special_color` and `line_number_special_background_color`. Those are used to highlight every N-th line, so you probably want something a tad brighter for them.

pygments/styles/zenburn.py
```diff
@@ -0,0 +1,78 @@
...
+    """
+    Low contrast Zenburn style.
+    """
+
+    default_style = ""
+    background_color = '#3f3f3f'
+    highlight_color = '#484848'
+    line_number_color = '#5d6262'
+    line_number_background_color = '#353535'
+    styles = {
```
Anteru, line 31: Please add colors for the `_special_` variant as well.

Anteru (changes requested):

sirex:
Added `_spcial_` colors for line numbers. But something does not work with HTML generator, because just by adding `linenospecial=2` does not work. It only worked for me by also adding `noclasses=True`. It seems, that generated HTML and CSS styles does not line up.

    pygmentize -f html -O style=zenburn,full,linenos=True,linenospecial=2,noclasses=True,hl_lines=9 test.py -o /tmp/test.html


Screenshot from 2021-02-07 00-02-22

paw-lu:
Thanks @sirex! I was going in circles trying to figure out why the special formatting wasn't applying. Thanks for documenting your temporary solution.

Anteru:
The class issue is fixed in #1701, I think. Thanks for your contribution!

PR #1662, opened 2021-01-05
Add Material theme style

### LINKED ISSUE #1661: Add material theme style
The [Material theme](https://github.com/material-theme/vsc-material-theme) is a popular style that could be added to pygments.

paw-lu:
Adds a material style based on the [popular theme of the same name](https://github.com/material-theme/vsc-material-theme).

Closes #1661


pygments/styles/material.py
```diff
@@ -0,0 +1,118 @@
...
+    violet= '#BB80B3'
+    foreground = '#EEFFFF'
+    faded = '#546E7A'
+
+    default_style = ""
+    background_color = dark_teal
+    highlight_color = '#507174'
+    line_number_color = '#37474F'
+    line_number_background_color = dark_teal
+    line_number_special_color = faded
```
Anteru, line 44: Can you please set `line_number_special_background_color` as well?

Anteru (changes requested):

pygments/styles/material.py
paw-lu: Good catch! Done!

paw-lu:
Adjusted the line numbers and highlight colors a bit to be more faithful to the original theme.

Ran into the same issue as @sirex on #1659—where I had to set `noclasses=True` to preview the special formatting.

Anteru:
Ok, I think the class issue is fixed in #1701. I'll merge the theme for now, thanks for your contribution, and going forward nobody should have to suffer like you did! Thanks!

Issue #1663, opened 2021-01-05
Issue with AutoIT Parser


trdean1:
I've noticed something the AutoIT parser gets wrong.  The following line of AutoIT code

`$foo = 'bar; baz'`

will get tokenized as 

```
[(Token.Name.Variable, '$foo'),
 (Token.Text, ' '),
 (Token.Operator, '='),
 (Token.Text, ' '),
 (Token.Text, "'"),
 (Token.Name, 'bar'),
 (Token.Comment.Single, "; baz'\n")]
```

The correct tokenization should be 

```
[(Token.Name.Variable, '$foo'),
 (Token.Text, ' '),
 (Token.Operator, '='),
 (Token.Text, ' '),
 (Token.Literal.String, '"'),
 (Token.Literal.String, 'bar; baz'),
 (Token.Literal.String, '"'),
 (Token.Text, '\n')]
```

I noticed the tokenization is done correctly if the string is enclosed in double quotes rather than single quotes, but both should be valid autoit syntax

trdean1:
Suggested fix for this issue in pull request #1667

PR #1667, opened 2021-01-07
Supports single quote strings in AutoItLexer

trdean1:
Fix for issue #1663


Anteru:
Can you please add a test for it?

trdean1:
Done - I have added a line to `tests/examplefiles/autoit/autoit_submit.au3` that tests this case and updated the corresponding output file after verifying the output from pytest was correct.

Anteru:
Thanks. Something doesn't look right though: The `'` appears both as `Token.String`, and then again with the rest of the string. That doesn't seem quite right to me, as you'd get `''boo; bar; "baz"''` when you paste them back together. Can you please double check that the output looks correct?

trdean1:
I think it looks correct to me. The `'`'s on line 177 of `autoit_submit.au3.output` are just displaying the token as a string.  Compare to lines 14-16 which holds the output of `"http://localhost:8080"`.  This becomes
```
'"'   Literal.String
'http://localhost:8080'  Literal.String
'"'   Literal.String
```

Anteru:
Ah, you're right, my bad! Thanks, merged for the next release!

PR #1670, opened 2021-01-08
Add __END_LINE__ to Crystal syntax highlighting

rhysd:
I found `__END_LINE__` magic constant is missing in syntax highlighting for Crystal language.

https://crystal-lang.org/reference/syntax_and_semantics/constants.html

### DIFF
--- a/pygments/lexers/crystal.py
+++ b/pygments/lexers/crystal.py
@@ -171,6 +171,7 @@ class CrystalLexer(ExtendedRegexLexer):
             '''.split(), suffix=r'\b'), Keyword),
             (words('''
                 previous_def forall out uninitialized __DIR__ __FILE__ __LINE__
+                __END_LINE__
             '''.split(), prefix=r'(?<!\.)', suffix=r'\b'), Keyword.Pseudo),
             # https://crystal-lang.org/docs/syntax_and_semantics/is_a.html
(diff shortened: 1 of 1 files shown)


rhysd:
I want to have review by @oprypin

pygments/lexers/crystal.py
```diff
@@ -171,6 +171,7 @@ def gen_crystalstrings_rules():
             '''.split(), suffix=r'\b'), Keyword),
             (words('''
                 previous_def forall out uninitialized __DIR__ __FILE__ __LINE__
+                __END_LINE__
```
oprypin, line 174: The way that the lines are split looks weird, but this is technically the most correct way, so nothing to do here.

oprypin (approved):
Good find!

pygments/lexers/crystal.py
rhysd: I'm not familiar with code style of this project, but all lines seem to be limited within 90 characters. If this is put at L173, L173 will exceed 90 chars. Please let me know if my understanding is not correct.

Anteru: Our "desired" target line length is 90 characters, so this looks fine. The 90 character rule hasn't been enforced but it's a general guideline.

Anteru:
Merged, good catch!

Issue #1188, opened 2019-08-31
Empty spaces in fortran lexer keyword recognition


Anteru:
(Original issue 1484 created by anphi on 2018-11-14T20:37:47.258741+00:00)

Fortran allows arbitrary empty space to be embedded in keywords.
For example, it is legal to write "e n d     d   o"

While probably nobody would do this, and we don't want to request the lexer to handle that, some people run together certain multi-word keywords, such as "enddo" instead of "end do". The current Fortran lexer does not recognize such them. Could you please make any inter-word space for keywords optional, so that both versions are matched?

PR #1677, opened 2021-01-13
Fotran lexer: detect keywords formed by two keywords without inter-wo…

ecasglez:
…rd spaces.

* Fix for issue #1188.
* I took the list of keywords affected by this issue from the VIM fortran lexer.
* All the affeced keywords are ending keywords, that is, "end" followed by another keyword without space.


Anteru:

PR #1683, opened 2021-01-14
Fix for lexing Python raw f-strings with backslashes

### LINKED ISSUE #1681: Error parsing Python raw f-string with backslash
Python(3) code using raw f-strings containing a backslash within escaped braces (i.e. double braces) cannot be parsed. A minimum example is

```python3
fr'{{\S'
```

Some pointers:
* Works: `f'{{\S'`.
* Works: `r'{{\S'`
* Works: `rf'{{'`.
* Works: `rf'\S'`.
* Also do not work: `rf'{{\S'`.
* The `S` is not important, though some character must be there in order for the syntax to be valid.

This might seem like a pretty esoteric case. For me it comes up a lot when writing LaTeX in Matplotlib.

jmd-dk:
Fix for #1681

* Introduce a new `'rfstringescape'`, used for raw f-strings.
* Build `'fstringescape'` as a composite of `'rfstringescape'` and `'stringescape'`.

### DIFF
--- a/pygments/lexers/python.py
+++ b/pygments/lexers/python.py
@@ -122,13 +122,17 @@ class PythonLexer(RegexLexer):
         'expr': [
             # raw f-strings
             ('(?i)(rf|fr)(""")',
-             bygroups(String.Affix, String.Double), 'tdqf'),
+             bygroups(String.Affix, String.Double),
+             combined('rfstringescape', 'tdqf')),
             ("(?i)(rf|fr)(''')",
-             bygroups(String.Affix, String.Single), 'tsqf'),
+             bygroups(String.Affix, String.Single),
+             combined('rfstringescape', 'tsqf')),
             ('(?i)(rf|fr)(")',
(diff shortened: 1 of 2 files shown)


Anteru:
Thanks for preparing the fix! Can you please add an unit test for this as well?

jmd-dk:
> Thanks for preparing the fix! Can you please add an unit test for this as well?

I've added a small test which checks for the previously missing "raw" behavior within double braces of raw f-strings.

jmd-dk:
@Anteru Note that I've rewritten the old `'fstringescape'` in terms of the new `'rfstringescape'`, as to not have to have the code
```python
(r'\{\{', String.Escape),
(r'\}\}', String.Escape),
```
twice. Is this done correctly?

Anteru:
Sounds reasonable to me, thanks for the contribution!

birkenfeld (approved):
LGTM.

Issue #1684, opened 2021-01-17
Matlab OOP support doesn't seem great


apjanke:
Hi Pygments! Thanks for making a really useful library.

I'm a Matlab programmer and I'm using Pygments along with AsciiDoc and Jekyll to generate documentation for my project.

It doesn't look like Pygments is doing a good job of supporting modern Matlab OOP language features like classdefs and `argument` blocks. Here's what the online Pygments demo looks like:

![image](https://user-images.githubusercontent.com/2618447/104827973-573a9c00-5832-11eb-84a1-4da213d9fe84.png)

That could use some work, IMHO. Here's what the same code looks like in Matlab R2021a itself:

![image](https://user-images.githubusercontent.com/2618447/104827982-6883a880-5832-11eb-90bc-9592b31e5066.png)

Would you entertain a PR to improve this? This is something I could probably do myself.

apjanke:
See also #1685.

Anteru:
Yes, we'd definitely be interested in a PR for this! Looks like a bunch of keywords is missing at the very least here.

apjanke:
#1466 has been idle since August; I guess I don't need to worry about coordinating my PR with that one? I could just include all its functionality when I do mine, using a `git pull` from that PR so the original author gets credit.

alexpovel:
For integration into [`listings`](https://ctan.org/pkg/listings), I had compiled a list of pretty much all built-in Matlab functionalities. I used <https://www.mathworks.com/help/matlab/referencelist.html> for that. I thought these keywords/names of built-in functions etc. could be slapped into e.g. here:

https://github.com/pygments/pygments/blob/d09fd511aa4484ae5ac0196041bad1ecba00f5f8/pygments/lexers/matlab.py#L51-L74

(Don't know anything about pygments!)

The list I have is sorted into six rough categories (language keywords, built-functions etc.). It contains approximately 2500 items. Let me know if that would be of help here.

Anteru:
That would help for sure to improve the precision. @alexpovel Can you please create a PR which _just_ updates the keywords?

crisluengo:
@apjanke highlight.js (used by Stack Overflow) has a much simpler parsing for MATLAB than Pygments, and seems to do quite well with MATLAB's OOP. Maybe this lexer here just needs to be simplified, rather than extended?

https://github.com/highlightjs/highlight.js/blob/master/src/languages/matlab.js

apjanke:
[...]

PR #1685, opened 2021-01-17
Add modern Matlab OOP examples

apjanke:
Want some examples of modern Matlab OOP syntax?


Anteru:
Thanks for getting the ball rolling here! Ideally this should go together with a test case and a lexer PR, because on it's own this file doesn't help that much.

apjanke:
I got you, babe. See #1684. I'll try to roll up a PR with some actual support here.

Gonna take me a week or so, prob. Busy at the day job.

PR #1686, opened 2021-01-17
fix escape in TypeScript templates literals

Mesteery:


birkenfeld:
Thanks, I missed that.

PR #1687, opened 2021-01-17
add markdown to aliases

Mesteery:
It would be interesting and useful to have markdown as aliases.


birkenfeld:
You also need to regenerate the map (`make mapfiles`).

birkenfeld:
Thanks!

PR #1688, opened 2021-01-18
Move test_examplefiles to new tests/lexers scheme.

birkenfeld:


Anteru:
I'll take a look tomorrow, I don't feel like I'll be able to review 1.2 million new lines today :)

birkenfeld:
Note: it looks like PyPy3 has an incompatibility in how repr() of some high codepoints work (CPython keeps them literally, PyPy escapes them). This let the tests fail before.

Fortunately there was only one instance of this in the examplefiles, and not in a semantically important position (it was in a comment).

So for now I deleted the comment.  If we ever get a PR where high codepoints are a problem with PyPy, we might have to change the repr() for ascii() or use custom escaping.

I'll also try to report this to PyPy.

birkenfeld:
https://foss.heptapod.net/pypy/pypy/-/issues/3384

oprypin:
Hmm I think it was quite nice that the example files had the language's extension, and so could just be viewed in a text editor or GitHub UI for posterity. Also seems like Git is not doing move detection here, which is a minor pity. Also I *really* think the examplefiles shouldn't be mixed with lexer tests, which are actually trying to point out some particular scenario, particularly during codereview with diffs (seems like the example files can actually come in the middle of tests in terms of order)

In my view, *this* would actually warrant a separate file for input & output. *Some* different handling at least.

oprypin:
As per my comment, I have submitted my take on this in #1689. Much less destructive, I think.

oprypin:
If my PR is not to your liking, please let me reserve the right to review this one before you merge :)

Anteru:
@birkenfeld I think the separate file approach has some merit here, if only to allow editors to syntax highlight while editing which makes it more likely the example files will be of higher quality. I'm not 100% convinced of having both the two-file approach and the one-file approach simultaneously is a great idea. I think we should add some guidelines when to use one over the other: I.e. for tests with comments etc. use the merged file, for larger examples use separate files, but that needs to be documented somewhere.

oprypin:
I think maybe these examplefiles just shouldn't have their tokens dumped. I was actually surprised that you went with that.

Now that it's easy, people can "organically" populate a lot more nice individual examples in the "tokens" folder, and it will outgrow the examplefiles eventually, making them unnecessary during review-time.

Anyway, the distinction isn't "small" vs "big", it's "hand-picked samples that try to show some particular point" vs "a low quality random dump of code". By putting them together, you likely lower the overall quality of future additions. And @Anteru already mentioned another great point about quality.

Whatever happens, I strongly believe these shouldn't be equated, especially in the same folder.

birkenfeld:
> I think maybe these examplefiles just shouldn't have their tokens dumped. I was actually surprised that you went with that.

Why not? I think it makes sense to have to regenerate these after changing a lexer, and to go through the changes to make sure everything is as intended.

> Now that it's easy, people can "organically" populate a lot more nice individual examples in the "tokens" folder, and it will outgrow the examplefiles eventually, making them unnecessary during review-time.

I don't think so - a larger file containing a lot of different syntax is much easier for contributors to procure than "many individual examples".  Small examples usually come later, when

> Whatever happens, I strongly believe these shouldn't be equated, especially in the same folder.

And I'm not seeing the big difference...

PR #1689, opened 2021-01-19
Also add auto-updatable output-based tests to examplefiles

oprypin:
Closes #1688


birkenfeld:
Ok, I like this one. Thanks!

PR #1691, opened 2021-01-20
Add Futhark lexer

Munksgaard:
The purpose of this PR is to add support for the [Futhark](https://futhark-lang.org/) programming language.

I've tried to follow the guide on https://pygments.org/docs/lexerdevelopment/, but please let me know if there's something else I need to do.

### DIFF
--- a/pygments/lexers/_mapping.py
+++ b/pygments/lexers/_mapping.py
@@ -169,6 +169,7 @@ LEXERS = {
     'FortranLexer': ('pygments.lexers.fortran', 'Fortran', ('fortran',), ('*.f03', '*.f90', '*.F03', '*.F90'), ('text/x-fortran',)),
     'FoxProLexer': ('pygments.lexers.foxpro', 'FoxPro', ('foxpro', 'vfp', 'clipper', 'xbase'), ('*.PRG', '*.prg'), ()),
     'FreeFemLexer': ('pygments.lexers.freefem', 'Freefem', ('freefem',), ('*.edp',), ('text/x-freefem',)),
+    'FutharkLexer': ('pygments.lexers.futhark', 'Futhark', ('futhark',), ('*.fut',), ('text/x-futhark',)),
     'GAPLexer': ('pygments.lexers.algebra', 'GAP', ('gap',), ('*.g', '*.gd', '*.gi', '*.gap'), ()),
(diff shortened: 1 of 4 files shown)


Munksgaard:
Oh my, perhaps I overdid it by adding so many examples...

Anteru:
Hi, yes, I think that's a bit much on the example side. Could you please try to condense it to 1-2 files at most? Please keep in mind that all code you submit here must be yours or BSD compatible, and I do get nervous when I see large amounts of code getting submitted :)

Munksgaard:
Hi @Anteru, thanks for your comment. I have added just a single example, that I've written myself for the occasion.

birkenfeld (changes requested):
Thanks for the update! Unfortunately quite a few of the tokens generated by the examplefile are `Error`s :)

(The current way examplefiles are handled is pretty new, which is why this wasn't caught by the test. We'll get a fix whipped up for that.)

Munksgaard:
> Thanks for the update! Unfortunately quite a few of the tokens generated by the examplefile are `Error`s :)

Ah, yes, I can see that. I didn't know that it needed to classify everything correctly. I'll try to find some time to address this.

Munksgaard:
@birkenfeld I've tried to update my lexer a bit, but now I'm getting an error in the linter: [overlap in character class](https://github.com/pygments/pygments/pull/1691/checks?check_run_id=1837969367#step:5:8). I don't know what that means, perhaps you can help me understand?

birkenfeld:
@Munksgaard it means that you have a single character multiple times in a character class. One way to do that is simply something like `[aabc]`, the other way is what seems to have happened here:

```
(r'[+-/%=!><|&*^][+-/%=!><|&*^.]*', Operator),
```

The `+-/` is a range from `+` to `/` which includes `+,-./`.  `-` is special in character classes unless it is escaped with `\` or put at the beginning or end of the class. So `[-+...` instead of `[+-...` will fix it.

Munksgaard:
> @Munksgaard it means that you have a single character multiple times in a character class. One way to do that is simply something like `[aabc]`, the other way is what seems to have happened here:
> 
> ```
> (r'[+-/%=!><|&*^][+-/%=!><|&*^.]*', Operator),
> ```
> 
> The `+-/` is a range from `+` to `/` which includes `+,-./`. `-` is special in character classes unless it is escaped with `\` or put at the beginning or end of the class. So `[-+...` instead of `[+-...` will fix it.

Thank you, I've tried that now :)

Anteru:
Merged, thanks!

PR #1693, opened 2021-01-22
cmdline: port to argparse

birkenfeld:


pygments/cmdline.py
```diff
@@ -216,88 +135,68 @@ def _print_list(what):
...
+            parser.print_help(sys.stderr)
             return 2
-
         # print version
         main(['', '-V'])
-        if not args:
-            args = ['lexer', 'formatter', 'filter', 'style']
-        for arg in args:
+        largs = argns.L
+        if not largs:
```
Anteru, line 156: I'd change this to:
```python
items = {'lexer', 'formatter', 'filter', 'style'}
if any(arg not in items for arg in largs):
    parser.print_help(sys.stderr)
# continue as before
```

Rationale: `_print_list` has no error checking, and if you don't set it correctly, it will silently terminate.

pygments/cmdline.py
```diff
@@ -330,24 +229,21 @@ def main_inner(popts, args, usage):
         return 0
 
     # if no -S is given, -a is not allowed
-    if a_opt is not None:
```
Anteru, line 333: Can this be moved up to line 211, right after `S_opt` and `a_opt` is read from `argsn`? It looks like you'd want to handle this early and not somewhere in the middle.

pygments/cmdline.py
```diff
@@ -380,17 +276,13 @@ def main_inner(popts, args, usage):
...
+    if argns.INPUTFILE:
+        if argns.s:
             print('Error: -s option not usable when input file specified',
                   file=sys.stderr)
             return 2
 
-        infn = args[0]
+        infn = argns.INPUTFILE
         try:
             with open(infn, 'rb') as infp:
```
Anteru, line 287: See my comment above about use of `argparse.FileType()`

Anteru (changes requested):
Sorry I didn't get to this earlier myself :(

The change looks good, I just have a few small nitpicks.

pygments/cmdline.py
```diff
@@ -556,22 +448,128 @@ def main_inner(popts, args, usage):
...
+        help='Give options to the lexer and formatter as a comma-separated '
+        'list of key-value pairs. '
+        'Example: `-O bg=light,python=cool`.')
+    operation.add_argument(
+        '-P', metavar='OPTION=value', action='append',
+        help='Give a single option to the lexer and formatter - with this '
+        'you can pass options whose value contains commas and equal signs. '
+        'Example: `-P "heading=Pygments, the Python highlighter"`.')
+    operation.add_argument(
+        '-o', metavar='OUTPUTFILE',
```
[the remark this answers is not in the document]
birkenfeld: I considered this, but the code using the input and output files is quite different so we can't handle them uniformly anyway.

pygments/cmdline.py
```diff
@@ -216,88 +135,68 @@ def _print_list(what):
...
+              'Chajdas and contributors.' % __version__)
         return 0
 
     # handle ``pygmentize -L``
-    L_opt = opts.pop('-L', None)
-    if L_opt is not None:
-        if opts:
-            print(usage, file=sys.stderr)
+    if argns.L is not None:
+        if any(v for (k, v) in vars(argns).items() if k != 'L'):
```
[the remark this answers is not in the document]
birkenfeld: yes, sure.

pygments/cmdline.py
birkenfeld: :+1:

pygments/cmdline.py
birkenfeld: no, otherwise `-a` would never be allowed. The point is to disallow `-a` if `-S` is not present.

pygments/cmdline.py
Anteru: Done

pygments/cmdline.py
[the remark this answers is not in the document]
Anteru: Done. I used a local function to capture `argsn`.

pygments/cmdline.py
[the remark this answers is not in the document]
Anteru: Ok, makes sense!

birkenfeld:
Nice!

Anteru (approved):

PR #1699, opened 2021-02-03
Improve Kotlin lexer

geralt-encore:
There were few issues with previous implementation of the lexer:
- Not handling string interpolation
- Unconditionally treating modifier keywords while they are considered keywords only in some context
- A lot of smaller fixes (e.g. annotations, built-in types and so on)


Anteru:
Merged, thanks a lot for the contribution!

Issue #1149, opened 2019-08-31
Problem with J syntax highlighting


Anteru:
(Original issue 1445 created by None on 2018-06-10T20:01:08.289510+00:00)

For some reason, Pygments does not recognize `?.` as a valid token, highlighting it in red as seen [here](http://pygments.org/demo/6748791/). This is probably just a problem with the lexer, but I'm not sure.

PR #1700, opened 2021-02-05
Fix for lexing J operator ?

amitkummer:
Fix for #1149

I made the lexer match ? as an operator and also added tests for it.


Anteru:
Merged, thanks!

Issue #1698, opened 2021-01-31
sphinx + emphasized lines + line numbering in inline mode = unexpected result


joerg-vandervlies:
Hi,

recently I read that sphinx will change the way lines are numbered in the next versions from table style to inline style. I made a quick test in my project and found that the combination of inline line numbers together with emphasized lines doesn't look right:

![image](https://user-images.githubusercontent.com/28155731/106386341-c5f02b80-63d4-11eb-9c89-cd350455a3f9.png)

As you can see, the emphasized lines are incorrectly wrapped and I couldn't find a quick fix with custom css on my side.

Is this issue addressed correctly to this project and if so, can you confirm it's a bug? Or do I need to address this to the sphinx project?

kind regards
Jörg

Anteru:
If it's broken with Pygments without Sphinx, then it's a bug here, if it's only broken when used within Sphinx it's an issue "over there". Can you try to reproduce this with just using Pygments? I'm not entirely sure what is actually going wrong here ... and it would be good to know if we need to notify the Sphinx folks to use a particular version of Pygments.

joerg-vandervlies:
I think I know now what's going on. Let me show you.

So this is how the snippets used to look in my project. I'm using the rtd theme and the line numbers were generated with a table:
![image](https://user-images.githubusercontent.com/28155731/106396846-d1f6e000-640a-11eb-8a77-0e96f60384e0.png)
Note that the highlighted lines are fully highlighted from left to right, not only the text portion.

When I read that Sphinx will switch to inline numeration within the next two versions I tried it and it looked like this:
![image](https://user-images.githubusercontent.com/28155731/106396937-3f0a7580-640b-11eb-8f4b-0899bedb4c4d.png)

The unexpected line wraps are the result of this small css rule:
```
.rst-content div[class^=highlight] pre .hll {
    display: block;
    margin: 0 -12px;
    padding: 0 12px;
}
```

Its purpose is to make highlighted lines or highlighted blocks look better. The approach was not a problem when the `span` tags with class `hll` had their own lines in the `pre` block, but now that they have to co-exist with the prepended line numbers, it doesn't work as intended any more.

For comparison: this is how the code snippet looks when generated directly with the `pygmentize` command:
![image](https://user-images.githubusercontent.com/28155731/106397218-82191880-640c-11eb-8adf-4e4fdbcd1dbc.png)

```
[...]

PR #1701, opened 2021-02-06
Fix/1698 lineos wrapping

Anteru:
This changes the HTML/CSS a bit so I'd appreciate a second set of eyes. See #1698 for details, but here's the gist:

* This changes the wrapping order such that line numbers are _inside_ the highlighting span, instead of outside
* This changes the default CSS for line numbers to be "neutral" w.r.t. to the rest of the line. This seems to look better on dark styles.

The full fix is to go through all styles individually and set the line number/special line number color to something sensible, but the current solution should work reasonably well.


birkenfeld:
This is bound to mess up someone's custom CSS, but I guess if it fixes issues that is acceptable?

Anteru:
Pretty much yes. But Sphinx seems to be an important enough customer and it's broken right now, so I'd rather fix Sphinx and brake some random CSS.

PR #1703, opened 2021-02-08
Updated _cocoa_builtins.py with types from iOS 14.4 SDK

nemecek-filip:
The current types appear to be from iOS 8.1 SDK which means a lot of new types are missing and therefore are not properly highlighted.

I also added sorting of these types, so next update will have better changelist for the `COCOA_INTERFACES`, `COCOA_PROTOCOLS` and `COCOA_PRIMITIVES`.

I had issues with file `PDFAnnotation.h` which currently has some enconding error, so I added the `try` block to catch this. This file does not contain any types needed to be included in the sets.


Anteru:
Merged, thanks!

PR #1708, opened 2021-02-12
LaTeX: avoid uneven baselines in case highlighting uses \fcolorbox

jfbu:
Without this, the \strut causes the background color to apply to a full
baseline height, then the framing done by \fcolorbox augments beyond
that the box height and as a result the line where this happens will
have extra distance from previous and next lines.

Correct that by reducing the apparent box size by exactly the width of
the frame (\fboxrule).

Also, make the change to \fboxsep local. Reason: if texcomments is True,
some arbitrary LaTeX mark-up can be executed in a later part and the
setting of \fboxsep could modify output; macro ``\<cmdprefix>@bc`` is
executed at top level, cf. ``\<cmdprefix>@do``.

Related:
- https://github.com/sphinx-doc/sphinx/issues/4249

But the https://github.com/sphinx-doc/sphinx/issues/4250 fix is not satisfying, cf https://github.com/sphinx-doc/sphinx/issues/8874, hence this PR.

### DIFF
--- a/pygments/formatters/latex.py
+++ b/pygments/formatters/latex.py
@@ -299,8 +299,8 @@ class LatexFormatter(Formatter):
                 cmndef += (r'\def\$$@tc##1{\textcolor[rgb]{%s}{##1}}' %
                            rgbcolor(ndef['color']))
             if ndef['border']:
-                cmndef += (r'\def\$$@bc##1{\setlength{\fboxsep}{0pt}'
-                           r'\fcolorbox[rgb]{%s}{%s}{\strut ##1}}' %
+                cmndef += (r'\def\$$@bc##1{{\setlength{\fboxsep}{-\fboxrule}'
+                           r'\fcolorbox[rgb]{%s}{%s}{\strut ##1}}}' %
                            (rgbcolor(ndef['border']),
                             rgbcolor(ndef['bgcolor'])))
             elif ndef['bgcolor']:


birkenfeld:
I'll just accept your judgment here, I have no idea anymore about any of the LaTeX stuff :)

jfbu:
To demonstrate the issue use style ``manni`` with a diff file, for example save this
```diff
diff --git a/pygments/formatters/latex.py b/pygments/formatters/latex.py
index bd7ef9fa..542ffdf2 100644
--- a/pygments/formatters/latex.py
+++ b/pygments/formatters/latex.py
@@ -299,8 +299,8 @@ class LatexFormatter(Formatter):
                 cmndef += (r'\def$tc##1{\textcolor[rgb]{%s}{##1}}' %
                            rgbcolor(ndef['color']))
             if ndef['border']:
-                cmndef += (r'\def$bc##1{\setlength{\fboxsep}{0pt}'
-                           r'\fcolorbox[rgb]{%s}{%s}{\strut ##1}}' %
+                cmndef += (r'\def$bc##1{{\setlength{\fboxsep}{-\fboxrule}'
+                           r'\fcolorbox[rgb]{%s}{%s}{\strut ##1}}}' %
                            (rgbcolor(ndef['border']),
                             rgbcolor(ndef['bgcolor'])))
             elif ndef['bgcolor']:
```
to file test.diff, then

```console
$ pygmentize -l diff -f latex -o test.tex -O style=manni,full=true  test.diff
```

then ``pdflatex test`` and one gets this
Capture d’écran 2021-02-12 à 20 14 48

Compare to the Sphinx output (Sphinx 3.4.3, I hope I get it right, because I was madly changing branches in my sphinx clone... master branch would have other font for monospace rendering; this one is courier and the one above must be the default latex one latin modern) under same style ``manni``

Capture d’écran 2021-02-12 à 20 00 16

oh got merged !

jfbu:
@birkenfeld I have added right now illustration. The second picture is one from running Sphinx which applies a fix, but that might not be exactly same because that fix is not correct it ovewrites style colors but here perhaps it is same.

jfbu:
Here is after I git pull the master with merged PR:

Capture d’écran 2021-02-12 à 20 21 44


you see the interline spacing are gone

(still with

```shell
$ pygmentize -l diff -f latex -o test.tex -O style=manni,full=true  test.diff
```

jfbu:
> I'll just accept your judgment here, I have no idea anymore about any of the LaTeX stuff :)

The mark-up of LaTeXFormatter to apply the styles via the `\PY@toks` etc... is indeed TeX magic!

birkenfeld:
Yep, the new version definitely looks better.

PR #1710, opened 2021-02-12
LaTeX: use \@namedef (refactoring)

jfbu:
```
$ latexdef @namedef

\@namedef:
macro:#1->\expandafter \def \csname #1\endcsname
```

And the `@` character is usable here in macro names (the variable with
name definition is authorized to contain for example `\PY@it`)

This makes the output shorter. It makes for a neater
sphinxhighlight.sty...

Related:

- was reported at https://github.com/sphinx-doc/sphinx/issues/8874


birkenfeld:
Yeah, definitely looks nicer.

PR #1711, opened 2021-02-12
LaTeX: \catcode assignment should be terminated

jfbu:
This is almost irrelevant. But ``\catcode`\_=8`` without termination might
let LaTeX expand prematurely its token stream. Of course here in the
context of fancyvrb.sty chances are 99.99% that what comes next does not
expand or expands safely. I did not look in code.

Only a matter of style, to be on the safe side.

### DIFF
--- a/pygments/formatters/latex.py
+++ b/pygments/formatters/latex.py
@@ -342,7 +342,8 @@ class LatexFormatter(Formatter):
                           (start and ',firstnumber=%d' % start or '') +
                           (step and ',stepnumber=%d' % step or ''))
         if self.mathescape or self.texcomments or self.escapeinside:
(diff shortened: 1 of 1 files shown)


jfbu:
> here in the
> context of fancyvrb.sty

I mean that most people use the LaTeXFormatter output with fancyvrb latex package, because the syntax used in clearly targeted at that package.

Actually on looking at its doc now I see it gives an example from which the Pygments is probably inspired, and there is no `\relax` there. However, at Sphinx we will probably drop fancyvrb in future (cf discussion at https://github.com/sphinx-doc/sphinx/issues/8686) to solve some issues with page breaks (already we have to work hard to get wrapping of code lines). So in future the ``codes={...}`` option might end in some other context than fancyvrb code. It would then be better to use the safer syntax I indicated, although,  as I said, this is almost of nil relevance.

birkenfeld:
Thanks!

PR #1712, opened 2021-02-12
Add blazor lexer

gandarez:
This PR adds blazor lexer.

### DIFF
--- a/pygments/lexers/dotnet.py
+++ b/pygments/lexers/dotnet.py
@@ -16,10 +16,10 @@ from pygments.token import Punctuation, \
 from pygments.util import get_choice_opt
 from pygments import unistring as uni
 
-from pygments.lexers.html import XmlLexer
+from pygments.lexers.html import HtmlLexer, XmlLexer
 
 __all__ = ['CSharpLexer', 'NemerleLexer', 'BooLexer', 'VbNetLexer',
-           'CSharpAspxLexer', 'VbNetAspxLexer', 'FSharpLexer']
+           'CSharpAspxLexer', 'VbNetAspxLexer', 'FSharpLexer', 'BlazorLexer']
 
 
 class CSharpLexer(RegexLexer):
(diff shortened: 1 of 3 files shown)


pygments/lexers/dotnet.py
```diff
@@ -704,3 +704,29 @@ def analyse_text(text):
...
+    source code.
+    """
+
+    name = 'Blazor'
+    aliases = ['blazor']
+    filenames = ['*.razor']
+    mimetypes = []
+
+    tokens = {
+        'root': [
```
Jean-Abou-Samra, line 721: You don't have anything to lex HTML in your lexer, is that really intended?

pygments/lexers/dotnet.py
```diff
@@ -704,3 +704,29 @@ def analyse_text(text):
...
+            (r'(@code |@page )', Keyword),
```
Jean-Abou-Samra, line 722: You normally don't want the space as part of the keyword. To achieve that, this rule should read

```python
(r'(@code|@page)( )', bygroups(Keyword, Whitespace))
```

Also, isn't `@code{` allowed?

pygments/lexers/dotnet.py
```diff
@@ -704,3 +704,29 @@ def analyse_text(text):
...
+            (r'.+', using(CSharpLexer)),
```
Jean-Abou-Samra, line 723: This only matches whole lines so it likely won't be able to parse all kinds of C#. I guess you need to push and pop braces to lex the inner content as C# and the rest as HTML.

tests/examplefiles/blazor.razor
```diff
@@ -0,0 +1,20 @@
+<div class="card" style="width:22rem">
```
Jean-Abou-Samra, line 1: This file should be named `tests/examplefile/blazor/blazor.razor` (inside a subdirectory) and you need to commit the test output obtained with `pytest --update-goldens tests/examplefiles/blazor`.

tests/examplefiles/blazor.razor
```diff
@@ -0,0 +1,20 @@
...
+    <div class="card-body">
+        <h3 class="card-title">@Title</h3>
+        <p class="card-text">@ChildContent</p>
```
Jean-Abou-Samra, line 4: The file has constructs that the lexer does not parse. It's fine if the lexer isn't exhaustive (this one appears quite minimal though), but can you at least make it so that everything in the test file is parsed correctly?

Jean-Abou-Samra (changes requested):

Issue #1490, opened 2020-06-30
Weird highlighting of Matlab lexer


duskmoon314:
I'm using minted to highlight my code. I got a very weird highlighting, and it is different from the [demo page](https://pygments.org/demo/#try). Please forgive me if I ignore any issue related to this problem.

I use `pygmentize -f html -O full -O style=rainbow_dash -l Matlab -o test.html exp2_7.m` and get the following html file: 

![image](https://user-images.githubusercontent.com/20477228/86083877-fa61d900-bacd-11ea-94c1-d419cc47c3c4.png)

```html
<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.01//EN"
   "http://www.w3.org/TR/html4/strict.dtd">
<!--
generated by Pygments <https://pygments.org/>
Copyright 2006-2019 by the Pygments team.
Licensed under the BSD license, see LICENSE for details.
-->
<html>
<head>
  <title></title>
  <meta http-equiv="content-type" content="text/html; charset=utf-8">
  <style type="text/css">
/*
generated by Pygments <https://pygments.org/>
Copyright 2006-2019 by the Pygments team.
Licensed under the BSD license, see LICENSE for details.
*/
td.linenos { background-color: #f0f0f0; padding-right: 10px; }
span.lineno { background-color: #f0f0f0; padding: 0 5px 0 5px; }
pre { line-height: 125%; }
body .hll { background-color: #ffffcc }
body  { background: #ffffff; color: #4d4d4d }
body .c { color: #0080ff; font-style: italic } /* Comment */
body .err { color: #ffffff; background-color: #cc0000 } /* Error */
body .k { color: #2c5dcd; font-weight: bold } /* Keyword */
body .o { color: #2c5dcd } /* Operator */
body .ch { color: #0080ff; font-style: italic } /* Comment.Hashbang */
body .cm { color: #0080ff; font-style: italic } /* Comment.Multiline */
body .cp { color: #0080ff } /* Comment.Preproc */
body .cpf { color: #0080ff; font-style: italic } /* Comment.PreprocFile */
body .c1 { color: #0080ff; font-style: italic } /* Comment.Single */
body .cs { color: #0080ff; font-weight: bold; font-style: italic } /* Comment.Special */
body .gd { background-color: #ffcccc; border: 1px solid #c5060b } /* Generic.Deleted */
body .ge { font-style: italic } /* Generic.Emph */
body .gr { color: #ff0000 } /* Generic.Error */
[...]

PR #1705, opened 2021-02-10
Provide (all?) Matlab built-ins

alexpovel:
This PR incorporates the 2500+ long list of the [Matlab builtins](https://de.mathworks.com/help/matlab/referencelist.html). This originated from #1684 ([here](https://github.com/pygments/pygments/issues/1684#issuecomment-774476862)). However, it's somewhat unrelated since it doesn't improve Matlab OOP support as the author there hopes to get done.

I haven't used Matlab in a while but had this list laying around for integration into latex's [listings](https://ctan.org/pkg/listings). But a project like pygments is a much better place I reckon.

Some thoughts:

1. Before submitting the PR I ran `make`, but the `check` target errored out the ass. However, `pytest` alone (on all targets) passes fine (after updating the Matlab goldens accordingly, see 7e70ec0)

    The existing Matlab `examplefiles` are super short, but we can already see improvements there (`Name` -> `Name.Builtin`, e.g. for `rand`):
2. `black` formatting is entirely optional. It's musle-memory and since I didn't see anything about using specific formatters/styles, I just commited that as well. So this is **assuming this project uses no auto-formatter** -- please correct me if wrong. Especially the (long) lists with one item on each line has some advantages:
     1. Number of items is easily seen, just subtract end and beginning line numbers
     2. alphabetical ordering makes it easy to add new entries and spot duplicates (there's no duplicates). With the usual style, when adding a new item, the line might become overly full and necessitate a new line break, leading to super unreadable diffs and lots of manual labour
 3. The list of built-ins is now generated using `words`, it was previously generated "by hand", by just adding the tuples together. This should improve performance thanks to the optimization that offers from:

    https://github.com/pygments/pygments/blob/54f9d2fec71d28b2f783969a34b2134f5f5b8588/pygments/regexopt.py#L26-L91

    However, this also of course produces *massive* regex strings. This is what `regex_opt` returns for the `Keyword` category:

    ```python
    "(?<!\.)(\s*)((break|c(?:a(?:se|tch)|lassdef|ontinue)|dynamicprops|e(?:lse(?:(?:if)?)|nd)|f(?:or|unction)|global|if|methods|otherwise|p(?:arfor|ersistent)|return|s(?:pmd|witch)|try|while))\b"
    ```

    And this is produced for the new `Name.Builtin` category:

    ```python
[...]

PR #1713, opened 2021-02-13
Fix for MATLAB lexer iterpreting command syntax wrong. Fixes #1706.

### LINKED ISSUE #1706: MatlabLexer parses command syntax wrong
I'm on Pygments 2.7.4. I see that some changes have been made to this file (lexers/matlab.py), however this issue has not been fixed (I copied the latest version of the file over into my Pygments installation).

Related issue #1327 and PR # 1328. The bug is introduced in this PR, it seems it was trying to fix a similar case.

Currently, the following code is misinterpreted:
```matlab
a
b = 5;
```
The `a` is seen as a "command syntax" (this is fine), but the command syntax follows on the next line, marking `b = 5` as a string. The command syntax should end at the end of the line.

The following change fixes it in the cases I've examined. [Old line 108](https://github.com/pygments/pygments/blob/4d555d0fffc914a2a4ac9874416cdaaf8f8c9e74/pygments/lexers/matlab.py#L108):
```python
(r'(?:^|(?<=;))(\s*)(\w+)([\s+)(?!=|\(|(?:%s)\s+)' % _operators,
             bygroups(Text, Name, Text), 'commandargs'),
```
New line 108:
```python
(r'(?:^|(?<=;))(\s*)(\w+)([ \t]+)(?!=|\(|(?:%s)\s+)' % _operators,
             bygroups(Text, Name, Text), 'commandargs'),
```

The same line happens in the latest version of the file, at [line number 142](https://github.com/pygments/pygments/blob/54f9d2fec71d28b2f783969a34b2134f5f5b8588/pygments/lexers/matlab.py#L142). The same change fixes the newest version too.

crisluengo:
As I mentioned in #1706, this simple tweak of changing `\s` into `[ \t]` avoids the command syntax spanning lines.

I have added a test for this particular case, and also one that addresses #1490. That one seemed fixed already though, the output was correct already.

However, looking through the other regexes in the MATLAB lexer, it seems that `\s` is used a lot where newlines should really not be accepted. For example, this is currently accepted as correct syntax (it is not!), with `foo` tagged as a function name:
```matlab
function [x]
= foo(y)
```
I'm not sure it is important to fix these cases, as the lever is not intended to identify syntax errors.

### DIFF
--- a/pygments/lexers/matlab.py
+++ b/pygments/lexers/matlab.py
@@ -139,7 +139,7 @@ class MatlabLexer(RegexLexer):
             # `cd ./ foo`.).  Here, the regex checks that the first word in the
             # line is not followed by <spaces> and then
             # (equal | open-parenthesis | <operator><space> | <space>).
-            (r'(?:^|(?<=;))(\s*)(\w+)(\s+)(?!=|\(|%s\s|\s)' % _operators,
+            (r'(?:^|(?<=;))(\s*)(\w+)([ \t]+)(?!=|\(|%s\s|\s)' % _operators,
              bygroups(Whitespace, Name, Whitespace), 'commandargs'),
 
             include('expressions')
--- a/tests/snippets/matlab/test_command_mode.txt
+++ b/tests/snippets/matlab/test_command_mode.txt
@@ -1,12 +1,28 @@
 # MATLAB allows char function arguments to not be enclosed by parentheses
 # or contain quote characters, as long as they are space separated. Test
 # that one common such function is formatted appropriately.
+# Also test that a single word on a line doesn't turn the next line into
+# the argument to a command.
 
 ---input---
 help sin
+
+a
+b = a;
 
 ---tokens---
 'help'        Name
 ' '           Text.Whitespace
 'sin'         Literal.String
+'\n\n'        Text.Whitespace
+
+'a'           Name
+'\n'          Text.Whitespace
+
+'b'           Name
+' '           Text.Whitespace
+'='           Punctuation
+' '           Text.Whitespace
+'a'           Name
+';'           Punctuation
 '\n'          Text.Whitespace
--- a/tests/snippets/matlab/test_elseif.txt
+++ b/tests/snippets/matlab/test_elseif.txt
@@ -1,0 +1,84 @@
+Issue #1490 shows this case to produce a wrong output.
+
+---input---
+if (i == r)
+    j = j + 1;
+elseif (j == 1)
+    i = i + 1;
+else
+    i = i + 1;
+    j = j - 1;
+end
+
+---tokens---
+'if'          Keyword
+' '           Text.Whitespace
+'('           Punctuation
+'i'           Name.Builtin
+' '           Text.Whitespace
+'=='          Operator
+' '           Text.Whitespace
+'r'           Name
+')'           Punctuation
+'\n    '      Text.Whitespace
+'j'           Name.Builtin
+' '           Text.Whitespace
+'='           Punctuation
+' '           Text.Whitespace
+'j'           Name.Builtin
+' '           Text.Whitespace
+'+'           Operator
+' '           Text.Whitespace
+'1'           Literal.Number.Integer
+';'           Punctuation
+'\n'          Text.Whitespace
+
+'elseif'      Keyword
+' '           Text.Whitespace
+'('           Punctuation
+'j'           Name.Builtin
+' '           Text.Whitespace
+'=='          Operator
+' '           Text.Whitespace
+'1'           Literal.Number.Integer
+')'           Punctuation
+'\n    '      Text.Whitespace
+'i'           Name.Builtin
+' '           Text.Whitespace
+'='           Punctuation
+' '           Text.Whitespace
+'i'           Name.Builtin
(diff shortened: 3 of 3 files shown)


Anteru:
Thanks! Could  you please take a look at the merge conflict?

crisluengo:
It seems that someone pushed a commit that causes wrong formatting: what used to be correctly identified as a string is now identified as a builtin. Not happy about this change. This is a commit that adds several thousand "builtins", which is a rather weird concept for MATLAB for two reasons: (1) Most toolboxes cost extra, and so most people will not have many of these "builtins" defined. (2) One can happily redefine these names to be a different function or even a variable. Highlighting code `rand = 4` such that `rand` is marked up as function name is just wrong.

Anyway, I've updated the expected test output to match what the code now does after that commit, but this output is now wrong.

alexpovel:
Those are good points @crisluengo . Maybe the long list of builtins can be moved into a "least priority" sort of category, as mentioned by @Anteru . In regards to your points:

1. That's right, but I thought it might be convenient to have an all-encompassing approach. Most people don't benefit, but if there's no hurt either, I'd argue it can stay. If there are regressions for (most) users, it should go.
2. This behaviour is in line with how pygments handles other cases, e.g. Python highlighting:

    https://github.com/pygments/pygments/blob/22b7ab94f570614c4f37cebc6fbed68326c5ef45/pygments/lexers/python.py#L210-L222

    In that example, if you use e.g. `sum` as a variable name, you're shadowing a built-in. That's discouraged anyway and in the case of that language, people customarily circumvent it using a trailing underscore, `sum_`. From the [demo](https://pygments.org/demo/#try):

    image

    Of course, the big difference here is that the list of builtins of #1705 is much, much longer and might contain too many terms that are both rarely used as builtins but at the same time often used as variables. I don't know about that.

    Other things to consider:

     - the more obscure and less frequently used (maybe because it's part of an expensive extension) a builtin name, the longer it probably also is and therefore the less chance of it conflicting
     - if you shadow built-ins, pygments highlighting it and thus alerting you to the situation (undesired behaviour, less readability) might even be welcome

That being said, the way GitHub does it currently would seem a suitable middle ground. I don't know how it does its highlighting, but it can differentiate between assignment and built-in usage:

```python
reversed = y
y = reversed(x)
```

Backup image:

image

Don't know if such a thing could be feasible for pygments.

Anteru:
That would require to check if it's a function-call use or not (i.e. followed by `(`, I presume that's all GitHub is doing.) My problem here is: I'm not a Matlab user and I can't tell if having all those built-ins makes sense or not :) My gut feeling is that moving them into a separate token type might be the best solution. If someone using Matlab could chime in that would be most helpful.

crisluengo:
The problem is that `reversed(x)` could be an indexing operation on a variable, or it could be a function call, you don't know unless you do a much more complex parsing of the code to identify variables. This, I'm sure, is way out of scope for Pygments.

I don't mean to be too critical of your work, @alexpovel, but I see several problems with such a large list of function names:

1. Who is going to keep it up to date? There's a new MATLAB release twice a year, there are always new functions, sometimes old functions are removed.
2. If we keep a list of only builtins for the latest MATLAB, do we not highlight properly code that was written 5 years ago?
3. Variables can be highlighted as functions just because there are now several thousand names that are always recognized as builtin functions, irrespective of how they're used.
4. Some of the names listed are for functions only defined for specific types (e.g. `width` is a method of the `table` class, but is not defined for any other type). So it depends on the specific set of arguments passed into the function whether it's a builtin function or not. There is no way that Pygments would be able to distinguish these cases. 

Note that it is very common in MATLAB to create variables with the same name as a function in the MATLAB Toolbox, just because there are so many functions it's hard to avoid using their names (nobody knows them all!). Even `i` is in this list (they used to recommend against using it as a variable, but now they have introduced the notation `1i` for the imaginary unit, letting people do the natural thing of using `i` as a loop index). The MATLAB Editor does not highlight names differently if they are standard functions (or even known functions), and does not warn about redefining standard function names. The official MATLAB documentation occasionally uses these names for different purpose.

This issue is compounded by the many, many official toolboxes they sell separately (not sure these are included in the list of names, I assume they are not?).

Python indeed has a very short list of builtins, because all of its standard library is in packages. Pygments does not try to highlight as builtins all the functions in the Python standard library. Most of this new list of MATLAB builtins are comparable to functions in the Python standard library, very few are comparable to Python builtins.

I'm all for including a short list of builtin functions: `size`, `numel`, `length`, `zeros`, `ones`, `eye`, ... I think that the original list was already too long (yet didn't include core functions such as `whos`, `dir`, `exists`, etc.). But I have no suggestion for how to pick this list.

Is there a way of making the highlighting of builtins optional?

I'll ask people in the Stack Overflow chat room for MATLAB to come and share their thoughts on this.

Anteru:
There's no nice way of making things optional short of assigning a special class to it and hope that whoever cares writes a CSS rule for it. I'd rather not have that in the codebase though, as it's bound to bit rot away. If someone can provide a better list of built-in functions that totally makes sense to me, but given I'm not a MATLAB user I rely on experts to figure this one out :) Curious to hear what the SO folks have to say, hopefully we can come to a conclusion here.

PR #1714, opened 2021-02-13
Add apdl and gcode lexers

averter:
Hey. This is a pull request to add an APDL and gcode lexers


pygments/lexers/apdlexer.py
```diff
@@ -0,0 +1,211 @@
...
+    elafung = ["NSEL()","ESEL()","KSEL()","LSEL()","ASEL()","VSEL()","NDNEXT()","ELNEXT()","KPNEXT()","LSNEXT()","ARNEXT()","VLNEXT()","CENTRX()","CENTRY()","CENTRZ()"]
+    elafunh = ["~CAT5IN","~CATIAIN","~PARAIN","~PROEIN","~SATIN","~UGIN","A","AADD","AATT","ABEXTRACT","\*ABBR","ABBRES","ABBSAV","ABS","ACCAT","ACCOPTION","ACEL","ACLEAR","ADAMS","ADAPT","ADD","ADDAM","ADELE","ADGL","ADRAG","AESIZE","AFILLT","AFLIST","AFSURF","\*AFUN","AGEN","AGLUE","AINA","AINP","AINV","AL","ALIST","ALLSEL","ALPHAD","AMAP","AMESH","/AN3D","ANCNTR","ANCUT","ANCYC","ANDATA","ANDSCL","ANDYNA","/ANFILE","ANFLOW","/ANGLE","ANHARM","ANIM","ANISOS","ANMODE","ANMRES","/ANNOT","ANORM","ANPRES","ANSOL","ANSTOAQWA","ANSTOASAS","ANTIME","ANTYPE"]
+
+    tokens = {
+        'root': [
+            (r'!.*\n', Comment),
+            include('strings'),
+            ("(" + "|".join(elafunb) + r')\b',  Name.Class),
```
Anteru, line 180: This should use the `words` function

pygments/lexers/apdlexer.py
```diff
@@ -0,0 +1,211 @@
+import re
```
Anteru, line 1: Missing copyright header

pygments/lexers/apdlexer.py
```diff
@@ -0,0 +1,211 @@
...
+from pygments.lexer import RegexLexer, bygroups, include
+from pygments.token import *
+
+__all__ = ['apdlexer']
+
+
+class apdlexer(RegexLexer):
+    name = 'ANSYS parametric design language'
```
Anteru, line 9: Please add documentation for this lexer here as a docstring, including a link to the language (if possible) and the versionadded macro

Anteru (changes requested):
Hi! Thanks, this seems to be missing a few things:
* Copyright blob
* Comments for the individual lexers
* Example files for testing
* A few changes to the regex rules

pygments/lexers/apdlexer.py
averter: Thanks. Can you give me more information? I am unsure on how to add/use this function.

pygments/lexers/apdlexer.py
averter: Thanks. Changed

pygments/lexers/apdlexer.py
averter: Thanks. Changed.

pygments/lexers/apdlexer.py
averter: Following [these](https://pygments.org/docs/lexerdevelopment/?highlight=words#handling-lists-of-keywords) instructions I've joined all the keywords lists with words, but I am unsure if I did it the right way. I am still a newbie at Python and I did this quite a long time ago, so I can't remember what I wrote very well... Thanks for any help

averter:
I have also added two test files in the main directory. I hope that's where they had to go.

averter:
Removed the example files and put them in the right place. Now all should be good to go.

Anteru:
Can you please run `make mapfiles`? This should fix the failing tests.

averter:
I managed to use the words function correctly now. I've tested the lexer on the two test files successfully and I have created a new commit with the mapfiles. There are still two tests failing though...

Anteru:
Well, you have an invalid escape sequence: `invalid escape sequence \*`. Please double-check your rules. This is related to tokens like `\*VOPER` and I'm not sure what you're trying to match there in the first place.

Anteru:
Please generate golden test outputs (you can see the command line in the test output.)

averter:
> Please generate golden test outputs (you can see the command line in the test output.)

This is the error message that I am receiving. It seems unrelated to the lexer itself. 
```` python
ImportError while loading conftest '/home/myusername/pygments/tests/conftest.py'.
tests/conftest.py:23: in <module>
    import pygments.lexers
E     File "/home/myusername/pygments/pygments/lexers/__init__.py", line 237
E       yield from find_plugin_lexers()
E                ^
E   SyntaxError: invalid syntax
````

**Comment by Anteru** [comment]
Are you using Python3.5 or later? This looks like you're using an outdated version of Python which doesn't know `yield from`.

**Comment by Anteru** [comment]
Thanks, that's looking better. Can you please do the same for gcode? I.e. add an example file and golden output?

**Comment by averter** [comment]
Thank you for your patience. I am not overly happy with the gcode lexer, but I suppose that it can/will be improved by other contributors in the long run.

**Comment by Anteru** [comment]
Thanks for the quick update. Those gcode files are quite massive and highly repetitive, could you please trim them? I suspect 20-30 lines is enough because it seems to be all the same.

**Comment by averter** [comment]
Thank you. It is true that gcode files tend to be quite large, and I agree that this is unnecessary for testing purposes. The latest commit should amend this.

**Comment by Anteru** [comment]
Thanks! Looks like some of your regex can be simplified (see the lint) failure, could you please take a look? There are a few with duplicated values in a class. I.e. `[66788]` = `[678]`, assuming that was indented in the first place :)

**Comment by averter** [comment]
Unfortunately I cannot make sense of that output. I am not familiar with regexlint and the output seems unintuitive. What is `[66788]` = `[678]` ? If regexlint pointed to line numbers in the apdlexer.py file I could , in principle, figure out what are the duplicated values. I would appreciate if you could give me some pointers. Thanks.

**Comment by Anteru** [comment]
Regexlint tells you which regular expression can be simplified. For instance, you can see this in the output:

```
/home/runner/work/pygments/pygments/pygments/lexers/apdlexer.py: (apdlexer:root:pat#5) W117: Overlap in character class: ['6', '8']
  ...SS|RUN|S(?:EL|TIF))|NCLUDE|PE(?:1(?:[66788])|2(?:0|8(?:[89]))|59|60)|VCH...
                                         ^ here
```

There's a class `[66788]` which will match the numbers 6, 7 and 8. There's no reason to repeat it, and `[678]` is equivalent (hence you get the error: "Overlap in character class: 6, 8".) Regexlint points you at that to simplify your regular expression (and maybe you made some mistake when you wrote `[66788]` in the first place?)

You can simply copy/paste the regex from regexlint and find it in the apdlexer.py file, regexlint doesn't do line numbers unfortunately.

birkenfeld:
> You can simply copy/paste the regex from regexlint and find it in the apdlexer.py file, regexlint doesn't do line numbers unfortunately.

That's not quite true: if the regex is an actual regex, it will point to the line number.

In this case, where the regex is compiled by the `words()` function, it doesn't find the line number (but it does tell you the index of the regex in the state).

In short, you have one or more duplicate words in your words() list.

Anteru:
Oh, ok. Should regexlint print that out then? Something like "compiled regex" to indicate it's not a literal?

birkenfeld:
well it already says "(apdlexer:root:pat#5)" to indicate the position.

averter:
> well it already says "(apdlexer:root:pat#5)" to indicate the position.

That makes sense. I was just getting confused by the numbers in the regexlint output. Thanks both.

Anteru:
Merged, thanks!

PR #1715, opened 2021-02-14
Update lists of operators/keywords in Julia and expand/refine highlighting

jmert:
It appears the Julia highlighting rules have not recently been updated. To me it appears the current syntax definitions probably predate Julia v0.7, and much has changed since then. I was motivated in particular to fix parsing of the "broadcast expression" helper macro `@.` which currently isn't recognized, but once I got going, I made many other changes to try to bring the definition up to date.

I'd like to recognize `@fredrikekre` in particular for inspiring me to make changes here based on some of what he did for highlight.js ([blog post](https://fredrikekre.se/posts/highlight-julia/) and [code branch](https://github.com/fredrikekre/highlight.js/tree/fe/julia-unleashed)).

Some of the more notable changes are:

- Updated keywords/builtins list from Julia v1.6 (currently at rc1) and greatly expanded set of operators/symbols which are recognized.
- Operator macros are recognized. In particular, the broadcast expression macro `@.` is now recognized as valid.
  - Unicode suffixes are handled in most contexts (such as infix `+¹` or custom transpose `'ᵀ`).
- Symbol literals are identified in simple contexts.
- Added command triple-quote expressions and special-cased `raw""` strings which do not interpolate.
  - Non-standard string literals such as `mystring"contents"suffix` now identify their prefix/suffix as affixes.
- Type names in unambiguous contexts are highlighted as types. This necessarily can't catch everything (and hence there's still an explicit list of built-in types), but there are many places like type assertions/annotations and `where` expression where names can be highlighted with little chance of being wrong.
- The `julia-repl` highlighting alias is added to match Julia's own convention for use in Markdown documentation files.
- Tests for everything which was changed


Anteru:
Merged, thanks a lot for this huge update!

PR #1722, opened 2021-02-23
Fixes for C and C++ functions and namespaces

### LINKED ISSUE #1561: c++: `has_value()` function marked as generic Name rather than Function.Name
Another change I noticed when bumping from 2.6.1 to 2.7.1, probably as part of <https://github.com/pygments/pygments/commit/fc56ab883a80f6182339d7fbf95e531ff32e4809>.

In the following code:

```cpp
constexpr bool has_value() const noexcept;
```

`has_value` was previously marked as Name.Function, but now it's marked as just Name. I suspect it should still be Name.Function.

### LINKED ISSUE #1719: C++ namespaces lexed as Name and not Name.Namespace
This happens when declaring a namespace:

```cpp
namespace std {}
```

`std` is lexed as Name and not Name.Namespace.

Also I don't see this being done in other lexers, maybe I didn't look enough so correct me if I'm wrong and this behavior is fine, but when prefixing names with their namespaces the namespaces are matched as just a Name :

```cpp
std::optional{}
```

`std` is matched as Name.

amitkummer:
This fixes #1561 and also fixes #1719. 

Fix lexing of function names (#1561) by adding a keywords state for matching keywords, inside and outside functions.
Before this, when a keyword would appear the lexer would go to the statements state, in which functions were not matched, so functions prefixed with keywords were not matched.

- Apart from the c and cpp lexers, I only updated the swig lexer and made sure it's keywords are in the new keywords state, as it was the only lexer that was breaking from this change.

- The new keywords state is (almost) entirely optional on purpose for deriving lexers. They don't benefit from using it themselves. Because of that I think it's a waste of time to incorporate this state in deriving lexers so I decided against it. 

Fix matching functions with a name or return argument that includes a '::' (also #1561) by introducing a second identifiers regex that matches all the previous identifiers, and also '::', and use it to match for function names and return arguments.

- I took the decision to create a second identifiers regex with '::' inside, simply because using the old identifiers regex would hurt
performance massively on every solution I tried to craft. I think this solution is still pretty elegant though, and the issue it fixes is big.

Fix lexing of namespaces (#1719) wherever possible, by adding a namespace state that is entered each time the namespace keyword is matched, and lexes all names matched as namespaces.

Cases this approach doesn't cover:
- Namespaces in using declarations.
- Namespaces that prefix names in random code.

Unfortunately, in both of these cases the names before and after '::' are not always namespaces so these cases can't be covered.

### DIFF
--- a/pygments/lexers/c_cpp.py
+++ b/pygments/lexers/c_cpp.py
@@ -42,7 +42,9 @@ class CFamilyLexer(RegexLexer):
     _intsuffix = r'(([uU][lL]{0,2})|[lL]{1,2}[uU]?)?'
 
     # Identifier regex with C and C++ Universal Character Name (UCN) support.
-    _ident = r'(?:[a-zA-Z_$]|\\u[0-9a-fA-F]{4}|\\U[0-9a-fA-F]{8})(?:[\w$]|\\u[0-9a-fA-F]{4}|\\U[0-9a-fA-F]{8})*'
+    chars = (r'[a-zA-Z_$]|\\u[0-9a-fA-F]{4}', r'\\U[0-9a-fA-F]{8})(?:[\w$]', r'\\u[0-9a-fA-F]{4}', r'\\U[0-9a-fA-F]{8}')
+    _ident = fr'(?:{"|".join(chars)})*'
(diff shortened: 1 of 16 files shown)


Anteru:
Merged, thanks a lot for your contribution! Nicely done!

PR #1724, opened 2021-02-23
encode `str` into `bytes` for `_lookup_ctag` argument;

cykerway:
check source code of `python-ctags` and `python-ctags3` to see what
argument type is expected (answer: `char *`);

guess this is python 2 legacy; but now it should work in python 3;

### DIFF
--- a/pygments/formatters/html.py
+++ b/pygments/formatters/html.py
@@ -877,7 +877,7 @@ class HtmlFormatter(Formatter):
 
     def _lookup_ctag(self, token):
         entry = ctags.TagEntry()
-        if self._ctags.find(entry, token, 0):
+        if self._ctags.find(entry, token.encode(), 0):
             return entry['file'], entry['lineNumber']
         else:
             return None, None


Anteru:
Could you please add a test case exercising this path? I've got no idea where this is triggered :)

cykerway:
I just ran the shipped test file `tests/test_html_formatter.py` and it failed.

~~~ python
def test_ctags():
    try:
        import ctags
    except ImportError:
        # we can't check without the ctags module, but at least check the exception
        assert pytest.raises(
            RuntimeError, HtmlFormatter, tagsfile='support/tags'
        )
    else:
        # this tagfile says that test_ctags() is on line 165, even if it isn't
        # anymore in the actual source
        fmt = HtmlFormatter(tagsfile='support/tags', lineanchors='L',
                            tagurlformat='%(fname)s%(fext)s')
        outfile = StringIO()
        fmt.format(tokensource, outfile)
        assert '<a href="test_html_formatter.py#L-165">test_ctags</a>' \
            in outfile.getvalue()
~~~

You can probably debug if the problematic value in `html.py` has type `str` or `bytes`. It's `str` on my machine and that caused the error.

Anteru:
Thanks, merged!

PR #1725, opened 2021-02-23
use `id` not `name` for line anchors;

cykerway:
`name` attribute on `a` element is obsolete in html5;


Anteru:
Thanks! I think at some point we'll need a "modern HTML" mode for the Html formatter.

PR #1727, opened 2021-02-27
Add formatter for Pango Markup

PhilippImhof:
New formatter for Pango Markup as described in https://developer.gnome.org/pango/stable/pango-Markup.html

Using Pango, the output of this formatter can then be converted to an SVG file. 

<details>
  <summary>Together with https://github.com/ManimCommunity/manim, the output can also be used to create fancy animations.</summary>

```python
from pygments.lexers import JavascriptLexer
from pygments.formatters import PangoMarkupFormatter

code = """
function bla(n, m) {
   console.log(n*m);
}
"""
print(highlight(code, JavascriptLexer(), PangoMarkupFormatter()))
```
yields
```
<tt><span fgcolor="#008000"><b>function</b></span> bla(n, m) {
   console.log(n<span fgcolor="#666666">*</span>m);
}
</tt>
```
and with this one can do
```python
from manim import *
from manimpango import *

class TestCode(Scene):
    def construct(self):
        text = MarkupText(
            """<tt><span fgcolor="#008000"><b>function</b></span> bla(n, m) {
   console.log(n<span fgcolor="#666666">*</span>m);
}
</tt>""")
        self.play(Write(text))
        self.wait()
```
to get:

https://user-images.githubusercontent.com/52650214/109392867-cff74280-791e-11eb-9238-49eeaf094870.mp4

</details>

### DIFF
--- a/pygments/formatters/pangomarkup.py
+++ b/pygments/formatters/pangomarkup.py
@@ -15,9 +15,10 @@ __all__ = ['PangoMarkupFormatter']
 
 
 class PangoMarkupFormatter(Formatter):
-    r"""
+    """
     Format tokens as Pango Markup code. It can then be rendered to an SVG.
     """
+
     name = 'Pango Markup'
     aliases = ['pango', 'pangomarkup']
     filenames = []
@@ -31,7 +32,7 @@ class PangoMarkupFormatter(Formatter):
             start = ''
             end = ''
             if style['color']:
-                start += f'<span fgcolor="#{style["color"]}">'
+                start += '<span fgcolor="#%s">' % style['color']
                 end = '</span>' + end
             if style['bold']:
                 start += '<b>'
--- a/pygments/formatters/pangomarkup.py
+++ b/pygments/formatters/pangomarkup.py
@@ -14,6 +14,17 @@ from pygments.formatter import Formatter
 __all__ = ['PangoMarkupFormatter']
 
 
+_escape_table = {
+    ord('&'): '&amp;',
+    ord('<'): '&lt;',
+}
+
+
+def escape_special_chars(text, table=_escape_table):
+    """Escape & and < for Pango Markup."""
+    return text.translate(table)
+
+
 class PangoMarkupFormatter(Formatter):
     """
     Format tokens as Pango Markup code. It can then be rendered to an SVG.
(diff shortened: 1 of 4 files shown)


pygments/formatters/pangomarkup.py
```diff
@@ -0,0 +1,69 @@
...
+"""
+
+from pygments.formatter import Formatter
+
+
+__all__ = ['PangoMarkupFormatter']
+
+
+class PangoMarkupFormatter(Formatter):
+    r"""
```
Anteru, line 18: Please remove the `r`, not needed here.

Anteru (changes requested):
Can you please add a minimal test case for this?

PhilippImhof:
Yes, I will prepare a test case.

tests/test_pangomarkup_formatter.py
```diff
@@ -0,0 +1,45 @@
...
+
+    nfmt = NullFormatter()
+    noutfile = StringIO()
+    nfmt.format(tokensource, noutfile)
+
+    stripped_markup = re.sub('</?(b|u|i)>', '', poutfile.getvalue())
+    stripped_markup = re.sub('<span[^>]+>', '', stripped_markup)
+    stripped_markup = re.sub('</span>', '', stripped_markup)
+
+    text = noutfile.getvalue()
```
Anteru, line 40: This test file doesn't tell you if the PangoMarkupFormatter did anything right. I'd suggest you pass in a one-line code snippet and check the whole output for correctness. In the current test I could simply return the input text and that would pass.

Anteru (changes requested):

tests/test_pangomarkup_formatter.py
PhilippImhof: Actually, returning just the input text (even wrapped between `<tt>` and `</tt>`) would fail the test, but I will change like you suggested, because that's indeed much better. (Unless the default styles are going to change one day.)

Anteru: They may change, but then it's on me to update this. You could use the BW style though to ensure that the colors remain fixed (because the BW style will not change.)

PhilippImhof: I'd rather keep the default style, because otherwise no `<span fgcolor="..."></span>` comes into play. In order to avoid any problems in the future, I simply strip the various color codes.

Anteru:
Merged, thanks a lot!

PR #1730, opened 2021-03-03
Fix typo in keywords of SquidConfLexer.

imomaliev:
cahce_replacement_policy -> cache_replacement_policy


Anteru:
Good catch, thanks!

PR #1731, opened 2021-03-03
LaTeX: Fix rare `missing number treated as zero`

clslgrnc:
I have a `.tex` document that uses a lot of packages. One of them (or a combination thereof) must mess with the `-` symbol used in `\setlength` (as defined in the calc package), resulting in:
`! Missing number, treated as zero`
on `\PYG{err}{XXX}` with the default highlighting.
Replacing `\setlength{\fboxsep}{-\fboxrule}` with the seemingly more robust `\setlength{\fboxsep}{0pt minus \fboxrule}` fixes the issue.

I do not have a MWE exhibiting the bug.

### DIFF
--- a/pygments/formatters/latex.py
+++ b/pygments/formatters/latex.py
@@ -299,7 +299,7 @@ class LatexFormatter(Formatter):
                 cmndef += (r'\def\$$@tc##1{\textcolor[rgb]{%s}{##1}}' %
                            rgbcolor(ndef['color']))
             if ndef['border']:
-                cmndef += (r'\def\$$@bc##1{{\setlength{\fboxsep}{-\fboxrule}'
+                cmndef += (r'\def\$$@bc##1{{\setlength{\fboxsep}{0pt minus \fboxrule}'
                            r'\fcolorbox[rgb]{%s}{%s}{\strut ##1}}}' %
                            (rgbcolor(ndef['border']),
                             rgbcolor(ndef['bgcolor'])))


birkenfeld:
@jfbu does this look reasonable?

jfbu:
> @jfbu does this look reasonable?

no, `\setlength{\fboxsep}{0pt minus \fboxrule}` is not correct. First it breaks in absence of `calc` package. Second in presence of `calc` package it does not raise an error but computes an erroneous result of setting `\fboxsep` to `0pt`, which is not what we want. So it may perhaps allow the build but it destroys what we want to do.

The `0pt minus ... plus... ` is core TeX syntax for setting so-called  skip variables which have some amount of stretch and shrink. The variables defined by LaTeX  `\newlength` are of this type but `\fboxsep` is a dimen variable, not a skip variable.

```
$ latexdef fboxsep

\fboxsep:
\dimen68


\the\fboxsep:
3.0pt
```

Similarly `\fboxrule` is a dimen variable.

Anyway, the core TeX syntax would be to simply issue `\fboxsep-\fboxrule` or `\fboxsep=-\fboxrule` if one wants to be more human-readable.

I am curious to know what is the cause of the breakage. 

```
\documentclass{article}
\usepackage{calc}
\begin{document}

%\show\fboxsep

\showthe\fboxsep

\setlength{\fboxsep}{0pt minus \fboxrule}% BAD !

\showthe\fboxsep

\fboxsep3pt

\showthe\fboxsep

\setlength{\fboxsep}{-\fboxrule}

\showthe\fboxsep

\fboxsep 0pt

\showthe\fboxsep

\fboxsep-\fboxrule

\showthe\fboxsep

\end{document}
```

when executing latex on this document the log or terminal output will show:

```
> 3.0pt.
l.7 \showthe\fboxsep
                    
> 0.0pt.
l.11 \showthe\fboxsep
                     
> 3.0pt.
l.15 \showthe\fboxsep
                     
> -0.4pt.
l.19 \showthe\fboxsep
                     
> 0.0pt.
l.23 \showthe\fboxsep
                     
> -0.4pt.
l.27 \showthe\fboxsep
                     
```

clslgrnc:
Wow, I did not expect such a detailed explanation so fast: Thanks !

I'll just fix my pipeline with `\setlength{\fboxsep}{0pt}` for now. I do not believe this repository is the place to open an issue, this was just a failed attempt at making pygments more robust.

If I ever build an MWE I'll post it in the relevant place and put a link in this closed PR if you are curious about it.

Best

jfbu:
@clslgrnc Your issue is probably the same as #1734. Can you confirm you use `minted` and if yes, re-open this?

clslgrnc:
Indeed, I use minted. And realize now that the machine I was trying to build an MWE from used pygments 2.7.4.
I can confirm that the machine exhibiting the bug used minted and pygments 2.8.0.

I can not reopen this PR as I already suppressed my fork.

jfbu:
@clslgrnc Thanks for confirmation. As explained in #1737, there is a specificity in minted's way of turning the Pygments latex stylesheets. As you actually initially suggested it does something funny and unexpected to character `-` which becomes unusable in standard syntax for numbers and dimensions, as appeared in Pygments 2.8.0. As I did not test with minted and use Pygments only via Sphinx I could not initially reproduce your issue (which arises only in case some token highlight uses "border", which is the case in many styles for error tokens). I understand a 2.8.1 release will let Pygments work around this minted idiosyncrasy which had so far not caused trouble.

Issue #1734, opened 2021-03-05
Latex unable to render


mtheiss:
In 2.8.0 Latex is unable to render a simple document (pdflatex -halt-on-error -shell-escape demo.tex) like
```latex
\documentclass{article}

\usepackage{minted}

\begin{document}
    X \mintinline{http}|abc| X
\end{document}
```
with the message:
```
! Missing number, treated as zero.
<to be read again> 
                   -
l.2 \PYG{err}{abc}
```
Might be related to #1710

birkenfeld:
please take a look @jfbu

jfbu:
@birkenfeld 

This is a bug of minted which seems to assign catcode 11 to the `-` character in this context. It is not a bug of Pygments. I will investigate more later the minted aspect of this

(clearly this is the same as #1731 previous report)

but I can't now and only did enough to get to the conclusion I reported.

We could fix it via ``\setlength{\fboxsep}{\string -\fboxrule}`` (mind the space after `\string`)  or TeX syntax `\foxbsep=\string -\fboxrule` (space after `\string`, else du to minted bug, TeX would be a control sequence with name `\string-` where `-` is part of the name).

With the above I get successful build of provided test document. But this should be a temporary hotfix as this is a `minted` bug and should be reported to `minted` maintainer. Should I make PR with it?

birkenfeld:
Thanks so much for the investigation!

PR #1735, opened 2021-03-05
LaTeX: workaround minted incompatibility with Pygments 2.8.0

jfbu:
Fix: #1731
Fix: #1734

Upstream issue: https://github.com/gpoore/minted/issues/294

### DIFF
--- a/pygments/formatters/latex.py
+++ b/pygments/formatters/latex.py
@@ -97,6 +97,10 @@ DOC_TEMPLATE = r'''
 
 STYLE_TEMPLATE = r'''
 \makeatletter
+%% workaround a minted <= v2.5 issue, which loads this file with - having
+%% catcode letter; also with _ having catcode 11 but this does not affect us
+\edef\%(cp)s@restore@catcode@of@minus@sign{\catcode`\noexpand\-=\the\catcode`\-\relax}
+\catcode`\-=12
 \def\%(cp)s@reset{\let\%(cp)s@it=\relax \let\%(cp)s@bf=\relax%%
     \let\%(cp)s@ul=\relax \let\%(cp)s@tc=\relax%%
     \let\%(cp)s@bc=\relax \let\%(cp)s@ff=\relax}
@@ -129,6 +133,7 @@ STYLE_TEMPLATE = r'''
 \def\%(cp)sZlb{[}
 \def\%(cp)sZrb{]}
 \makeatother
+\%(cp)s@restore@catcode@of@minus@sign
 '''
 
 


pygments/formatters/latex.py
```diff
@@ -97,6 +97,10 @@ def escape_tex(text, commandprefix):
 
 STYLE_TEMPLATE = r'''
 \makeatletter
+%% workaround a minted <= v2.5 issue, which loads this file with - having
+%% catcode letter; also with _ having catcode 11 but this does not affect us
+\edef\%(cp)s@restore@catcode@of@minus@sign{\catcode`\noexpand\-=\the\catcode`\-\relax}
+\catcode`\-=12
```
jfbu, line 103: defining the restoring macro is for maximal compatibility but anyhow regarding minted problem with `-` of catcode 11 it seems it happens within a group so it is unneeded to restore at end of this style file the catcode `-` to its (erroneous) minted setting anyhow

pygments/formatters/latex.py
```diff
@@ -129,6 +133,7 @@ def escape_tex(text, commandprefix):
 \def\%(cp)sZlb{[}
 \def\%(cp)sZrb{]}
 \makeatother
+\%(cp)s@restore@catcode@of@minus@sign
```
jfbu, line 136: as commented above we could also decide to go for only ``\catcode`\-=12 `` approach because nobody in his/her right mind should input a TeX/LaTeX file of macros with any other catcode for `-` than 12

birkenfeld:
I think this should definitely be merged - I think tex package release cycles are quite long and we don't want to break minted.

birkenfeld:
@Anteru pick into a branch and make a 2.8.1?

jfbu:
@birkenfeld sorry for not having detected the problem at time of #1708, but I use only Sphinx! :-)

Anteru:
@birkenfeld Yes, this seems important enough.

Anteru:
Cherry picked onto 48f58c9fcf01e1cc48169c102610586348a88a3a which can be used for a 2.8.1 release.

pygments/formatters/latex.py
jfbu, line 103: As minted loads this file with `\endlinechar=-1` setting it would have been better here to add explicitly `\relax` after `12`. This is cosmetic as next line starts with a `\def` which does not expand hence the TeX parsing of the digits will end correctly.

Just mentioning this out of professional sense of duty... (there is no higher level LaTeX interface to set a catcode, so I used the core level TeX syntax, but one has to be aware of some things) (I would have probably used core TeX syntax anyhow anyway as I like it)

birkenfeld: It's certainly interesting to read :laughing:

jfbu:
There was a silly oversight (they all are silly) in my latex code, which I fixed at #1736. Sorry for all the trouble! @birkenfeld

PR #1736, opened 2021-03-05
Fix #1735 pull request

jfbu:
I apologize for the trouble. I made a mistake in #1735, with `\makeatother` too early.


jfbu:
@anteru sorry for oversight on my part.

jfbu:
I am closing this because the commit was incorporated in the one for PR #1737, and #1737 solves #1734 in a more complete way than #1735/#1736 (which create another problem due to actual minted constraints).

PR #1741, opened 2021-03-08
add missing keywords to JS and TS

Mesteery:
`from` keyword was missing (`import something from 'something'`)

### DIFF
--- a/pygments/lexers/javascript.py
+++ b/pygments/lexers/javascript.py
@@ -81,7 +81,7 @@ class JavascriptLexer(RegexLexer):
             (r'[})\].]', Punctuation),
             (r'(for|in|while|do|break|return|continue|switch|case|default|if|else|'
              r'throw|try|catch|finally|new|delete|typeof|instanceof|void|yield|await|async|'
-             r'this|of|static|export|import|debugger|extends|super)\b', Keyword, 'slashstartsregex'),
+             r'this|of|static|export|import|from|as|debugger|extends|super)\b', Keyword, 'slashstartsregex'),
             (r'(var|let|const|with|function|class)\b', Keyword.Declaration, 'slashstartsregex'),
             (r'(abstract|boolean|byte|char|double|enum|final|float|goto'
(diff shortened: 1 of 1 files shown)


Anteru:
`from` is not a keyword though, is it? I was just trying to look up ES5/6 reserved words and from doesn't show up there (source: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar). Is it only contextual?

And should `as` be also supported?

Mesteery:
The `from` (and `as`) is used with ESM [imports](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import)/[exports](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export).

Anteru:
I know, my point is though that from/as are not official keywords, or are they? I'm trying to understand if treating them the same as "import" is correct, given that is reserved. `as` seems to be missing in any case, could you please add that while you're at this?

Mesteery:
> *Those that are always allowed as identifiers, but also appear as keywords within certain syntactic productions, at places where Identifier is not allowed: **`as`**, `async`, **`from`**, `get`, `of`, `set`, and `target`.*

https://tc39.es/ecma262/2020/#sec-keywords-and-reserved-words

Mesteery:
For a more general PR, we could add :
  * [ ] more errors (`Error`, `EvalError`, `InternalError`, `RangeError`, `ReferenceError`, `SyntaxError`, `TypeError`, `URIError`) as `Name.Exception` ?
  * [ ] the `global` object (`Name.Builtin`, Node.js)
  * [ ] more `Name.Builtin` (such as typed arrays, `Intl`, `DataView`, `Set`, `Map`, `WeakSet`, `WeakMap`, `Reflect`, `JSON`, `ArrayBuffer`) ?

Anteru:
Yes, that sounds good, but I'd rather do it in a separate PR. This here is nice and self-contained. Can you please un-draft it so I can merge it, then file another PR for the other changes?

PR #1745, opened 2021-03-11
Update command in doc cmdline.rst 

haphaeu:
Command was not including the line numbers, as described in the text.
Option `linenos=1` was missing.


Anteru:
Nice catch, thanks!

PR #1746, opened 2021-03-12
Refactor CFamilyLexer

amitkummer:
A refactor for #1722. Sorry for the extra work.


Anteru:
What's the reason for making `inline` a reserved keyword vs. a keyword?

amitkummer:
Inline is currently a reserved keyword, see here:

https://github.com/pygments/pygments/blob/d86dd2bbd07ed880d908aeab2a99af50a64142aa/pygments/lexers/c_cpp.py#L105

PR #1747, opened 2021-03-12
Leading underscore in APL variable names (GNU, Dyalog)

xpqz:
Two small tweaks to the APL lexer

1. Add all APL file suffixes
2. Allow leading _ in variable names (GNU APL, Dyalog)


RojerGS:
This PR is more useful than it may seem, as `_←` is a common idiom in APL programming.
There's also the fact that leading underscores are used commonly as the leading character of dop names.

Anteru:
Thanks! Can you please regenerate the mapfiles as well (`make mapfiles`), otherwise the file endings won't get picked up. Would you mind adding/extending one of the APL tests to capture this as well? Not required but would be nice to have so it doesn't regress.

xpqz:
@Anteru Thanks for your comments. I've regenerated the mapfiles and added a testcase (there weren't many for APL...) to demonstrate the modification as per your suggestions.

Anteru:
Merged, thanks a lot!

PR #1748, opened 2021-03-16
Don't mix foreground & background in _deprecated_ansicolors lookup

encukou:
Hello!
A static analyzer alerted us that the wrong variable is used in the `_deprecated_ansicolors` lookup for `bgcolor`.
I'm aware that this is deprecated functionality, but since the code is in the project, you might want to keep it correct.

I looked into writing a test, but I'm new to the codebase and don't know where to start. If you think fixing this issue is worth it, I'd appreciate a pointer.

### DIFF
--- a/pygments/style.py
+++ b/pygments/style.py
@@ -134,7 +134,7 @@ class StyleMeta(type):
             color = _ansimap[color]
         bgcolor = t[4]
         if bgcolor in _deprecated_ansicolors:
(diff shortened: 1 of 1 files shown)


Anteru:
Good catch, thanks! Out of curiosity: Which static analyzer found this?

encukou:
I'm not sure, as I got merged results from a bunch of tools. I assume this was Coverity.

PR #1755, opened 2021-03-23
lexers: devicetree: allow node label without root node

### LINKED ISSUE #1754: Devicetree lexer does not node label without root node
It looks like the `devicetree` lexer throws some syntax warnings when trying to render code snippets like:

```devicetree
foo: foo@1234 {
    a = <1000 2000 3000>; /* array */
    b = [aa bb cc dd];    /* uint8-array */
    c = "bar", "baz";     /* string-array */
};
```

That is, something like `foo: bar@xxx {};`. However, it works if enclosed on a root node `\{};`:

```
/{
    foo: foo@1234 {
        a = <1000 2000 3000>; /* array */
        b = [aa bb cc dd];    /* uint8-array */
        c = "bar", "baz";     /* string-array */
    };
};
```

Ref. https://github.com/zephyrproject-rtos/zephyr/pull/33417

gmarull:
It is sometimes useful to render DeviceTree snippets such as:

```
foo: bar@1234 {
    foo = "bar";
};
```

However, the snippet shown above does not render properly unless it is
enclosed on a root node, i.e.

```
/{
    foo: bar@1234 {
        foo = "bar";
    };
};
```

Fixes #1754

### DIFF
--- a/tests/test_devicetree_lexer.py
+++ b/tests/test_devicetree_lexer.py
@@ -1,0 +1,32 @@
+"""
+    Devicetree Lexer Tests
+    ~~~~~~~~~~~~~~~~~~~~~~
+
+    :copyright: Copyright 2006-2021 by the Pygments team, see AUTHORS.
+    :license: BSD, see LICENSE for details.
+"""
+
+import pytest
+
+from pygments.lexers.devicetree import DevicetreeLexer
+from pygments.token import Token
+
+
+@pytest.fixture(scope="module")
+def lexer():
+    yield DevicetreeLexer()
+
+
+@pytest.mark.parametrize(
+    "fragment",
+    (
+        'nodelabel: node@0 { foo = "bar"; };',
+        'nodelabel: node { foo = "bar"; };',
+        'nodelabel0: nodelabel1: node@0 { foo = "bar"; };',
+    ),
+)
+def test_fragment_out_of_root_node(lexer, fragment):
+    """Validate that a devicetree fragment out of a root node is parsed correctly."""
+
+    tokens = list(lexer.get_tokens(fragment))
+    assert all(x[0] != Token.Error for x in tokens)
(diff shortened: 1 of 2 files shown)


birkenfeld:
Thanks for the PR! Can you add a test snippet to exercise the new behavior?

gmarull:
@birkenfeld done

tests/test_devicetree_lexer.py
```diff
@@ -0,0 +1,25 @@
...
+
+@pytest.fixture(scope='module')
+def lexer():
+    yield DevicetreeLexer()
+
+
+def test_fragment_out_of_root_node(lexer):
+    """Validate that a devicetree fragment out of a root node is parsed correctly."""
+
+    fragment = 'nodelabel: node@0 { foo = "bar"; };'
```
mbolivar-nordic, line 23: Just confirming, but these fragments all work as expected (i.e. are valid)  with this PR, right?

1. `nodelabel: node { foo = "bar"; };`
2. `nodelabel1: nodelabel2: node@0 { foo = "bar"; };`
3. `nodelabel1: nodelabel2: node { foo = "bar"; };`

gmarull: good spot, multi label is not parsed correctly.

gmarull: actually, `nodelabel1: nodelabel2: node { foo = "bar"; };` is parsed correctly even with current implementation

gmarull: current lexer doesn't seem to play well when we have the combination of `label + node + @ + addr` out of root node.

mbolivar-nordic: Makes sense since that's only valid in a snippet.

gmarull: latest changes should fix all the cases mentioned above.

PR #1756, opened 2021-03-24
Support Terraform 0.14 syntax

lgiordani:
This is my second attempt to write a Pygments lexer, and the first one didn't go very well, so apologies in advance if I misused part of the API.

This has been tested against my whole code base which is more than 100k lines of TF code. Even though it doesn't span the whole syntax and resources i thing it's a good test. I cannot publish it, though, so I went for smaller tests that highlight the issues I had.

Let me know if you think that is enough.


lgiordani:
I'm not actually sure what `W119: Instead of negating character class, flip case of builtin class` means in the regexlint output. Any help?

birkenfeld:
Ah yes - the builtin classes like `\s`, `\w` have opposites `\S`, `\W`.

lgiordani:
Oh! That simple :sweat_smile: Thanks, I'm going to fix it

lgiordani:
@Anteru Thanks for merging it!

PR #1759, opened 2021-03-27
Filename to <th> if linenos=table (fixes #1757)

### LINKED ISSUE #1757: HtmlFormatter: filename doesn't work well with linenos=table
When using HtmlFormatter with both the `linenos=table` and `filename=testfile` options, you get some output like that shown below. The problem is that the filename being output within the `<td>` for the code throws off the line number alignment.

```.html
<table class="highlighttable">
  <tr>
    <td class="linenos">
      <div class="linenodiv">
        <pre>...</pre>
      </div>
    </td>
    <td class="code">
      <div class="highlight">
        <span class="filename">testfile</span>
        <pre>...</pre>
      </div>
    </td>
  </tr>
</table>
```

I suggest that this is unlikely to be useful to anyone, since it will invariably throw off the line numbering alignment. Hence, in the special case of someone using `linenos=table` and specifying a `filename` I would suggest that another table row is output as a caption:

```.html
<table class="highlighttable">
  <tr>
    <td colspan="2" class="filename">testfile</td>
  </tr><tr>
    <td class="linenos">
      <div class="linenodiv">
        <pre>...</pre>
      </div>
    </td>
    <td class="code">
      <div class="highlight">
        <pre>...</pre>
      </div>
    </td>
  </tr>
</table>
```

I can look at entering a pull request to implement this, but I wanted to get some feedback on whether this would be acceptable first.

Cartroo:
* Emit `<th>` for `filename` if `linenos=table`.

* Added test cases for filename inclusion.


Anteru:
Thanks, merged! And thanks for the detailed change description as well, much appreciated!

PR #1765, opened 2021-04-05
groovy lexer: detect quoted function names

### LINKED ISSUE #1753: Groovy lexer doesn't handle quoted function names
Noticed when opening a PR on GitHub with Spock framework tests. Spock uses Groovy's support for quoted function names to describe the test in readable language. Unfortunately this is misinterpreted and we lose syntax highlighting for part of the test.

Here is a shorter example which exhibits the problem. The following two snippets should be highlighted the same way:

```groovy
def thisFunctionLooksLikeJava() {
    println "hello"
}

thisFunctionLooksLikeJava()
```

```groovy
def "this is a (long) function"() {
    println "hello"
}

"this is a (long) function"()
```

alisonatwork:
Should fix #1753. I don't have a VM with Python set up in a way where I can easily test this, so I am depending on the tests in CI.

In Groovy it is valid to have a function name which is a quoted string. Single quote (`'`), double quote (`"`) and the triple-single/triple-double formats are all allowed, but I suspect most people only use standard single or double. In most syntax highlighters the quoted function name is displayed as if it were a string (e.g. https://groovyconsole.appspot.com/), which perhaps we could also do by switching the priority to parse strings before function declarations. But since we already detect a function name here, I just try to expand the pattern a little bit to match the quoted string as a function name too.

While we are here, also fix a bug where sometimes keywords would be interpreted as function names (already fixed in Java lexer).

### DIFF
--- a/tests/examplefiles/groovy/test.gradle.output
+++ b/tests/examplefiles/groovy/test.gradle.output
@@ -88,7 +88,7 @@
 ' '           Text
 'String'      Name
 ' '           Text
-'getWorldString' Name.Function
+'getWorldString' Name
 '('           Operator
 ')'           Operator
 ' '           Text
--- a/CHANGES
+++ b/CHANGES
@@ -27,6 +27,7 @@ Version 2.9.0
   * Chapel (#1743)
   * Coq (#1721)
   * Cython (#853)
+  * Groovy (#1765)
   * Julia (#1715)
   * Octave: Allow multiline and block-percent comments (#1726)
   * PowerShell: Improve lexing of ``:`` (#1682, #1758)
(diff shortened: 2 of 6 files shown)


tests/examplefiles/groovy/test.gradle.output
```diff
@@ -88,7 +88,7 @@
 ' '           Text
 'String'      Name
 ' '           Text
-'getWorldString' Name.Function
+'getWorldString' Name
```
alisonatwork, line 91: This should be detected as `Name.Function` and not `Name`, but I think this is not a big problem considering it's much more annoying for keywords to be detected as functions (see `test.groovy.output` fixes, below).

Anteru:
Merged, thanks!

PR #1766, opened 2021-04-06
Allow C-style line comments in F* to start anywhere in a line

SECtim:
F* actually allows for C-style `//` line comments to start anywhere in a line, not just at the beginning.

@denismerigoux @protz Since you came up with the original lexer code (thanks!), I tagged you here so you can take a look as well - maybe I overlooked an obvious problem with starting comments in the middle of a line.

### DIFF
--- a/pygments/lexers/ml.py
+++ b/pygments/lexers/ml.py
@@ -908,7 +908,7 @@ class FStarLexer(RegexLexer):
             (r'\b([A-Z][\w\']*)(?=\s*\.)', Name.Namespace, 'dotted'),
             (r'\b([A-Z][\w\']*)', Name.Class),
             (r'\(\*(?![)])', Comment, 'comment'),
-            (r'^\/\/.+$', Comment),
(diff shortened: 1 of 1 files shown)


SECtim:
The failing tests are actually errors in the test file, e.g.,

https://github.com/pygments/pygments/blob/1417210b1a4d64d6c761605cff07b63b24397282/tests/examplefiles/fstar/example.fst#L242

is expected to be lexed as

https://github.com/pygments/pygments/blob/1417210b1a4d64d6c761605cff07b63b24397282/tests/examplefiles/fstar/example.fst.output#L1993-L2023

Which is not what the actual F* compiler does. Will try to update this PR with a fixed test, but this might take some time, as I have to set up a local testing environment first (and am currently working on something else).

Anteru:
Thanks for the contribution! This adds support for `//` anywhere, and given the example file is incorrect (if I understood you correctly) the fact that `//` is used there as a comment is acceptable for now (even if it's used incorrectly)

PR #1767, opened 2021-04-06
Swift lexer: add 'associatedtype' keyword

lukaskollmer:
Adds the `associatedtype` keyword to the Swift lexer.

I also moved the `init` keyword to the next line to keep them at a roughly even length.


Anteru:

Issue #1740, opened 2021-03-06
Python lexer does not consider .5 as float


khoda81:
In python floats are sometimes written as fractions where decimal part is not written for example:
```python
a = .3 # instead of 0.3
```
But seems like pygments does not consider them as float, considers a dot (operator) and a number.

Code to reproduce:

```python
from pygments import lex
from pygments.lexers.python import PythonLexer
from pygments.token import Number as Number

content = "a = .5"

lineno = 0
for token in lex(content, PythonLexer()):
    print(token)
```
Results:
```
(Token.Name, 'a')
(Token.Text, ' ')
(Token.Operator, '=')
(Token.Text, ' ')
(Token.Operator, '.')
(Token.Literal.Number.Integer, '5')
(Token.Text, '\n')
```

Interestingly does consider 5. as float.

Liam-Stevens:
We are a team of students from the University of Adelaide, we are currently working on this issue. Expect a pull request within the next month.

PR #1768, opened 2021-04-07
Fix python lexer parsing floats incorrectly

Liam-Stevens:
This PR fixes the issue discussed in issue #1740. We've also added some test cases to specifically test the python lexer against floats and integers. Let us know if there's any changes that need to be made.

The issue was that numbers were parsed after operators. This just required rearranging the order in which the tokens are processed.


Anteru:
Merged, thanks!

Alvinchong1989 (commented):
>
Duplicate of #

Anteru:
Duplicate of ...?

Alvinchong1989:
Sorry my mistake.. I don’t know what’s wrong

PR #1772, opened 2021-04-12
Fix capnproto URL

sblondon:
`https://capnproto.com` does not work, so this PR changes to working `https://capnproto.org` URL.


Alvinchong1989 (commented):
-

PR #1775, opened 2021-04-13
AMDGPU: Updated modifier tokenization.

### LINKED ISSUE #1717: AMDGPU Lexer not working on annotations
Example code:

```buffer_load_dword v4, v2, s[8:11], 0 offen```

The token "offen" is not recognized by the lexer.

tsymalla:
This should fix #1717 by adding all possible modifiers , also removes a duplicate part of the previous modifier regex.


Anteru:
Merged, thanks!

PR #1778, opened 2021-04-19
friendly: lighter line numbers

mgeier:
The default color for line numbers looks too intense in comparison with the rest of the colors.


Anteru:
Merged, thanks!

Issue #1718, opened 2021-02-16
Some styles have bad contrasts


not-my-profile:
After encountering a bad color contrast in the syntax highlighting of the [official PIP documentation](https://pip.pypa.io/en/stable/reference/pip_show/#examples),

![image](https://user-images.githubusercontent.com/73739153/108076429-b545e380-706b-11eb-8c10-10c0a005711a.png)

I wrote [a small Python script](https://git.push-f.com/tools/tree/pygments_wcag_check.py) to check how Pygments styles fare when judged according to the [WCAG contrast standard](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html).

The summary of my script is:
```
found 840 contrasts that fail to meet the WCAG AA standard
According to WCAG:
     AA contrast is >= 4.5
    AAA contrast is >= 7.0
    
=== Styles ranked by contrast ===
 avg   min  name
 3.4   2.5  solarized-light
 4.2   1.4  stata-dark
 4.3   2.8  solarized-dark
 5.5   1.3  paraiso-light
 6.2   1.3  paraiso-dark
 6.3   1.0  rrt
 7.0   1.0  inkpot
 7.3   1.7  friendly
 7.5   1.6  perldoc
 7.5   2.1  native
 7.6   1.8  emacs
 7.8   1.3  vim
 7.9   1.4  monokai
 8.0   1.4  manni
 8.7   3.1  lovelace
 8.7   1.8  default
 8.9   2.6  arduino
 9.2   1.9  autumn
 9.3   1.9  pastie
 9.6   1.0  rainbow_dash
 9.9   1.9  trac
10.1   1.4  murphy
10.4   1.0  tango
10.7   1.9  fruity
10.7   1.9  colorful
11.1   1.9  sas
13.0   2.5  abap
13.2   1.9  stata
13.2   1.9  stata-light
13.8   1.9  borland
13.9   5.1  xcode
14.4   3.6  igor
14.7   3.5  algol
14.7   3.5  algol_nu
15.7   3.6  vs
21.0  21.0  bw
```

My script also reports which rules don't meet the AA standard per style sorted by contrast, e.g. for the default style:
```
=== default ===
1.81 Token.Text.Whitespace
2.63 Token.Name.Label
2.68 Token.Name.Entity
3.28 Token.Generic.Inserted
3.34 Token.Generic.Output
3.34 Token.Comment.Preproc
3.36 Token.Name.Attribute
3.68 Token.Literal.String.Interpol
[...]

PR #1779, opened 2021-04-19
native: darker line numbers

mgeier:
Similar to #1778, this makes the line numbers stand out a bit less.

I chose the same color as is used for prompts, which seems consistent to me.

According to Firefox, this is still OK contrast for accessibility (see also #1718).


Anteru:
Merged, thanks!

PR #1780, opened 2021-04-20
uniformize aliases order

Mesteery:
The lexers' have been ordered according to their "quality". For example, if an alias is more common than another or if an alias is more complete than another.
Also, the (dead) link for LiveScriptLexer has been updated.

Finally, I have a question: `LighttpdConfLexer` should have `lighttpd.conf` as filename, right?


Anteru:
Thanks! Yes, the `lighttpd.conf` change sounds right, can you please add this?

Anteru:
Merged, thanks!

PR #1786, opened 2021-04-28
autopygmentize: use file before pygmentize; add JSON support

rrthomas:
Although pygmentize -N is much cheaper than file, it makes some bad guesses,
so use file first.

Add support for MIME type application/json.


Anteru:
Merged, thanks! Also happy to see a contributor return :)

PR #1790, opened 2021-05-04
Add new file extensions to Mathematica lexer

gandarez:
This PR adds missing new file extensions to Mathematica lexer.

https://mathematica.stackexchange.com/questions/63130/whats-the-complete-list-of-file-name-extensions-used-by-mathematica

### DIFF
--- a/pygments/lexers/algebra.py
+++ b/pygments/lexers/algebra.py
@@ -95,7 +95,7 @@ class MathematicaLexer(RegexLexer):
     """
     name = 'Mathematica'
     aliases = ['mathematica', 'mma', 'nb']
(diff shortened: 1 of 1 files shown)


Jean-Abou-Samra (changes requested):
This will not work as-is, you need to regenerate `pygments/lexers/_mapping.py` using `make mapfiles`.

Also, there is an issue with `.mt` files, as these are also declared for the Monte lexer. A `.mt` file will thus never be recognized as a Mathematica file, because Monte comes after Mathematica in alphabetical order. You need to add `analyse_text` methods to both lexers if you want guessing to work.

Issue #1526, opened 2020-08-26
Tracebacks unreadable on dark styles


aroberge:
Python tracebacks include content that is essentially unreadable on dark styles that are included with pygments.

Example traceback:
```pytb
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
NameError: name 'a' is not defined
```
Screenshots below taken from https://pygments.org/demo/#try

monokai:
![image](https://user-images.githubusercontent.com/629698/91289011-15686680-e768-11ea-983b-38d069c2d15b.png)

native:
![image](https://user-images.githubusercontent.com/629698/91289095-3a5cd980-e768-11ea-9c1b-261c24f0719f.png)

fruity:
![image](https://user-images.githubusercontent.com/629698/91289220-6d9f6880-e768-11ea-8957-c6f5f8c7d125.png)

vim:
![image](https://user-images.githubusercontent.com/629698/91289262-827bfc00-e768-11ea-992b-138196cb8d8b.png)

rrt:
![image](https://user-images.githubusercontent.com/629698/91289358-a17a8e00-e768-11ea-8569-08322be22d17.png)

paraiso-dark:
![image](https://user-images.githubusercontent.com/629698/91289467-c242e380-e768-11ea-93f6-b4b8528aba45.png)

solarized-dark:
![image](https://user-images.githubusercontent.com/629698/91289788-2f567900-e769-11ea-81f4-df8a209b95a4.png)

stata-dark:
![image](https://user-images.githubusercontent.com/629698/91289655-01713480-e769-11ea-8ce9-632ef9405fe0.png)

kdeldycke:
Not only tracebacks. It seems all dark themes are affected by this issue:
![Screen Shot 2020-09-01 at 10 05 51](https://user-images.githubusercontent.com/159718/91824425-0b000e00-ec3b-11ea-8c03-0bfa5b17ed21.png)

kdeldycke:
It seems this is due to the lack of `Generic.Output` and `Generic.Prompt` definition in some style. See: https://github.com/pygments/pygments/pull/1529

daroGH:
Hi there!
We are a group of 5 people (Timothy, Guanqi, Chirag, Yuting and Daro) currently doing the course Software Process Improvement at University of Adelaide. As a part of the course, we have to contribute to an open source project and we found that this issue is very interesting to work on. 
We aim to submit our initial PRs by early May
Any guidance would be much appreciated.

Anteru:
[...]

PR #1794, opened 2021-05-07
1526/fusumwan/unreadablestyle

a1714470:
Creating a way to check if a style defines all colors and if those have good contrast. The "check.py" is used to fixed issue 

https://github.com/pygments/pygments/issues/1526

In general, our check.py includes a color similarity function. Such a function is used to list out all the color similarities from one style. Besides, the check function also includes the contrast rato, relative luminance, contrast AA, and contrast AAA. We hope it can check all styles.


Anteru:
Thanks for preparing this. I've got a few comments:

* First, please separate out the style update itself from this change. They're not directly related and if I had to undo one of them, ths would take the other with it.
* There's no helper script to actually run the checks. It would be good to add one and hook it up to `make check`.
* There are styles with low contrast ratio, so the script should warn if a style is below the W3 recommendation.
* The checker itself has no tests. It would be helpful to test it somehow -- though that might be not needed once there is an automated way to check styles.

a1714470:
Thanks for your advice, we will pick that up one month later.

Anteru:
I don't know what happened here. Did someone hack your account? There's a ton of unrelated files which makes this PR unacceptable.

a1714470:
> I don't know what happened here. Did someone hack your account? There's a ton of unrelated files which makes this PR unacceptable.

Actually, Leo777i and Hannah40 are my team members. Because they didn't want to do programming tasks, so they uploaded a lot of irrelevant documents to prove their work this semester. As our lecturer needs to review our jobs, so I cannot delete their files. I warned them not to upload these irrelevant files before. But they didn't accept my advice. My team has five members. However, I am the only one dedicated to solving this issue. Sorry, I stopped updating this task now. I will report to our lecturer first.

Sorry for all the inconvenience.

Anteru:
Ok. That's rather sad to hear. Thanks for clearing this up! Any chance you can open a new PR with just the `check.py` and I'll happily approve it; without all the other files/changes which are irrelevant to the problem at hand?

a1714470:
> Ok. That's rather sad to hear. Thanks for clearing this up! Any chance you can open a new PR with just the `check.py` and I'll happily approve it; without all the other files/changes which are irrelevant to the problem at hand?

Thank you for your advice, I need to seek advice from my lecturer first. Then maybe I just delete their files or open a new PR.

a1714470:
HI, after discussing with my lecturer, I will open a new PR. Thanks for your advice.

PR #1801, opened 2021-05-12
feat(lexers): add more man page filename globs

scop:
Numbered ones go up to 9, Perl ones may have 1p or 3pm suffixes.


Anteru:
Merged, thanks!

PR #1807, opened 2021-05-17
feat(lexers): add ASCII armored

scop:
Test data to follow later once questions raised have been sorted out.


pygments/lexers/asc.py
```diff
@@ -0,0 +1,49 @@
...
+
+__all__ = ['AscLexer']
+
+
+class AscLexer(RegexLexer):
+    """
+    Lexer for ASCII armored files, containing `-----BEGIN/END ...-----` wrapped base64 data.
+    """
+    name = 'ASCII armored'
+    aliases = ['asc', 'pem']
```
scop, line 23: Should we add some more here?

pygments/lexers/asc.py
```diff
@@ -0,0 +1,49 @@
...
+    filenames = [
+        '*.asc',  # PGP; *.gpg, *.pgp, and *.sig too, but those can be binary
+        '*.pem',  # X.509; *.cer, *.crt, *.csr, and key etc too, but those can be binary
+        'id_dsa', 'id_ecdsa', 'id_ecdsa_sk', 'id_ed25519', 'id_ed25519_sk', 'id_rsa',  # SSH private keys
+    ]
+    mimetypes = ['application/pgp-keys', 'application/pgp-encrypted', 'application/pgp-signature'],
```
scop, line 29: There are some MIME types around for X.509 files, but they don't differentiate ASCII armored and binary encodings.

pygments/lexers/asc.py
```diff
@@ -0,0 +1,49 @@
...
+
+    tokens = {
+        'root': [
+            (r'\s+', Text),
+            (r'^-----BEGIN [^\n]+-----$', Generic.Heading, 'data'),
+            (r'\S+', Comment),
+        ],
+        'data': [
+            (r'\s+', Text),
+            (r'^([^:]+)(:)([ \t]+)(.*)', bygroups(Name.Attribute, Operator, Text, String)),
```
scop, line 41: I wonder if we should use a different token type for the value here and the base64 data (`\S+` below), both are `String` now.

pygments/lexers/asc.py
Anteru: As long as they're not ambiguous it's fine.

pygments/lexers/asc.py
```diff
@@ -0,0 +1,49 @@
...
+
+from pygments.lexer import RegexLexer, bygroups
+from pygments.token import Comment, Generic, Name, Operator, Text, String
+
+__all__ = ['AscLexer']
+
+
+class AscLexer(RegexLexer):
+    """
+    Lexer for ASCII armored files, containing `-----BEGIN/END ...-----` wrapped base64 data.
```
Anteru, line 20: Please add `.. versionadded:: 2.10`

Anteru:
Thanks for starting to work on this! Please note that any new lexer should come with an example file and golden test output -- can you please add that?

scop:
Re test data: sure, see initial comment.

pygments/lexers/asc.py
scop: I was thinking about certificates and the like, but as/if we're not adding filenames and mimetypes for them, I think it's ok to leave them out from here, too.

pygments/lexers/asc.py
scop: Done.

pygments/lexers/asc.py
```diff
@@ -0,0 +1,51 @@
...
+        '*.pem',  # X.509; *.cer, *.crt, *.csr, and key etc too, but those can be binary
+        'id_dsa', 'id_ecdsa', 'id_ecdsa_sk', 'id_ed25519', 'id_ed25519_sk', 'id_rsa',  # SSH private keys
+    ]
+    mimetypes = ['application/pgp-keys', 'application/pgp-encrypted', 'application/pgp-signature']
+
+    flags = re.MULTILINE
+
+    tokens = {
+        'root': [
+            (r'\s+', Text),
```
Anteru, line 37: That should be whitespace, right?

pygments/lexers/asc.py
```diff
@@ -0,0 +1,51 @@
...
+            (r'^-----BEGIN [^\n]+-----$', Generic.Heading, 'data'),
+            (r'\S+', Comment),
+        ],
+        'data': [
+            (r'\s+', Text),
```
Anteru, line 42: Same here, whitespace.

pygments/lexers/asc.py
Anteru: No strong preference.

Anteru (changes requested):

pygments/lexers/asc.py
scop: I don't know, seems a lot of lexers mark whitespace as Text, and significantly fewer as Whitespace.

```shellsession
$ git grep -E "'\\\s[+*]', Text\)" | wc -l
272
$ git grep -E "'\\\s[+*]', .*Whitespace" | wc -l
33
```

But I'm fine either way, just let me know.

Anteru: Yeah, that's because we haven't paid enough attention to this previously :)

scop: Ok, changed to Whitespace.

Anteru (approved):

PR #1808, opened 2021-05-17
Add procfile lexer

sblondon:
This PR adds [Procfile](https://devcenter.heroku.com/articles/procfile#procfile-format) lexer with very basic unit tests.

### DIFF
--- a/pygments/lexers/procfile.py
+++ b/pygments/lexers/procfile.py
@@ -22,6 +22,8 @@ class ProcfileLexer(RegexLexer):
     Honcho tools.
     For more information about the definition of the format, see:
     https://devcenter.heroku.com/articles/procfile#procfile-format
+
+    .. versionadded:: 2.10
     """
     name = 'Procfile'
     aliases = ['procfile']
--- a/pygments/lexers/procfile.py
+++ b/pygments/lexers/procfile.py
@@ -34,7 +34,7 @@ class ProcfileLexer(RegexLexer):
             (r'^([a-z]+)(:)', bygroups(Name.Label, Punctuation)),
             (r'"[^"]*"', String),
             (r"'[^']*'", String),
-            (r'[0-9.]+', Number),
+            (r'[0-9]+', Number.Integer),
             (r'\$[a-zA-Z_][\w]*', Name.Variable),
             (r'(\w+)(=)(\w+)', bygroups(Name.Variable, Punctuation, String)),
             (r'([\w\s\-\./])', Text),
--- a/pygments/lexers/procfile.py
+++ b/pygments/lexers/procfile.py
@@ -32,11 +32,12 @@ class ProcfileLexer(RegexLexer):
     tokens = {
         'root': [
             (r'^([a-z]+)(:)', bygroups(Name.Label, Punctuation)),
+            (r'\s+', Text.Whitespace),
             (r'"[^"]*"', String),
             (r"'[^']*'", String),
             (r'[0-9]+', Number.Integer),
             (r'\$[a-zA-Z_][\w]*', Name.Variable),
             (r'(\w+)(=)(\w+)', bygroups(Name.Variable, Punctuation, String)),
-            (r'([\w\s\-\./])', Text),
+            (r'([\w\-\./]+)', Text),
         ],
     }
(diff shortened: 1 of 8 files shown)


Anteru:
Thanks for the contribution. Can you please add an example file with golden test output (see https://github.com/pygments/pygments/blob/master/Contributing.md for details)? This also seems to add a spurious `Procfile` which I think was for testing, can you please remove it?

sblondon:
@Anteru Thank you for the quick review!

`./Procfile` is removed.
`./tests/examplefiles/procfile/Procfile` and its .output file are updated.
The .output file is now added to the repository.

pygments/lexers/procfile.py
```diff
@@ -0,0 +1,40 @@
...
+
+    tokens = {
+        'root': [
+            (r'^([a-z]+)(:)', bygroups(Name.Label, Punctuation)),
+            (r'"[^"]*"', String),
+            (r"'[^']*'", String),
+            (r'[0-9.]+', Number),
+            (r'\$[a-zA-Z_][\w]*', Name.Variable),
+            (r'(\w+)(=)(\w+)', bygroups(Name.Variable, Punctuation, String)),
+            (r'([\w\s\-\./])', Text),
```
Anteru, line 38: I think you want a `+` in here. Right now this matches each individual letter as a single token as you can see in the output, and I doubt that's what you're after. Additionally, this matches whitespace -- I'd suggest a separate rule for whitespace (and using the right token type for whitespace.) In that case you need to match whitespace first because you have a dot in this rule (which I hope is intentional :) )

pygments/lexers/procfile.py
```diff
@@ -0,0 +1,40 @@
...
+from pygments.token import Name, Number, String, Text, Punctuation
+
+__all__ = ["ProcfileLexer"]
+
+
+class ProcfileLexer(RegexLexer):
+    """
+    Lexer for Procfile file format.
+
+    The format is used to run processus on Heroku or is used by Foreman or
```
Anteru, line 21: Is this a typo? "processus"?

pygments/lexers/procfile.py
```diff
@@ -0,0 +1,40 @@
...
+    Honcho tools.
+    For more information about the definition of the format, see:
+    https://devcenter.heroku.com/articles/procfile#procfile-format
+    """
```
Anteru, line 25: Please add a `.. versionadded:: 2.10` line to the comment.

pygments/lexers/procfile.py
```diff
@@ -0,0 +1,40 @@
...
+    name = 'Procfile'
+    aliases = ['procfile']
+    filenames = ['Procfile']
+
+    tokens = {
+        'root': [
+            (r'^([a-z]+)(:)', bygroups(Name.Label, Punctuation)),
+            (r'"[^"]*"', String),
+            (r"'[^']*'", String),
+            (r'[0-9.]+', Number),
```
Anteru, line 35: This will match `..9..` as a number -- is that intentional? Could you please add a number into your Procfile example?

Anteru (changes requested):

pygments/lexers/procfile.py
sblondon: I added it in the docstring.

pygments/lexers/procfile.py
sblondon: Sorry, it's a typo. It's the french word for `processes`. It's fixed.

Anteru: TIL :) Thanks!

pygments/lexers/procfile.py
sblondon: I added a basic number in Procfile file and removed the dot from the regular expression so it detects `Number.Integer`. I think it's good enough because I doubt someone use float number in commands. It is ok for you?

pygments/lexers/procfile.py
sblondon: There are 2 patterns: one for `Whitespace` and one for `Text`. These two patterns end with `+`.
Is it good for you?

pygments/lexers/procfile.py
Anteru: It's not great but I guess it'll do for now. I assume people will try to highlight valid Procfiles with it :)

Anteru:
Merged, thanks!

PR #1812, opened 2021-05-21
feat(lexers): add various SSH files

scop:


pygments/lexers/ssh.py
```diff
@@ -0,0 +1,200 @@
...
+
+class SshConfigLexer(RegexLexer):
+    """
+    Lexer for SSH client config files.
+
+    .. versionadded:: 2.10
+    """
+    name = 'SSH client config'
+    aliases = ['ssh_config']
+    filenames = ['ssh_config', 'config']
```
scop, line 140: `config` is pretty generic, don't know if we see that as a problem here.

Anteru: Nobody else is claiming config yet but I don't think this lexer should claim it. It feels too generic -- please remove it.

Anteru (changes requested):
Are you sure you don't want the quotes be part of the string? It's quite uncommon to highlight a string as `"`, 'the string', `"`, other lexers include the quotation marks. Is there a particular reason for SSH files to deviate from this behavior?

pygments/lexers/ssh.py
scop: Done.

scop:
> Are you sure you don't want the quotes be part of the string?

I have no personal preference. But I should note that I'm no longer using pygments, and have no immediate plans to work on this further, so I suggest someone else completes it (or closes without merging, but that would be a shame as this is fairly close I think). My branch should be accepting collaborator pushes.

Anteru:
Ok, thanks. Out of curiosity: Which solution are you using now?

scop:
After a short pygments try I was using https://github.com/alecthomas/chroma/ for a bit, but I've moved on to https://github.com/sharkdp/bat as it's only console syntax highlighting I'm after, and bat seems to serve me best I've found this far.

PR #1813, opened 2021-05-22
Improved Rust attributes

kaivol:
Rust attributes with nested brackets are currently not correctly highlighted. 

For example, in 
````rust
#[foo(bar = [baz, qux])]
````
the attribute is only highlighted until the **first** closing square bracket.

I think this should fixed with these changes.

### DIFF
--- a/pygments/lexers/rust.py
+++ b/pygments/lexers/rust.py
@@ -213,22 +213,10 @@ class RustLexer(RegexLexer):
         'attribute_common': [
             (r'"', String, 'string'),
             (r'\[', Comment.Preproc, 'attribute['),
-            (r'\(', Comment.Preproc, 'attribute('),
-            (r'\{', Comment.Preproc, 'attribute{'),
         ],
         'attribute[': [
             include('attribute_common'),
             (r'\];?', Comment.Preproc, '#pop'),
-            (r'[^"\]\[({]+', Comment.Preproc),
-        ],
-        'attribute(': [
-            include('attribute_common'),
-            (r'\);?', Comment.Preproc, '#pop'),
-            (r'[^")\[({]+', Comment.Preproc),
-        ],
-        'attribute{': [
-            include('attribute_common'),
-            (r'};?', Comment.Preproc, '#pop'),
-            (r'[^"}\[({]+', Comment.Preproc),
+            (r'[^"\]\[]+', Comment.Preproc),
         ],
     }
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by Anteru** [comment]
Can you please add a test for this? There's a rust example file, it should be sufficient to add one line with an attribute like that there.

**Comment by kaivol** [comment]
You mean like that?

Unfortunately I'm not familiar with pygments so I'm not exactly sure what I'm doing here...

**Comment by Anteru** [comment]
Yeah, adding it to the example file is sufficient, thanks!

**Comment by dario23** [comment]
i just stumbled over this problem as well. as far as i can see, the diff needed to make the rust tests pass is
```
diff --git a/tests/examplefiles/rust/eval.rs.output b/tests/examplefiles/rust/eval.rs.output
index 186a7c0c..dd0f7997 100644
--- a/tests/examplefiles/rust/eval.rs.output
+++ b/tests/examplefiles/rust/eval.rs.output
@@ -218,7 +218,10 @@
 '/// Represents a value (either 16-bit or 32-bit) at runtime.\n' Literal.String.Doc
 
 '#['          Comment.Preproc
-'derive(Clone, PartialEq, Eq, Debug)' Comment.Preproc
+'derive'      Comment.Preproc
+'('           Comment.Preproc
+'Clone, PartialEq, Eq, Debug' Comment.Preproc
+')'           Comment.Preproc
 ']'           Comment.Preproc
 '\n'          Text.Whitespace
 ```

aka the `#[derive(..)]` line is now looked at in more detail as well, just like the initial example.

birkenfeld:
Actually, why are we tracking `(` and `)` pairs? The only thing that's relevant is nested `[` and `]`...

kaivol:
For highlighting purposes it should indeed be sufficient to look at square brackets.
Should I change the code accordingly?

birkenfeld:
That would be nice, and it will also reduce the number of individual tokens.

pygments/lexers/rust.py
```diff
@@ -213,16 +213,10 @@ class RustLexer(RegexLexer):
         'attribute_common': [
             (r'"', String, 'string'),
             (r'\[', Comment.Preproc, 'attribute['),
-            (r'\(', Comment.Preproc, 'attribute('),
         ],
         'attribute[': [
             include('attribute_common'),
             (r'\];?', Comment.Preproc, '#pop'),
```
kaivol, line 219: While we are at it, I don't think Rust allows trailing semicolons after attributes so the `:?` is probably incorrect here. 
Please correct me if I'm mistaken.

birkenfeld: Yes, indeed.

birkenfeld:
Thanks!

Issue #1781, opened 2021-04-20
JavaScript/TypeScript (big) update


Mesteery:
The JavaScript (and TypeScript lexer) should be updated :

  * [ ] more errors (`Error`, `EvalError`, `InternalError`, `RangeError`, `ReferenceError`, `SyntaxError`, `TypeError`, `URIError`) as `Name.Exception` ?
  * [ ] the `global` object (`Name.Builtin`, Node.js)
  * [ ] more `Name.Builtin` (such as typed arrays, `Intl`, `DataView`, `Set`, `Map`, `WeakSet`, `WeakMap`, `Reflect`, `JSON`, `ArrayBuffer`, `Atomics`) ?
  * [ ] highlight `\\...`, `\n`, `\r`, `\v`, `\t`, `\b`, `\f`, `\uXXXX` (U+0000 - U+FFFF), `\xXX` (0x00 - 0xFF) as `String.Escape` (same as Python lexer)
  * [ ] `**`, `**=`, `>>>=`, `??`, `&&=`, `||=`, `??=` as `Operator`
  * [ ] `typeof`, `instanceof`, `in`, `void`, `delete` as `Operator.Word`

_Originally posted by @Mesteery in https://github.com/pygments/pygments/issues/1741#issuecomment-792643216_

Anteru:
What's the rationale for `new` being `Operator.Word`?

Mesteery:
According to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/new), `new` is an operator.

Mesteery:
I also have a question, should the `Keyword.Reserved` be the same as https://tc39.es/ecma262/#prod-ReservedWord?

Anteru:
Thanks for the quick reply -- `Operator` it is. On the second question, yes, that sounds right.

Mesteery:
(https://github.com/pygments/pygments/pull/1814)

HarelM:
I'm not sure this is the right place to bring it up, so sorry in advance...
I'm looking into replacing the documentation of my site to Mkdocs-material which uses pygments under the hood for syntax highlighting.
The following is how I see it in github, which highlight class names and method names for example:
```ts
// Adds a full screen control to the map
map.addControl(new maplibregl.FullscreenControl({container: document.querySelector('body')}));
```
screen shot in dark mode of github:
[...]

PR #1814, opened 2021-05-23
Update javascript lexer

Mesteery:
Partially resolv*es* https://github.com/pygments/pygments/issues/1781.

### DIFF
--- a/pygments/lexers/_mapping.py
+++ b/pygments/lexers/_mapping.py
@@ -468,7 +468,7 @@ LEXERS = {
     'TurtleLexer': ('pygments.lexers.rdf', 'Turtle', ('turtle',), ('*.ttl',), ('text/turtle', 'application/x-turtle')),
     'TwigHtmlLexer': ('pygments.lexers.templates', 'HTML+Twig', ('html+twig',), ('*.twig',), ('text/html+twig',)),
     'TwigLexer': ('pygments.lexers.templates', 'Twig', ('twig',), (), ('application/x-twig',)),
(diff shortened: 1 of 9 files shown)


Anteru:
Merged, thanks! What's the rationale for removing `.tsx` as a Typescript alias?

Mesteery:
> What's the rationale for removing .tsx as a Typescript alias?

Because this lexer doesn't really support TSX.

vlakoff:
Using a variable named `from` in plain JavaScript code, I noticed it was highlighted as a reserved word (though it isn't a reserved word actually).

I pinpointed the cause to the following code, introduced by this PR:

https://github.com/pygments/pygments/blob/513f8fc34f7ec915994ac226ff333b2f6864b54c/pygments/lexers/javascript.py#L88-L89

Mesteery:
`from` is an reserved word. By the way, when you use `'strict mode'`, your code will fail if you use `from` as a variable name.
`from` is used with `import` like `import something from 'package'`.

vlakoff:
My code is in strict mode, and it is currently using a variable named `from`, which doesn't fail.

I think that's because `import`, then subsequently `from`, is reserved only if the script if interpreted as a module. See the [`import` documentation on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import):

> In order to use the `import` declaration in a source file, the file must be interpreted by the runtime as a [module](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules). In HTML, this is done by adding `type="module"` to the [`<script>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script) tag. Modules are automatically interpreted in [strict mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode).

Though, I don't think Pygments can distinguish between the "regular" and "module" types…

Issue #1816, opened 2021-05-24
Duplicate token types and token documentation


doerwalter:
I've used the following script to get all token types that are defined in any of the lexers:

```python
from pygments import token, lexers

for (name, aliases, patterns, mimetypes) in lexers.get_all_lexers():
	for a in aliases:
		l = lexers.get_lexer_by_name(a)
		break
	else:
		for p in patterns:
			l = lexers.get_lexer_for_filename(p)
			break
		else:
			for m in mimetypes:
				l = lexers.get_lexer_for_mimetype(m)
				break

def list_tokens(t):
	yield t
	for st in t.subtypes:
		yield from list_tokens(st)

tokens = list(list_tokens(token.Token))

for t in sorted(tokens):
	print(t)
```

The output is the following:

```
Token
Token.Comment
Token.Comment.Directive
Token.Comment.Doc
Token.Comment.Hashbang
Token.Comment.Multi
Token.Comment.Multiline
Token.Comment.Preproc
Token.Comment.PreprocFile
Token.Comment.Single
Token.Comment.Singleline
Token.Comment.SingleLine
Token.Comment.Special
Token.Error
Token.Escape
Token.Generic
Token.Generic.Deleted
Token.Generic.Emph
Token.Generic.Error
Token.Generic.Heading
Token.Generic.Inserted
Token.Generic.Output
Token.Generic.Prompt
Token.Generic.Strong
Token.Generic.Subheading
Token.Generic.Traceback
Token.Generic.Whitespace
Token.Keyword
Token.Keyword.Builtin
Token.Keyword.Constant
Token.Keyword.Control
Token.Keyword.Declaration
Token.Keyword.Keyword
Token.Keyword.Namespace
Token.Keyword.PreProc
Token.Keyword.Pseudo
Token.Keyword.Removed
Token.Keyword.Reserved
Token.Keyword.Token
Token.Keyword.Tokens
Token.Keyword.Type
Token.Keyword.Word
Token.Literal
Token.Literal.Char
Token.Literal.Date
Token.Literal.Number
Token.Literal.Number.Attribute
Token.Literal.Number.Bin
Token.Literal.Number.Dec
Token.Literal.Number.Decimal
Token.Literal.Number.Float
Token.Literal.Number.Hex
Token.Literal.Number.Int
Token.Literal.Number.Integer
Token.Literal.Number.Integer.Long
Token.Literal.Number.Oct
Token.Literal.Number.Octal
Token.Literal.Number.Radix
[...]

PR #1819, opened 2021-05-26
Add scripts/count_token_references.py to check for "unicorn" tokens.

doerwalter:
This pull request "fixes" issue https://github.com/pygments/pygments/issues/1816

It implements a script that counts how often a token is referenced by the lexer sources to find any typos in token names.


Anteru:
Thanks a lot for preparing this!

PR #1822, opened 2021-05-27
Add a tooltips option to HtmlFormatter.

doerwalter:
This pull requests adds an option `tooltips` to `HtmlFormatter`. When `tooltips` is true, the generated HTML includes a `title` attribute for each token `<span>` tag, that includes the token name.

This helps with developing and debugging styles.


Anteru:
Thanks for the PR. This sounds like it's only useful for debugging, right? There's no reason to ever have that in production -- and I think that should be reflected in the name. Maybe `debug_token_types` instead of `tooltips`? It would be also good to mention this on the "how to write a style" page.

doerwalter:
Yes, this is mostly useful for debugging and developing lexers and styles.

It is less useful on sites that simply use Pygments for syntax highlight. But it might be somewhat useful on the Pygments site itself.

I've updated the pull request to rename the option from `tooltips` to `debug_token_types`.

I've added a paragraph to `lexerdevelopment.rst` to explain the new `debug_token_types` option.

Anteru:
Merged, thanks!

PR #1825, opened 2021-05-29
Add JavaScript Node.js Console Lexer

cltrudeau:
I needed something like pycon for Node in a project I was working on, once written it was easy enough to put in here.

I've tested under py36 - py39. I don't have py35 or pypy installed, hopefully everything will work there.

Drop me a note if you need something formatted differently / more tests / whatever.

Thanks for the great tool!

### DIFF
--- a/pygments/lexers/javascript.py
+++ b/pygments/lexers/javascript.py
@@ -10,16 +10,17 @@
 
 import re
 
-from pygments.lexer import RegexLexer, include, bygroups, default, using, \
-    this, words, combined
+from pygments.lexer import RegexLexer, Lexer, include, bygroups, default, \
+    using, this, words, combined, do_insertions
 from pygments.token import Text, Comment, Operator, Keyword, Name, String, \
-    Number, Punctuation, Other
+    Number, Punctuation, Other, Generic
 from pygments.util import get_bool_opt
 import pygments.unistring as uni
 
 __all__ = ['JavascriptLexer', 'KalLexer', 'LiveScriptLexer', 'DartLexer',
            'TypeScriptLexer', 'LassoLexer', 'ObjectiveJLexer',
-           'CoffeeScriptLexer', 'MaskLexer', 'EarlGreyLexer', 'JuttleLexer']
+           'CoffeeScriptLexer', 'MaskLexer', 'EarlGreyLexer', 'JuttleLexer',
+           'NodeConsoleLexer']
 
 JS_IDENT_START = ('(?:[$_' + uni.combine('Lu', 'Ll', 'Lt', 'Lm', 'Lo', 'Nl') +
                   ']|\\\\u[a-fA-F0-9]{4})')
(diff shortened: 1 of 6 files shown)


pygments/lexers/javascript.py
```diff
@@ -1534,3 +1536,63 @@ class JuttleLexer(RegexLexer):
...
+        undefined
+        > a
+        3
+        > let b = '4'
+        undefined
+        > b
+        '4'
+        > b == a
+        false
+    """
```
Anteru, line 1557: Please add a `.. versionadded: 2.10` line at the end.

pygments/lexers/javascript.py
```diff
@@ -1534,3 +1536,63 @@ class JuttleLexer(RegexLexer):
...
+        for match in line_re.finditer(text):
+            line = match.group()
+            if line.startswith('> '):
+                insertions.append((len(curcode), 
+                    [(0, Generic.Prompt, line[:2])]))
+
+                curcode += line[2:]
+            elif line.startswith('...'):
+                # node does a nested ... thing depending on depth
+                code = line.lstrip('.')
```
Anteru, line 1578: If you know it's `...`, wouldn't `line [4:]` be faster here, or is it possible that the prefix part is longer?

Anteru (changes requested):

pygments/lexers/javascript.py
Mesteery: > or is it possible that the prefix part is longer?

Yes
image

Mesteery:
I find 'JavaScript Node.js console session' too ambiguous.
Wouldn't talking about 'REPL' and removing 'JavaScript' be better?
Also :
```
        aliases = ['nodejs', 'nodecon'] <- nodejs alias is false, nodecon is too ambiguous
        mimetypes = ['application/javascript', 'application/x-javascript', ...] <- these mime types are false
```

Anteru:
> I find 'JavaScript Node.js console session' too ambiguous.

Can you please elaborate as to how this is ambiguous? What would you suggest? `nodejsrepl`?

Mesteery:
> Can you please elaborate as to how this is ambiguous? What would you suggest? nodejsrepl?

Yes, something like nodejsrepl.

It's not really ambiguous, but it would be better to have a reference to 'REPL', without 'JavaScript'.

For example: 'Node.js REPL session'.

cltrudeau:
> I find 'JavaScript Node.js console session' too ambiguous.
> Wouldn't talking about 'REPL' and removing 'JavaScript' be better?
> Also :
> 
> ```
>         aliases = ['nodejs', 'nodecon'] <- nodejs alias is false, nodecon is too ambiguous
>         mimetypes = ['application/javascript', 'application/x-javascript', ...] <- these mime types are false
> ```

Alias has been fixed, mimetypes have been reduced to text/x-nodejsrepl.

I tried renaming the Lexer to NodeREPLLexer for consistency with this suggestion, but it breaks the tests cases -- the have a special condition in them for things named Console that seems to be needed here. I didn't dig any further, just put it back to NodeConsoleLexer.

Anteru (approved):

Anteru:

PR #1827, opened 2021-05-31
Add golang alias for the Go language

Situphen:
In our main project at @zestedesavoir we write our publications in (a slightly modified) Markdown and use hightlight.js to render it as HTML and Pygments to render it as PDF. We found that highlight.js supports the `golang` alias for the Go language but Pygments doesn't and I think that it would be nice to support it so here is the pull request!


birkenfeld:
LGTM, however please run `make mapfiles` and commit the resulting change as well.

Situphen:
Done!

birkenfeld:
Thanks!

Situphen:
No, thank you! :)

PR #1830, opened 2021-06-02
Fix CFamilyLexer preprocessor tokenization errors

### LINKED ISSUE #1820: CFamilyLexer fails to tokenize spaces before preprocessor macro
The following string code fails to tokenize even though it is valid C++ code:
```cpp
; 
 #define A 0
```
Note that there are spaces around the line break. I would expect the `#define A 0` to be tokenized as macro, but `#` produces an error token. The tokens produced are
- ` print(list(CFamilyLexer().get_tokens("; \n #define A 0")))`:
  ```
  [(Token.Punctuation, ';'), (Token.Text, ' \n '), (Token.Error, '#'), (Token.Name, 'define'), (Token.Text, ' '), (Token.Name, 'A'), (Token.Text, ' '), (Token.Literal.Number.Integer, '0'), (Token.Text, '\n')]
  ```

## What I have found
I have done some debugging and found that the spaces around the line break are important. Removing either of the spaces produces correct tokenization.
- `print(list(CFamilyLexer().get_tokens("; \n#define A 0")))`:
  ```
  [(Token.Punctuation, ';'), (Token.Text, ' \n'), (Token.Comment.Preproc, '#'), (Token.Comment.Preproc, 'define A 0'), (Token.Comment.Preproc, '\n')]
  ```
- `print(list(CFamilyLexer().get_tokens(";\n #define A 0")))`
  ```
  [(Token.Punctuation, ';'), (Token.Text, '\n'), (Token.Text, ' '), (Token.Comment.Preproc, '#'), (Token.Comment.Preproc, 'define A 0'), (Token.Comment.Preproc, '\n')]
  ```

As far as I understand, the relevant part of the tokenizer are the following definitions:
```python
'whitespace': [
    # preprocessor directives: without whitespace
    ('^#', Comment.Preproc, 'macro'),
    # or with whitespace
    ('^(' + _ws1 + ')(#)',
     bygroups(using(this), Comment.Preproc), 'macro'),
    (r'\n', Text),
    (r'\s+', Text),
],
```
Here in the failing case the `";"` first matches as a punctuation. Then `'\s+'` matches `" \n "`. Now we would want either `'^#',` or `'^(' + _ws1 + ')(#)'` to match the `"#"`, but this isn't the case because `^` matches only the start of the line, and here we have already matched the space at the start of the line. Hence the tokenizer produces an error.

In the case that we remove the space after line break, `'\s+'` matches `" \n"`. As there is no space at the start of the line, ` '^#'` matches `"#"` and the lexer begins to tokenize a macro.

In the case that we remove the space before the line break, `'\n'` comes before `\s+` in the list of regexes and consumes `"\n"`. At this point we still haven't matched the space, but it gets matched by `'^(' + _ws1 + ')(#)'` as it is at the start of the line. Hence the lexer again begins to tokenize a macro.

## My analysis

I fail to see why `^` at the start of `Comment.Preproc` regexes would be necessary. Maybe that could be removed? I also think that `\s+` should not match line breaks, but I'm not sure how easy that would be as `\s` is local-aware.

Ps. The above example is a minimal example I have found. I have encountered this bug in the wild, for example the following code also fails to tokenize:
```cpp
int main() {
  prepare(); 
  #pragma omp parallel for
  for (int i = 0; i < 10; i++) {}
}
```

henkkuli:
CFamilyLexer fails to tokenize preprocessor macros when they are preceded by a line break surrounded by spaces. This is the case because prerpocessor regex rule expects to start at the beginning of the line, but the space regex rule matches also the whitespace after the line break. Now the space rule has been refined not to match the line break. Because of this, the preprocessor regex rule correctly matches prerpocessor tokens even when they are preceded by spaces, at the cost of adding some more tokens in the token stream in some cases.

The main change is in `pygments/lexers/c_cpp.py`. The generic whitespace rule `\s+` has been changed to `[^\S\n]` to avoid matching line breaks. As a consequence of this, many files under `tests/examplefiles` changed. All of the changes seem to be of the form
```
'      \n      ' Text
```
changed to
```
'      '      Text
'\n'          Text

'      '      Text
```

In addition to these, the PR adds three new tests under `tests/snippets` which test the behavior of the preprocessor tokenizer in different situations. The test `tests/snippets/c/test_preproc_file5.txt` can be controversial as it tests the behavior in situation where the code is invalid and hence the output contains an error token. I'll let the maintainers decide whether that should be included or removed.

Fixes #1820.

### DIFF
--- a/pygments/lexers/c_cpp.py
+++ b/pygments/lexers/c_cpp.py
@@ -56,7 +56,7 @@ class CFamilyLexer(RegexLexer):
             ('^(' + _ws1 + ')(#)',
              bygroups(using(this), Comment.Preproc), 'macro'),
             (r'\n', Text),
-            (r'\s+', Text),
+            (r'[^\S\n]+', Text),
             (r'\\\n', Text),  # line continuation
             (r'//(\n|[\w\W]*?[^\\]\n)', Comment.Single),
             (r'/(\\\n)?[*][\w\W]*?[*](\\\n)?/', Comment.Multiline),
--- a/tests/examplefiles/arduino/Blink.ino.output
+++ b/tests/examplefiles/arduino/Blink.ino.output
@@ -1,7 +1,8 @@
 '/*\n  Blink\n  Turns on an LED on for one second, then off for one second, repeatedly.\n \n  This example code is in the public domain.\n */' Comment.Multiline
 '\n'          Text
 
-' \n'         Text
+' '           Text
+'\n'          Text
 
 '// Pin 13 has an LED connected on most Arduino boards.\n' Comment.Single
 
@@ -42,7 +43,8 @@
 'OUTPUT'      Keyword.Reserved
 ')'           Punctuation
 ';'           Punctuation
-'     \n'     Text
+'     '       Text
+'\n'          Text
 
 '}'           Punctuation
 '\n'          Text
--- a/tests/examplefiles/c/example.c.output
+++ b/tests/examplefiles/c/example.c.output
@@ -313,7 +313,10 @@
 '"'           Literal.String
 ')'           Punctuation
 ';'           Punctuation
-' \n\t'       Text
+' '           Text
+'\n'          Text
+
+'\t'          Text
 'sprintf'     Name
 '('           Punctuation
 'tempString'  Name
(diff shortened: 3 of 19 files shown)


Anteru:
Merged, thanks!

PR #1834, opened 2021-06-05
tcl lexer: no error for @ or ,

### LINKED ISSUE #1742: tcl lexer bare words
In Tcl, everything is a string by default. For example, you can write `puts @12,34` and it will be interpreted same as `puts "@12,34"`. This is useful for Tk text widget, which uses this to denote the character at x coordinate `12` and y coordinate `34`.

Pygments highlights `@` and `,` as syntax error.

Akuli:
Fixes #1742


Anteru:
Merged, thanks a lot for the contribution!

PR #1842, opened 2021-06-18
Swift lexer: added concurrency keywords

nemecek-filip:
Swift got new concurrency features in version 5.5, that contain the async, await and actor keywords


Anteru:
Merged, thanks!

PR #1848, opened 2021-06-26
CFamilyLexer: support unicode identifiers

### LINKED ISSUE #998: C++ Unicode identifiers
(Original issue 1294 created by mickk on 2016-10-24T01:40:04.372130+00:00)

C++11 has some support for the use of Unicode in identifiers:

```c++
int main()
{
    double \u03B1 = 1.0;
    α += 1;
    return \u03B1;
}
```

It appears Pygments [doesn’t like that](http://pygments.org/demo/6310561/).

Breakdown by lines:

 - lines 3 and 5: individual Unicode code points can be specified in the source as `\uNNNN` or `\UNNNNNNNN`, much like they can be inside string literals -- the spec calls them 'universal character names'
 - line 4: compilers map characters that fall outside of the usual a-zA-Z0-9 etc. to their UCNs, hence a literal `α` in the source is effectively the same as having `\u03B1` which is how we're referring to the variable we introduced on line 3

There's also a bunch of rules to prohibit ill-formed Unicode, the use of non-letter-like symbols in identifiers, and digit-like symbols at the start of identifiers. [cppreference](http://en.cppreference.com/w/cpp/language/identifiers) has some resources on the topic, although it's not authoritative and I haven't double-checked it myself.

To the best of my knowledge Clang has support for this ([Coliru demo](http://coliru.stacked-crooked.com/a/f554b1e4968a99fc)), GCC has partial support (only UCNs, no mapping).

amitkummer:
Fixes #998.

### DIFF
--- a/pygments/lexers/c_cpp.py
+++ b/pygments/lexers/c_cpp.py
@@ -42,8 +42,8 @@ class CFamilyLexer(RegexLexer):
     _intsuffix = r'(([uU][lL]{0,2})|[lL]{1,2}[uU]?)?'
 
     # Identifier regex with C and C++ Universal Character Name (UCN) support.
-    _ident = r'(?:[a-zA-Z_$]|\\u[0-9a-fA-F]{4}|\\U[0-9a-fA-F]{8})(?:[\w$]|\\u[0-9a-fA-F]{4}|\\U[0-9a-fA-F]{8})*'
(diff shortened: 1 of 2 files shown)


amitkummer:
I also wanted to add this change to the CHANGES file, but wasn't sure if I should touch it. For future PRs, is it ok to do so?

Anteru:
Sorry, missed that completely. That's removing `_` from the expression, is `_` contained in `\w`? Otherwise this seems fine to be, just surprised it's `[\w$]` and not `[\w_$]`.

PR #1849, opened 2021-06-29
use ini lexer for systemd service files

Akuli:
A `foo.service` specific lexer could highlight valid and invalid keys differently, but this is much better than no syntax highlighting.


birkenfeld:
Makes sense. Do you also want to add the other systemd related filenames (`.socket`, `.target`, `.slice`, and so on)?

As you can see from CI, you need to run `make mapfiles` afterwards to regenerate the mapping of patterns to lexers.

Anteru:
It seems unlikely that someone will try to syntax highlight a `.swap` swap file though, so I don't have concerns. And there's no programming language using `.path` as the file ending, is there?

Akuli:
Good points. I don't think this will be a problem in practice.

Anteru:
Merged, thanks!

PR #1854, opened 2021-07-06
Fix FutureWarning for words() with "["

### LINKED ISSUE #1853: Warning upon words() on list containing "["
The following code:

```python
from pygments.lexer import RegexLexer, words
from pygments.token import Text

class MyLexer(RegexLexer):
    tokens = {'root': [(words(["[", "a"]), Text)]}

list(MyLexer().get_tokens("a"))
```

causes a `FutureWarning: Possible nested set at position 2`. This warning was introduced in Python 3.7 (https://docs.python.org/dev/whatsnew/3.7.html#re).

Jean-Abou-Samra:
regex_opt() groups characters in sets when possible. The
warning was caused when the "[" character ended up at the beginning
of a set: r"[[...]". This emits a FutureWarning since Python 3.7
due to possible changes in semantics in the future
(https://bugs.python.org/issue30349). Just add "[" to the list
of characters that should be escaped in sets. Add unit tests
for words().

[Closes #1853.]

### DIFF
--- a/pygments/regexopt.py
+++ b/pygments/regexopt.py
@@ -15,7 +15,7 @@ from os.path import commonprefix
 from itertools import groupby
 from operator import itemgetter
 
-CS_ESCAPE = re.compile(r'[\^\\\-\]]')
+CS_ESCAPE = re.compile(r'[\[\^\\\-\]]')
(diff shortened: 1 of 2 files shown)


birkenfeld:
Very nice!

Issue #1856, opened 2021-07-09
Scala lexer: `\` is considered an error token


lukapeschke:
A standalone `\` character is considered to be an error token by the scala lexer. However, it is a commonly-used json4s operator (https://github.com/json4s/json4s#extracting-values): It can for example be used as follows:

```scala
val userNameJValue = json \ "user_name"
```

I've been able to fix this issue by adding `'\\'` to `ScalaLexers`'s operators. However, since it is not a standard operator of the language, I'm not quite sure this is the right approach. If you're fine with this approach, let me know and I'll make a PR. 

Thanks for your work on pygments!

PR #1857, opened 2021-07-09
Scala lexer: Add support for the '\' operator

lukapeschke:
This operator is defined by json4s, which is one of scala's most popular JSON libraries.

This would fix https://github.com/pygments/pygments/issues/1856 .

Feel free to discard this if you don't like this solution


birkenfeld:
LGTM!

PR #1861, opened 2021-07-14
Add DO CONCURRENT statement to the list of keywords in the Fortran le…

ecasglez:
…xer.


Anteru:
Does `DO CONCURRENT` require exactly one space between the words, or can it be more?

ecasglez:
Good point. Yes, it can have any number of spaces.

ecasglez:
It is wierd. I just noticed that the keyword "concurrent" is already present in the list so it should be automatically highlighted. I will look deeper into it.

PR #1862, opened 2021-07-14
Update Fennel lexer to include forms from latest version.

technomancy:
Also separate out declarations from other special forms and macros.


Anteru:

PR #1864, opened 2021-07-17
Updates to Markdown lexer

brezniczky:
Some improvements to how bold/italic markdown formatting is tokenized e.g. in multiline situations.
Not that these things became 100% accurate, but a few useful improvements for my current purposes.

Opening a pull request before these get forgotten forever :)
May or may not be "ready" - no harm is done if the PR gets rejected, I appreciate comments if that would be the case.
Might be good to clear things up early, more commits may be on the way soon.

I see there are months old PRs so I'm not expecting any immediate reaction, but anyway, thanks in advance!

### DIFF
--- a/pygments/lexers/markup.py
+++ b/pygments/lexers/markup.py
@@ -555,7 +555,7 @@ class MarkdownLexer(RegexLexer):
             (r'^(\s*)([0-9]+\.)( .+\n)',
             bygroups(Text, Keyword, using(this, state='inline'))),
             # quote
-            (r'^(\s*>\s)(.+\n)', bygroups(Keyword, Generic.Emph)),
+            (r'^(\s*[> ]+\s)(.+\n)', bygroups(Keyword, Generic.Emph)),
             # code block fenced by 3 backticks
             (r'^(\s*```\n[\w\W]*?^\s*```$\n)', String.Backtick),
             # code block with language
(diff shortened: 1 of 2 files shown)


pygments/lexers/markup.py
```diff
@@ -555,7 +555,7 @@ def _handle_codeblock(self, match):
             (r'^(\s*)([0-9]+\.)( .+\n)',
             bygroups(Text, Keyword, using(this, state='inline'))),
             # quote
-            (r'^(\s*>\s)(.+\n)', bygroups(Keyword, Generic.Emph)),
+            (r'^(\s*[> ]+\s)(.+\n)', bygroups(Keyword, using(this, state='inline'))),
```
Jean-Abou-Samra, line 558: Are you sure about his? `[> ]+` matches indentation, and I recall that in some Markdown dialects an indented block is a code block.

pygments/lexers/markup.py
```diff
@@ -571,13 +571,17 @@ def _handle_codeblock(self, match):
             # warning: the following rules eat outer tags.
             # eg. **foo _bar_ baz** => foo and baz are not recognized as bold
             # bold fenced by '**'
-            (r'([^\*]?)(\*\*[^* \n][^*\n]*\*\*)', bygroups(Text, Generic.Strong)),
+            (r'(?<!\*)(\*\*[^* \n]([^*\n]+\n?)*(?<![ \n])\*\*)(?!\*)',
```
Jean-Abou-Samra, line 574: `bygroups` is no longer necessary.

More importantly, this regex (as well as the others below) has catastrophic backtracking. Try

```python
>>> import re
>>> re.match(r'(?<!\*)(\*\*[^* \n]([^*\n]+\n?)*(?<![ \n])\*\*)(?!\*)', '**' + 'a'*100) # hangs
```

See `Contributing.md` for explanations and remedies.

Jean-Abou-Samra (commented):
Here's some preliminary review.

PR #1866, opened 2021-07-21
Added lexer for parsing GSQL files

DanBarkus:
Created lexer for TigerGraph GSQL query language and .gsql files.


Anteru:
Looks like the test files are not matching? Could you please regenerate the test output files?

Anteru:
Thanks. Is this correct that you want things like `QUERY` identified as `Builtin.Name` and not `Keyword`?

PR #1867, opened 2021-07-25
1526/fusumwan/unreadablestyle patch

a1714470:
Creating a way to check if a style defines all colors and if those have good contrast. The "check.py" is used to fixed issue

[#1526](https://github.com/pygments/pygments/issues/1526)

In general, our check.py includes a color similarity function. Such a function is used to list out all the color similarities from one style. Besides, the check function also includes the contrast rato, relative luminance, contrast AA, and contrast AAA. We hope it can check all styles.


Anteru:
Thanks, that looks much better. Can you please add a comment at the top with the license note and a quick blurb about what this script is doing?

birkenfeld:
Some comments:

* Looks like this PR still includes the unnecessary commits. It doesn't matter so much for us, since we will squash the PR when merging, but for projects that don't do that a clean commit history should be preferred.

* This adds a new module in the Pygments library, but without using it somewhere.  Since the goal here should be to check the styles before merge/release, it should rather be a script in `scripts/` together with a target in the Makefile to run the check over all existing styles.  We can then add this to the CI runs.

* The code seems to need some fixes from a cursory glance.  For example, some methods reference an object `p1` which does not exist.  Also, a few methods go through `get_style_arr()`, filtering by `self.style`, each time they are called, instead of once retrieving the desired style and then just looking up its attributes.

a1714470:
Thanks, I added some comments to explain how this script works.

a1714470:
> Some comments:
> 
> * Looks like this PR still includes the unnecessary commits. It doesn't matter so much for us, since we will squash the PR when merging, but for projects that don't do that a clean commit history should be preferred.
> * This adds a new module in the Pygments library, but without using it somewhere.  Since the goal here should be to check the styles before merge/release, it should rather be a script in `scripts/` together with a target in the Makefile to run the check over all existing styles.  We can then add this to the CI runs.
> * The code seems to need some fixes from a cursory glance.  For example, some methods reference an object `p1` which does not exist.  Also, a few methods go through `get_style_arr()`, filtering by `self.style`, each time they are called, instead of once retrieving the desired style and then just looking up its attributes.


I apologize for my mistake.  I uploaded my test version by mistake.   Object "p1" should be "self". 

Thanks for your comments. I agree with you, it is good to add it from the Makefile to the bash script to check all existing styles.

Anteru:
This is handled via #1919, #1937, #1938, #1940 now so I'm closing this because we'd merge the same work twice otherwise. Sorry for the trouble!

PR #1872, opened 2021-07-29
Only allow tables at line start to fix arrays.

eerimoq:
Hi!

Arrays are matched as keywords on master. This PR aims to fix that.

On master:
image

With this PR:
image

### DIFF
--- a/pygments/lexers/configs.py
+++ b/pygments/lexers/configs.py
@@ -968,9 +968,11 @@ class TOMLLexer(RegexLexer):
 
     tokens = {
         'root': [
+            # Table
+            (r'^(\s*)(\[.*?\])$', bygroups(Text, Keyword)),
 
             # Basics, comments, strings
-            (r'\s+', Text),
+            (r'[ \t]+', Text),
             (r'#.*?$', Comment.Single),
             # Basic string
             (r'"(\\\\|\\[^\\]|[^"\\])*"', String),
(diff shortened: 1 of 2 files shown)


pygments/lexers/configs.py
```diff
@@ -968,9 +968,11 @@ class TOMLLexer(RegexLexer):
 
     tokens = {
         'root': [
+            # Table
+            (r'^(\s*)(\[.*?\])$', bygroups(Text, Keyword)),
 
             # Basics, comments, strings
-            (r'\s+', Text),
+            (r'[ \t]+', Text),
```
birkenfeld, line 975: is this change necessary? You're not matching newlines between entries anywhere now; they are getting caught by the "out of rules" handler in lexer.py.

eerimoq: Yes, otherwise tables with preceding spaces are not found. I guess `\s+` consumes the newline character.

image

image

birkenfeld: Right, but then please include an explicit rule for `'\n'` too.

eerimoq: Done.

eerimoq: Anything else needed?

birkenfeld:
LGTM, thanks!

eerimoq:
When do you think the next release will be created?

birkenfeld:
@Anteru will know :)

Anteru:
I'm hoping for sometime during August. The milestone is tentatively scheduled for the first weekend in August, I hope to get around to cleaning up the release notes etc. and push for a release then.

PR #1873, opened 2021-07-29
Added groff formatter

marcinzelent:
This is a formatter that can output the source code in a form that can be used with [GNU Troff (groff)](https://www.gnu.org/software/groff/).

Groff is a typesetting system that has its roots in some of the oldest text formatting software out there. It can be used to create formatted documents, papers, books, etc. It is similar to LaTeX in that sense but much more minimal and older. Nowadays, it's mostly used for creation of `man` manual pages.

This formatter is using groff escapes to change the color and font style (regular, bold, italic) of the text. It also has a few options that allow the user to print the source code with line numbers and wrap it to a specified width. By default, the output is using a monospace font but it can be disabled too. Some special characters that are used by groff are replaced by escapes for printing Unicode characters. The same applies for non-ASCII characters, which groff can't print. The formatter isn't using any macro specific commands, so it should work with all of them (ms, mom, man, etc.).

I tested the formatter thoroughly using various source code files of my own and the ones available in the tests/examplefiles directory. I also used different combinations of options to make sure they don't conflict with each other. I didn't notice any issues. The only time when the output might look different than expected is when input contains characters that groff doesn't support (possibly to the lack of them in the default font). It will then give a warning that it cannot find the character and won't print it.

To test it yourself, you can use the following command:
```
pygmentize -f groff <input> | groff -Tpdf > output.pdf
```

It is my first contribution to the project and I'm not programming in Python that often, so I'm open to comments and feedback.


Anteru:
That works quite nicely. Thanks a lot for your contribution!

PR #1875, opened 2021-07-30
Add numerical identifiers to LLVM label syntax

### LINKED ISSUE #1874: LLVM syntax for block labels excludes numbered blocks
In LLVM IR, it's valid syntax to write:

```llvm
5:
  %v = add i32 ...
  %x = ...
```

as long as `%5` is the correct implictly generated name for that block. Pygments excludes this - the LLVM lexer seems to only allow identifiers as block names. 

https://github.com/pygments/pygments/blob/master/pygments/lexers/asm.py#L362

I've put in a PR (#1875) that fixes the problem

Baltoli:
Fixes #1874 

Implicitly-named labels of the form `\d+:` are valid as well as the
explicitly-identified ones already supported using the identifier
syntax.

This PR extends the LLVM lexer to include the numbered syntax.

### DIFF
--- a/pygments/lexers/asm.py
+++ b/pygments/lexers/asm.py
@@ -353,7 +353,7 @@ class LlvmLexer(RegexLexer):
     #: optional Comment or Whitespace
     string = r'"[^"]*?"'
     identifier = r'([-a-zA-Z$._][\w\-$.]*|' + string + ')'
-    block_label = identifier + r'|(\d+)'
+    block_label = r'(' + identifier + r'|(\d+))'
 
     tokens = {
         'root': [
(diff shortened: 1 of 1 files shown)


pygments/lexers/asm.py
```diff
@@ -353,13 +353,14 @@ class LlvmLexer(RegexLexer):
...
     identifier = r'([-a-zA-Z$._][\w\-$.]*|' + string + ')'
+    block_label = identifier + r'|(\d+)'
 
     tokens = {
         'root': [
             include('whitespace'),
 
             # Before keywords, because keywords are valid label names :(...
-            (identifier + r'\s*:', Name.Label),
+            (block_label + r'\s*:', Name.Label),
```
birkenfeld, line 363: This will match the `\s*:` part only for the numeric case - you need to put the whole `identifier|number` alteration in a group.

Baltoli: Fixed

birkenfeld:
LGTM, thanks!

PR #1876, opened 2021-07-30
Add `pwsh` alias for “PowerShell” lexers

ExE-Boss:
This matches the official executable name [of **PowerShell 6** and newer](https://github.com/PowerShell/PowerShell):
- <https://github.com/PowerShell/PowerShell/issues/4214>

---

Also, the `posh` alias should be deprecated as it clashes with the “[Policy‑compliant Ordinary SHell](https://www.commandlinux.com/man-page/man1/posh.1.html)”.

### DIFF
--- a/pygments/lexers/shell.py
+++ b/pygments/lexers/shell.py
@@ -655,7 +655,7 @@ class PowerShellLexer(RegexLexer):
     .. versionadded:: 1.5
     """
     name = 'PowerShell'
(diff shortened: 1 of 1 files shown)


birkenfeld:
Please run `make mapfiles` and include the change in the PR.

birkenfeld:
As for `posh`, since that isn't something we'll have a lexer for, let's keep the alias.

Anteru:
Merged, thanks a lot!

Issue #1865, opened 2021-07-21
Fortran keywords composed by two words not highlighted


ecasglez:
There are two statments that are not properly highlighted by the Fortran Lexer:

* `goto` is only highlighted if there is no space between go and to. However, `go to` is also accepted by the Fortran standard, and it should be highlighted by pygments. "Go" and "To" can be used as name for variables, so they should only be highlighted when "go" preceedes "to".

* `do concurrent`. First of all, the word `concurrent` is misspelled in fortran.py where it is written with 3 R (`CONCURRRENT`). Fixing that typo it highlights well the `do concurrent` statement. However, "concurrent" itself can be used as a variable name which shouldn't be highlighed. I mean CONCURRENT should only by highlited when it follows a DO, like `DO CONCURRENT`.

I have tried to fix both trying regexp in file fortran.py but I did not succeed. Following a comment by Anteru in pull request #1861 I tried to make it work with a variable number of spaces between the two works. For example I tried adding `'go[ ]*to'` in fortran.py but it didn't work.

amitkummer:
Thank you for the detailed description!

If you still have the time/energy/will for fixing this, I think something like this rule should do the trick for a `go to` with a space:

```python
(r'(go)(\s*)(to)', bygroups(Keyword, Whitespace, Keyword))
```

blu-base:
@Anteru i guess, this issue is closable =)

Anteru:
Yes. good catch! I do go over all PRs before a release and look for related issues that haven't been referenced so I would have hopefully found that before the next release :)

PR #1877, opened 2021-08-02
Fix "do concurrent" and "go to" keywords in the Fortran lexer.

ecasglez:
* "Go to" statement was only highlighted if there was no space between "go" and "to".
* "Concurrent" keyword in the "Do Concurrent" statement was never highlighted because of a typo. It has been fixed. In addition, it now highlights them only if "Concurrent" is right after the "Do" keyword.
* I had to put the "do concurrent" changes before the already available list of keywords. Otherwise it won't highlight "Concurrent" because it finds first the "Do" keyword in the other list and stops searching for more keywords.
* This PR fixes issue #1865 

I used the following program to test the changes:

```
program myProgram
  implicit none
  integer :: x
  character :: concurrent
  integer :: go, to
  concurrent = '1'
  do concurrent (x = 1: 2)
     write(*,*) x, concurrent
     go = 100
     to = 200
  end do
  Do ConCuRRENT (x = 1: 2)
     write(*,*) x, concurrent
     go = 100
     to = 200
  end do
  goto 100
  go to 100
  GO   to 100
  do  x = 1, 2
     write(*,*) x
  end do
100   write(*,*) go, to
end program myProgram
```

### DIFF
--- a/pygments/lexers/fortran.py
+++ b/pygments/lexers/fortran.py
@@ -12,7 +12,7 @@ import re
 
 from pygments.lexer import RegexLexer, bygroups, include, words, using, default
 from pygments.token import Text, Comment, Operator, Keyword, Name, String, \
-    Number, Punctuation, Generic
+    Number, Punctuation, Generic, Whitespace
 
 __all__ = ['FortranLexer', 'FortranFixedLexer']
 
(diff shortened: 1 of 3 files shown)


pygments/lexers/fortran.py
```diff
@@ -48,19 +48,23 @@ class FortranLexer(RegexLexer):
         ],
         'core': [
             # Statements
+
+            (r'\b(DO)(\s+)(CONCURRENT)\b', bygroups(Keyword, Whitespace, Keyword)),
+            (r'\b(GO)(\s*)(TO)\b', bygroups(Keyword, Whitespace, Keyword)),
```
birkenfeld, line 53: Other whitespace is not given `Whitespace`, so please stay with `Text`.

Anteru: You mean in this particular lexer? I've changed a few to use Whitespace for all whitespace. I presume if I normalize this lexer to use Text.Whitespace you have no objections?

birkenfeld: Yep!

ecasglez: Hi. Do you mean changing:

```
(r'\b(DO)(\s+)(CONCURRENT)\b', bygroups(Keyword, Whitespace, Keyword)),
(r'\b(GO)(\s*)(TO)\b', bygroups(Keyword, Whitespace, Keyword)),
```
to:

```
(r'\b(DO)(\s+)(CONCURRENT)\b', bygroups(Keyword, Text.Whitespace, Keyword)),
(r'\b(GO)(\s*)(TO)\b', bygroups(Keyword, Text.Whitespace, Keyword)),
```

?

ecasglez: @Anteru @birkenfeld I added that Text.Whitespace, but I am not sure if this is the correct way to go. If it is not, feel free to close this pull request.

PR #1885, opened 2021-08-16
Lexer for new language Maxima

robert-dodier:
This PR implements a Pygments lexer for Maxima, a computer algebra system. See: https://maxima.sourceforge.io

The Maxima lexer is derived from the lexer for MuPAD, another computer algebra system.

The PR contains the lexer implementation and an example input, foo.mac, and its expected output from pytest --update-goldens.


pygments/lexers/maxima.py
```diff
@@ -0,0 +1,70 @@
+# -*- coding: utf-8 -*-
+"""
+    pygments.lexers.maxima
+    ~~~~~~~~~~~~~~~~~~~~~~~
+
+    Lexer for the computer algebra system Maxima.
+
+    Derived from pygments/lexers/algebra.py.
+"""
```
Anteru, line 9: This is missing the license blurb. Please remove the first line indicating the coding style as well.

pygments/lexers/maxima.py
```diff
@@ -0,0 +1,70 @@
...
+    tokens = {
+        'root': [
+            (r'/\*', Comment.Multiline, 'comment'),
+            (r'"(?:[^"\\]|\\.)*"', String),
+            (r'\(|\)|\[|\]|\{|\}', Punctuation),
+            (r'''(?x)\b(?:
+                if|then|else|elseif|
+                do|while|
+                repeat|until|
+                for|from|to|downto|step|thru
```
Anteru, line 39: Please use the `words` function for this (and similarly below)

pygments/lexers/maxima.py
```diff
@@ -0,0 +1,70 @@
...
+              ((?:[a-zA-Z_#][\w#]*|`[^`]*`)
+              (?:::[a-zA-Z_#][\w#]*|`[^`]*`)*)(\s*)([(])''',
+             bygroups(Name.Function, Text, Punctuation)),
+            (r'''(?x)
+              (?:[a-zA-Z_#%][\w#%]*|`[^`]*`)
+              (?:::[a-zA-Z_#%][\w#%]*|`[^`]*`)*''', Name.Variable),
+            (r'[-+]?\d+\.?' + terminated, Number.Integer),
+            (r'[-+]?(\d*\.\d+([bdefls][-+]?\d+)?|\d+(\.\d*)?[bdefls][-+]?\d+)'
+                + terminated, Number.Float),
+            (r'.', Text)
```
Anteru, line 62: Is there any valid text left or just whitespace? We're trying to be a bit more strict about `Text.Whitespace` to capture whitespace instead of having everything as `Text`.

pygments/lexers/maxima.py
```diff
@@ -0,0 +1,70 @@
...
+        ],
+        'comment': [
+            (r'[^*/]', Comment.Multiline),
+            (r'/\*', Comment.Multiline, '#push'),
+            (r'\*/', Comment.Multiline, '#pop'),
+            (r'[*/]', Comment.Multiline)
```
Anteru, line 68: This should be `+` so you don't create one token per character (see test output.)

Anteru (changes requested):
Thanks! I found a few minor things, please check the review comments.

pygments/lexers/maxima.py
robert-dodier: Hi @Anteru , I see what you mean about not creating one token per character, but I don't understand what you mean where you say "This should be +" -- what is it that should be `+`? Thanks for the clarification.

Anteru: The regex rule should have a `+` after it, so it will match multiple tokens. Right now `[^*/]` matches _one_ character that is not from the set `*`, `/`, and immediately emits a token. If you have a line with `ab` you'll thus get two tokens, one per character. If you convert it to `[^*/]`, `ab` gets matched as a single token `ab`.

pygments/lexers/maxima.py
robert-dodier: I put in a license blurb and removed the encoding comment.

pygments/lexers/maxima.py
robert-dodier: I reworked the lexer to make use of `words` instead of alternation in the regex.

pygments/lexers/maxima.py
robert-dodier: I've reworked the rules for Text and Text.Whitespace, so I think the classification is more accurate now.

pygments/lexers/maxima.py
robert-dodier: I've reworked the comment patterns so that the whole content of the comment is captured as one token.

robert-dodier:
Hi @Anteru  I'm sorry for the bother, but can I ask you to review this PR again? I made the changes you requested (and some other improvements as well).

Anteru:
I'll take another look this weekend. Please keep in mind that the next release will be in 2-3 months so it'll be a while until you'll see this packaged in a PyPI release.

robert-dodier:
> I'll take another look this weekend. Please keep in mind that the next release will be in 2-3 months so it'll be a while until you'll see this packaged in a PyPI release.

Sounds good, you can let me know how it goes. There's no hurry, I just want to make sure the wheels are in motion. Thanks a lot for your help.

Anteru:
Merged, thanks a lot!

PR #1890, opened 2021-09-04
Allow printing lists as JSON.

### LINKED ISSUE #1437: Have pygmentize output as JSON instead
It'd be great if `pygmentize` had the option of emitting lists (`-L`) as `JSON`.

I could go learn Python enough to figure out how to call `get_lexers` to do this. But I'd rather just do:

`pygmentize -L lexers -f json` 

or similar.

Anteru:
Fixes #1437

### DIFF
--- a/pygments/cmdline.py
+++ b/pygments/cmdline.py
@@ -198,21 +198,17 @@ def main_inner(parser, argns):
             if v:
                 arg_set.add(k)
 
-        arg_set.remove('L')
+        arg_set.discard('L')
+        arg_set.discard('json')
 
         # json can be only used with L, so we remove L, and check if there is
         # only one option left (which must be JSON).
-        if len(arg_set) == 1 and 'json' not in arg_set:
+        if arg_set:
             parser.print_help(sys.stderr)
             return 2
-        elif len(arg_set) > 1:
-            parser.print_help(sys.stderr)
-            return 2
-
-        json = 'json' in arg_set
 
         # print version
-        if not json:
+        if not argns.json:
             main(['', '-V'])
         allowed_types = {'lexer', 'formatter', 'filter', 'style'}
         largs = [arg.rstrip('s') for arg in argns.L]
@@ -221,7 +217,7 @@ def main_inner(parser, argns):
             return 0
         if not largs:
             largs = allowed_types
-        if not json:
+        if not argns.json:
             for arg in largs:
                 _print_list(arg)
         else:
--- a/pygments/cmdline.py
+++ b/pygments/cmdline.py
@@ -177,7 +177,7 @@ def _print_list_as_json(requested_items):
             }
         result['styles'] = info
 
-    json.dump(result, sys.stdout, indent=4)
+    json.dump(result, sys.stdout)
 
 def main_inner(parser, argns):
     if argns.help:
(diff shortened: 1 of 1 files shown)


pygments/cmdline.py
```diff
@@ -150,20 +193,35 @@ def is_only_option(opt):
...
-        if not is_only_option('L'):
+        arg_set = set()
+        for k, v in vars(argns).items():
+            if v:
+                arg_set.add(k)
+
+        arg_set.discard('L')
+        arg_set.discard('json')
+
+        # json can be only used with L, so we remove L, and check if there is
```
birkenfeld, line 204: comment isn't accurate anymore

pygments/cmdline.py
```diff
@@ -135,6 +135,49 @@ def _print_list(what):
...
 
+def _print_list_as_json(requested_items):
+    import json
+    result = {}
+    if 'lexer' in requested_items:
+        info = {}
+        for fullname, names, exts, _ in get_all_lexers():
+            info[fullname] = {
+                'names': names,
+                'extensions': exts
```
birkenfeld, line 146: This is usually called `filenames` since it's a glob matching the whole filename.

Any reason you're not returning the mimetypes (third element)?

pygments/cmdline.py
```diff
@@ -135,6 +135,49 @@ def _print_list(what):
...
 
 
+def _print_list_as_json(requested_items):
+    import json
+    result = {}
+    if 'lexer' in requested_items:
+        info = {}
+        for fullname, names, exts, _ in get_all_lexers():
+            info[fullname] = {
+                'names': names,
```
birkenfeld, line 145: let's call this 'aliases'.

pygments/cmdline.py
Anteru: I compared with the current output and it wasn't printed there, so I omitted it here as well, but I have no feelings about adding it to the JSON output :)

pygments/cmdline.py
Anteru: Fixed, thanks!

pygments/cmdline.py
Anteru: Changed to filenames, and added mimetypes

pygments/cmdline.py
Anteru: Done

pygments/cmdline.py
```diff
@@ -135,6 +135,50 @@ def _print_list(what):
...
+    if 'style' in requested_items:
+        info = {}
+        for name in get_all_styles():
+            cls = get_style_by_name(name)
+            info[name] = {
+                'doc': docstring_headline(cls)
+            }
+        result['styles'] = info
+
+    json.dump(result, sys.stdout, indent=4)
```
legoktm, line 180: minor: Since the point of JSON is to be machine readable, I don't think you need to specify indent=4 to make it more human readable. I think most people who want to look at the JSON output would know how to run it through a pretty printer like jq.

legoktm (approved):
I tested this and it provides all the functionality I was looking for, thank you for working on it!

pygments/cmdline.py
Anteru: Oh absolutely, this was mostly for my debugging sanity :)

Anteru:
> 
> 
> I tested this and it provides all the functionality I was looking for, thank you for working on it!

 Thanks a lot for testing & confirming this works!

PR #1894, opened 2021-09-08
Lexer for new language Elpi

gares:
Well, the title says it all.

### DIFF
--- a/pygments/lexers/elpi.py
+++ b/pygments/lexers/elpi.py
@@ -29,7 +29,7 @@ class ElpiLexer(RegexLexer):
     lcase_re = r"[a-z]"
     ucase_re = r"[A-Z]"
     digit_re = r"[0-9]"
-    schar2_re = r"(\+|\*|/|\^|<|>|`|'|\?|@|#|~|=|&|!)"
+    schar2_re = r"([+*^?/<>`'@#~=&!])"
     schar_re = r"({}|-|\$|_)".format(schar2_re)
     idchar_re = r"({}|{}|{}|{})".format(lcase_re,ucase_re,digit_re,schar_re)
     idcharstarns_re = r"({}+|(?=\.[a-z])\.{}+)".format(idchar_re,idchar_re)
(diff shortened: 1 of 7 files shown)


gares:
Thanks for running CI!
I did fix the check and tests. I could not run `make check` locally, so maybe the copyright banner is still wrong.

gares:
Thanks again for running CI.
This time I need help understanding the results, since the failure seems to be about the HTML backend which I did not touch, and about a missing pyflakes which seems unrelated as well.

blu-base:
I think the failed make check is only because the copyrighted period is expected to be `2006-2021`.
The other tests seem to have failed due to timeouts. maybe they are not related with your PR.

Anteru:
Hm, I can't re-run the workflow, which is weird :(

gares:
@blu-base can you unblock this please?

blu-base:
Sorry, @gares , I can't. I am just a contributor myself.

Anteru:
This is broken somewhere else. I need to run this manually to merge but I need to find some spare time to work on this.

birkenfeld:
Hey the old "close and open" trick still works 🎆

Anteru:
What the ...?

birkenfeld:
I think the "rerun" feature somehow depends on details of the old runs, and GitHub doesn't keep those around that long. What happens on "close and reopen" or if someone pushes some new changes is a completely fresh run.

pygments/lexers/elpi.py
```diff
@@ -0,0 +1,143 @@
...
+            (r"\b(external pred|pred)(\s+)({})".format(const_sym_re),bygroups(Keyword.Declaration,Text,Name.Function),'elpi-pred-item'),
+            (r"\b(external type|type)(\s+)(({}(,\s*)?)+)".format(const_sym_re),bygroups(Keyword.Declaration,Text,Name.Function),'elpi-type'),
+            (r"\b(kind)(\s+)(({}|,)+)".format(const_sym_re),bygroups(Keyword.Declaration,Text,Name.Function),'elpi-type'),
+            (r"\b(typeabbrev)(\s+)({})".format(const_sym_re),bygroups(Keyword.Declaration,Text,Name.Function),'elpi-type'),
+            (r"\b(accumulate)(\s+)(\")",bygroups(Keyword.Declaration,Text,String.Double),'elpi-string'),
+            (r"\b(accumulate|namespace|local)(\s+)({})".format(constant_re),bygroups(Keyword.Declaration,Text,Text)),
+            (r"\b(shorten)(\s+)({}\.)".format(constant_re),bygroups(Keyword.Declaration,Text,Text)),
+            (r"\b(pi|sigma)(\s+)([a-zA-Z][A-Za-z0-9_ ]*)(\\)",bygroups(Keyword.Declaration,Text,Name.Variable,Text)),
+            (r"\b(constraint)(\s+)(({}(\s+)?)+)".format(const_sym_re),bygroups(Keyword.Declaration,Text,Name.Function),'elpi-chr-rule-start'),
```
Anteru, line 59: `\s+` should be matched as `Text.Whitespace`, not `Text`.

pygments/lexers/elpi.py
```diff
@@ -0,0 +1,143 @@
...
+    aliases = ['elpi']
+    filenames = ['*.elpi']
+    mimetypes = ['text/x-elpi']
+    
+    lcase_re = r"[a-z]"
+    ucase_re = r"[A-Z]"
+    digit_re = r"[0-9]"
+    schar2_re = r"(\+|\*|/|\^|<|>|`|'|\?|@|#|~|=|&|!)"
+    schar_re = r"({}|-|\$|_)".format(schar2_re)
+    idchar_re = r"({}|{}|{}|{})".format(lcase_re,ucase_re,digit_re,schar_re)
```
Anteru, line 34: Instead of merging three separate rules here, `[\w\-$_]` would also do the trick and is much easier to parse for me. Similarly, I'd prefer `[<>+...]` instead of `+|<|>|...` as it reduces the "noise" while reading.

pygments/lexers/elpi.py
```diff
@@ -0,0 +1,143 @@
...
+            (r'\{[^\{]', Text, 'elpi-spill'),
+            (r"\(",Text,'elpi-in-parens'),
+            (r'\d[\d_]*', Number.Integer),
+            (r'-?\d[\d_]*(.[\d_]*)?([eE][+\-]?\d[\d_]*)', Number.Float),
+            (r"[\+\*\-/\^\.]", Operator),
+        ],
+        '_elpi-comment': [
+            (r'%[^\n]*\n',Comment),
+            (r'/\*',Comment,'elpi-multiline-comment'),
+            (r"\s+",Text),
```
Anteru, line 79: Same as above, `\s` should be matched `Text.Whitespace` (not going to highlight the remaining cases.)

tests/snippets/elpi/test_pred.txt
```diff
@@ -0,0 +1,61 @@
...
+
+'pred'        Keyword.Declaration
+' '           Text
+'p3'          Name.Function
+' '           Text
+'i:'          Keyword.Mode
+'('           Keyword.Type
+'bool'        Keyword.Type
+' '           Text
+'-'           Error
```
Anteru, line 37: This is missing some rule, we don't allow error tokens.

Anteru (changes requested):

Anteru:
@birkenfeld This somehow passed the tests despite producing an `Error` token. Do you happen to know what is going on here? I thought we never allow `Error` tokens to be produced.

birkenfeld:
Strange indeed, maybe it isn't caught anymore when it has been "blessed" as correct with the output file.

The easiest way to fix this is probably to add the `raiseonerror` filter.

birkenfeld:
Ah, the `snippets` series of tests *does* explicitly allow errors. I think this is correct, since you also should be able to check that erroneous inputs *do* give error tokens.

Not much to fix here except closely read the blessed output for new contributions ...

gares:
Thanks for the review, I'll try to fix that tomorrow!

pygments/lexers/elpi.py
gares: done

pygments/lexers/elpi.py
gares: done

tests/snippets/elpi/test_pred.txt
gares: the discussion seems to say that this is OK

gares:
hopefully done

tests/snippets/elpi/test_pred.txt
Anteru: Is this an error in the source language? I assume this is a valid piece of code and in that case Error is not OK. I'd also prefer error cases to be in a separate file instead of mixed with the rest because they are exceptional.

gares:
Sorry for the slow reaction. I've updated the golden since I now tag spaces with Text.Whitespace instead of just Text

pygments/lexers/elpi.py
gares: I've addressed the second issue. The former, I'm not too happy, since I have this grammar in 3 places already (here, the actual parser, and vscode) and in all 3 I use these non-terminals (schar, schar2, ...). So I'd rather keep them in sync.

If this is blocking, I'll amend. Otherwise bare with me.

gares:
OK, thanks for running CI again. This time it is green.

gares:
Wow, thanks!

Anteru:
Thanks for the contribution and sorry it took so long to merge. This quarter has been unusually busy in terms of contributions overall!

Issue #1806, opened 2021-05-16
New BibLaTeX-entry @commentary


titanic-jb:
BibLaTeX defines a `@commentary` entry, which should be used, to store bibliographical information of special legal documents.   But it is displayed as comment type, because it shares the same root as the `@comment`-entry, which is a comment type.  

Please update `pygments` to obeye the difference between `@comment` and the new `@commentary`-entry.

Anteru:
Fixed!

PR #1899, opened 2021-09-25
Fix Bibtex comment/commentary keyword difference

blu-base:
This PR is a fix for #1806

A negative lookahead will jump over a `@commentary` keyword avoiding mislexing the `@comment` keyword.


Anteru:
Merged, thanks!

PR #1904, opened 2021-10-01
Fix #1237 cpp whitespace token usage expanded

### LINKED ISSUE #1237: Broken C++ Class Highlighting
(Original issue 1533 created by None on 2019-07-20T14:08:30.735114+00:00)

I am using https://github.com/gohugoio/hugo for a personal blog.

evidence
code
config

blu-base:
It seems likely, that the original issuer had a highlighting theme where the token `Text` has been highlighted.

Whereever applicable, the `Text` token is replaced with the more specific `Whitespace` token - not just the `class` keyword, but also similar occurrences, such as enums and structs.

This should fix #1237, and should generally be more precise lexing as well.

### DIFF
--- a/pygments/lexers/c_cpp.py
+++ b/pygments/lexers/c_cpp.py
@@ -82,7 +82,7 @@ class CFamilyLexer(RegexLexer):
             (r'[~!%^&*+=|?:<>/-]', Operator),
             (r'[()\[\],.]', Punctuation),
             (r'(true|false|NULL)\b', Name.Builtin),
-            (r'(' + _ident + r')(\s*)(:)(?!:)', bygroups(Name.Label, Text, Punctuation)),
+            (r'(' + _ident + r')(\s*)(:)(?!:)', bygroups(Name.Label, Whitespace, Punctuation)),
             (_ident, Name)
         ],
         'types': [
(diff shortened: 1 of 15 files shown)


blu-base:
I did not run the tests initially... The proposed change impacts more places than i thought initially.

It seems there are also more cases of Keyword + Text instead of Keyword + Text.Whitespace. (run grep -r "Keyword, Text in lexers)
That raises the question, is this a design choice?

Anteru:
I've been trying to _slowly_ clean up existing lexers and be more attentive to `Text` vs. `Whitespace` in new lexers. It's a pervasive problem to be honest, I wasn't aware how much of a problem it actually already is. This is definitely not by design, but mostly because we didn't pay enough attention to it. If you're interested in fixing this in more places this would be most welcome!

blu-base:
> I've been trying to _slowly_ clean up existing lexers and be more attentive to `Text` vs. `Whitespace` in new lexers. It's a pervasive problem to be honest, I wasn't aware how much of a problem it actually already is. This is definitely not by design, but mostly because we didn't pay enough attention to it. If you're interested in fixing this in more places this would be most welcome!

This sounds like a problem i can handle. Just a question for the process, would you like individual PRs per changed lexers?

Anteru:
Merged, thanks a lot for working on this!

Anteru:
Thanks for picking this up! I think one PR for everything should be fine given it's fairly easy to review. If you're building any scripts to automate it, it would be great if you could submit them as well should we ever need to go through this exercise again.

PR #1907, opened 2021-10-04
Erb mapping

Murray2015:
.erb files currently match nothing. This tiny fix enables .erb highlighting.


Jean-Abou-Samra (changes requested):
The file `_mapping.py` is not maintained by hand but generated automatically. To update it, you need to change the lexer definition and run `make mapfiles`.

Generally speaking I am not sure that this change is correct: ERB seems to be a templating language used with various target languages, so maybe the `ErbLexer` is just intended to be subclassed? My worry is that for a `.erb` file, you'll no longer get automatic detection based on `analyse_text` (see the respective methods of `RhtmlLexer`, `XmlErbLexer`, etc.) and the templated material in the target language will no longer get highlighted. Can you test this?

Issue #1905, opened 2021-10-02
WIP - Whitespace lexing after Keywords - collecting issue


blu-base:
In a recent PR, it become apparent, that there are a lot of places where the whitespace following a Keyword has been often implemented as the `Text` token. Instead the more specific `Text.Whitespace` token is more applicable here.

This issue shall serve as reference and log for similar issues/PRs.

Since initial tests showed a large number of lexers which can be updated, a more systematic approach is tried with the table below. The table lists all lexers in pygments, the status of work, and related issues and PRs. The status shows whether a quick check shows a need for change, whether Whitespace token are already used (hence nothing needs to be done in these cases), or whether the lexers still needs to be inspected whether there is this issue.

## State of the cleanup

| Lexer | Status | PRs and issues |
| --- | --- | --- |
| actionscript | :heavy_check_mark: merged |  #1911 |
| agile | :white_check_mark: nothing to do| |
| algebra | :white_check_mark: nothing to do| |
| ambient | :heavy_check_mark: merged  | #1914 |
| amdgpu | :white_check_mark: nothing to do| |
| ampl | :heavy_check_mark: merged  | #1914 |
| apdlexer | :heavy_check_mark: merged  | #1914 |
| apl | :heavy_check_mark: merged  | #1914 |
| archetype | :heavy_check_mark: merged  | #1914 |
| arrow | :heavy_check_mark: merged  | #1914 |
| asc | :white_check_mark: nothing to do| |
| asm | :heavy_check_mark: merged  | #1914 |
| automation | :triangular_flag_on_post: more elaborate change necessary  |  |
| bare | :heavy_check_mark: merged |  #1911 |
| basic |  :heavy_check_mark: merged  | #1923|
| bibtex |   :heavy_check_mark: merged | #1923|
| boa |  :heavy_check_mark: merged  | #1923|
| business | :heavy_check_mark: merged :curly_loop: needs refactor  | #1911 |
| capnproto |  :heavy_check_mark: merged  | #1923|
| c_cpp | :heavy_check_mark: merged |   #1237(#1904) |
| cddl |  :heavy_check_mark: merged  | #1923|
| chapel | :heavy_check_mark: merged  | #1923|
| clean | :white_check_mark: nothing to do| |
| c_like |  :heavy_check_mark: merged  | #1923|
| compiled | :white_check_mark: nothing to do| |
| configs |  :heavy_check_mark: merged  | #1923|
| console | :heavy_check_mark: merged  | #1923|
[...]

PR #1908, opened 2021-10-04
lexer SQL to use whitespace token - regarding #1905

blu-base:
This PR specifes whitespace-only matches with the respective Text.Whitespace token.
It is submitted in the context of #1905. The changes are minimal, and related test files have been adapted.

----

Anyhow, I noticed a possible inconsitency and would like a second opinion on how to deal with it.

While the following regex does not include the trailing white space, the sqlite prompt does.
https://github.com/pygments/pygments/blob/339ad2b31f5359dc0e554399a35a464ec839c62b/pygments/lexers/sql.py#L266
https://github.com/pygments/pygments/blob/339ad2b31f5359dc0e554399a35a464ec839c62b/pygments/lexers/sql.py#L786-L801

In this state there are Prompt Token lexed such as tests/examplefiles/psql/psql_session.txt.output line 27 (matched by the re_prompt regex) which are trailed by a Whitespace token. However, the `sqlite> ` prompt does include the trailing space in the token group (matched via the startswith expressions). The latter has an example in tests/examplefiles/sqlite3/sqlite3.sqlite3-console.output line 98 (at the end of the diff)

### DIFF
--- a/pygments/lexers/sql.py
+++ b/pygments/lexers/sql.py
@@ -161,7 +161,7 @@ class PostgresLexer(PostgresBase, RegexLexer):
     flags = re.IGNORECASE
     tokens = {
         'root': [
-            (r'\s+', Text),
+            (r'\s+', Whitespace),
             (r'--.*\n?', Comment.Single),
             (r'/\*', Comment.Multiline, 'multiline-comments'),
             (r'(' + '|'.join(s.replace(" ", r"\s+")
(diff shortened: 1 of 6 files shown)


Anteru:
Thanks for preparing this change! On the SQLite prompt: I'd like to see it produce whitespace and not put it into the prompt itself, similar to the PgSQL prompt. I assume if you add manual whitespace after the prompt it will actually highlight it as whitespace? In which case all whitespace should have been marked as such.

blu-base:
The two additional commits deal with the sqlite prompt, and therefore complete this PR from my perspective.

Thanks to the well prepared test file, i caught a bug I'd introduce with the second commit only. Leaving the trailing whitespace "in the open" for further lexing, it could result in grouping the space with other long running tokens, such as string literals in the test file.

PR #1912, opened 2021-10-10
Support HCL's heredoc syntax

### LINKED ISSUE #1909: Terraform lexer does not support HCL heredoc syntax
I am using pygments 2.10.0 to highlight a terraform config containing this:

```hcl
resource "local_file" "tf_ansible_vars_file_new" {
  content  = <<-DOC
    # Ansible vars_file containing variable values from Terraform.
    # Generated by Terraform mgmt configuration.

    iot_credential_provider_endpoint: ${data.aws_iot_endpoint.credentials.endpoint_address}
    iot_credential_role_alias: ${aws_iot_role_alias.cert-dns.alias}
    DOC
  filename = "./vars/tf_ansible_vars_file.yml"
}
```

What I get out of it is the attached image. 
image

offbyone:
Fixes #1909


Anteru:
Thanks for preparing this. While this _is_ super elegant, it's unfortunately resulting in catastrophic backtracking. Try it with:
`<<'DDfffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffDD` (example https://regex101.com/r/bZprT4/1)

I suspect you'll want something like the PowerShell approach to change state, instead of backtracking.

offbyone:
Yikes! 

I will be honest, I wasn't really able to follow what the Powershell version was doing, and I modeled this off of the bash shell, instead. Okay, I'll go back to it and see if I can grok what's happening there. Mostly, I wasn't sure how to have the opening delimiter carry forward to be the closing delimiter; powershell has static delimiters for the heredoc, while HCL supports user-selected ones.

offbyone:
So, for this, I'm thinking that I should enter the heredoc state like this:

```
            # here-doc style delimited strings
            (r'(<<-?)\s*(?:\'?)\\?(\w+)', (bygroups(Operator, String.Delimiter)), 'heredoc'),
```

That gives me the terminating delimiter. The question is, how do I access that delimiter in the `heredoc` state so that I can pop out of it when it's encountered? Basically, exiting the state occurs on `r'^\s*DELIMITER'`

offbyone:
I'm reading all the other lexers that have versions of this document type that I can find, and it seems that there's three classes:

1. PowerShell style, where heredocs have no variable structure to them
2. Ruby style where the logic is so ridiculously complex there's a massive callback just to handle them
3. Bash style, where this backtracking regexp is what's in there already

I think that if the bash shell lexer has survived with this backtracking problem for this long, odds are good that the terraform one will survive with it, too. Am I missing something?

offbyone:
I buy that :)

I'd be happy to contribute; I'm still trying to grok if there's an obvious way to get the parser to tell the next context what token was sent in to it, and let rules in the context reference that. I suspect there is not :/

birkenfeld:
I think most of the different languages' heredoc syntaxes are pretty similar?  We can move that callback into a util module which keeps the lexers trim. Still need to be ExtendedRegexLexers though, which comes with a performance penalty (which I have never measured...)

Anteru:
Merged manually, thanks!

PR #1914, opened 2021-10-10
Whitespace token modernization - a* lexers - regarding #1905

blu-base:
This PR is chunk of the effort (#1905)  to insert the Whitespace token where ever it applies.

The automation lexer also contains a minor fix for multiline comments (making the multiline-comment-content regex greedy).

### DIFF
--- a/pygments/lexers/asm.py
+++ b/pygments/lexers/asm.py
@@ -15,7 +15,7 @@ from pygments.lexer import RegexLexer, include, bygroups, using, words, \
 from pygments.lexers.c_cpp import CppLexer, CLexer
 from pygments.lexers.d import DLexer
 from pygments.token import Text, Name, Number, String, Comment, Punctuation, \
-    Other, Keyword, Operator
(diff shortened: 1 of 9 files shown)


blu-base:
Further playing with the automation lexer wasn't very fruitful...
Having a brief look into the ahk tutorial shows, there is a feature to label hotkey functions and setting options: https://www.autohotkey.com/docs/Hotstrings.htm

I tried to get detect the labels correctly for the following snippet. 
```ahk
#p::
Run, https://www.pygments.org/
return

^t::
Run, calc.exe
return

:*:pyg::pyg
::pygmentize::pygmentize
```
This snippet results in differing tokens as well. `:*:pyg::` and `::pygmentize::` are so called *hotstrings* (labels for a hotkey sequence), but one receives the plain `Name` token and the other `Name.Label`.

----

Anyhow, I am unable to find a quick solution. Rethinking this lexer would require some time... I'd drop the respective commit stashing it for later. Maybe I have a better understanding on pygments' code then - or someone else will address this...

PR #1915, opened 2021-10-10
WIP - Test for Whitespace token - Don't merge yet

blu-base:
This included test might be a way to "enforce" the usage of the `Whitespace` token. The trivial case of a single whitespace is lexed by every lexer in `pygments.lexers.LEXER`. The way it is implemented now, allows either two tokens of type `Whitespace` for `' '` and `'\n'` or a single Token for `' \n'`of type `Whitespace`.  Having this test fully pass can be seen as the goal of #1905.

```bash
$ pytest tests/test_trivial_whitespace.py
[...]
============= 426 failed, 82 passed in 8.55s ========
```

There are so many fails right now. For this reason, I think it would be good to **not merge** this PR any time soon. 

Any opinion on this?

### DIFF
--- a/tests/test_trivial_whitespace.py
+++ b/tests/test_trivial_whitespace.py
@@ -1,0 +1,33 @@
+"""
+    Pygments whitespace test
+    ~~~~~~~~~~~~~~~~~~~~~~~~
+
+    This test checks whether a single whitespace is lexed with the
+    Whitespace token.
+
+    :copyright: Copyright 2006-2021 by the Pygments team, see AUTHORS.
+    :license: BSD, see LICENSE for details.
+"""
+
+import pytest
+
+from pygments import lexers
+from pygments.token import Whitespace
+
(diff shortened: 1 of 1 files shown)


blu-base:
Current state;

` 370 failed, 138 passed in 7.34s `

blu-base:
With my latest PR, the current status is:
`263 failed, 254 passed`

Anteru:
Yes, I have an opinion (and thanks a lot for preparing this!) We'll want that eventually integrated to ensure that all lexers pass; potentially with an explicit whitelist of lexers for now so we can enforce this on new lexers. There's also https://github.com/pygments/pygments/blob/master/scripts/check_repeated_token.py which probably should end up in a test as well and whitelist the existing lexers. The more future additions can be automatically tested, the better.

Jean-Abou-Samra:
(Setting "update needed" just to trim down the list of PRs to be reviewed, as this one isn't ready yet.)

PR #1924, opened 2021-10-21
Creating one dark style

TobiZog:
I creating a new style, orientated by the "One Dark Theme" for Atom. I hope you will add it to the package.

### DIFF
--- a/pygments/styles/onedark.py
+++ b/pygments/styles/onedark.py
@@ -1,6 +1,6 @@
 """
     pygments.styles.onedark
-    ~~~~~~~~~~~~~~~~~~~~~~
+    ~~~~~~~~~~~~~~~~~~~~~~~
 
     One Dark Theme for Pygments by Tobias Zoghaib (https://github.com/TobiZog)
 
--- a/pygments/styles/onedark.py
+++ b/pygments/styles/onedark.py
@@ -1,5 +1,5 @@
 """
-    pygments.style.onedark
+    pygments.styles.onedark
     ~~~~~~~~~~~~~~~~~~~~~~
 
     One Dark Theme for Pygments by Tobias Zoghaib (https://github.com/TobiZog)
--- a/pygments/styles/onedark.py
+++ b/pygments/styles/onedark.py
@@ -1,0 +1,44 @@
+from pygments.style import Style
+from pygments.token import Token, Punctuation, Keyword, Name, Comment, String, \
+     Number, Operator, Generic, Whitespace
+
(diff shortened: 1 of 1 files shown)


TobiZog:
Can you please approve this?

Anteru:
Merged, thanks!

ranelpadon:
@TobiZog @Anteru Thanks a lot for this **One Dark** theme, which is my favorite. This is timely PR also since I'm tinkering with IPython/Pygments recently, and will write an article for it. Hopefully, this will be released to PyPI asap since the last release was still **2.10.0** (August 15), and so that I could just say `pip install pygments` to the readers. Otherwise, need to instruct them to do some manual changes to the **styles** folder also. :)

ranelpadon:
@TobiZog I'm using One Dark theme in my **Alacritty** terminal and **Vim**/**Neovim**/**MacVim** editors. I just noticed that some colors in Pygments' One Dark theme are bit off when running IPython. When I dig further, there are indeed subtle differences in some of the colors used. The **top** colors are the Pygments' One Dark, the **bottom** colors are the "expected" colors:

Screen Shot 2021-12-04 at 3 37 42 PM

I recommend to use the [Vim's One Dark](https://github.com/joshdick/onedark.vim/blob/main/colors/onedark.vim) or [Alacritty's One Dark](https://github.com/rajasegar/alacritty-themes/blob/master/themes/One-Dark.yml) as the "expected" values since they are quite popular and their color values are consistent with each other. Likewise, I checked with [Atom's original HSL](https://github.com/atom/atom/blob/master/packages/one-dark-syntax/styles/colors.less) values, they are correct/consistent. Other colors in Pygments' One Dark are correct/consistent though. 

Is it okay to submit a PR to refine them? Thanks :)

CC: @Anteru

Anteru:
@ranelpadon I'm ok with a PR but I'm afraid it's going to make the contrast even lower. I guess we'll see once you submit it. We're trying to be a bit more conscious about accessibility and not have tons of new styles with bad contrast.

ranelpadon:
@Anteru I don't think the contrast is an issue here since it's being used in other popular open-source projects for a long time already including VS Code/PyCharm. I value consistency, the Atom's One Dark version I think should be the source of truth since they were the creator/pioneer of it. The fans of One Dark theme expect the same colors, regardless if it's high contrast or low contrast for other users/non-fans. They're fans for a reason. They love the exact values as is. Anyway, I'll submit a PR. Cheers

PR #1928, opened 2021-10-26
Clarifying some documentation

Arthur-Milchior:
While reading Pygments documentation, there were parts that confused me. That
required that I read the code to understand them. I hope that this contribution
will help clarify for future readers.

I do not know pygments usage. I have chosen to do atomic commits so that if you agree that some commits are an improvement and others need more work, we can cherry pick the good one.


00251911391303 (commented):
Muhamed

PR #1929, opened 2021-10-26
smithy: remove single-quoted strings and add text blocks

JordonPhillips:
I was updating the dependencies for Smithy's  docs when I noticed that pygments has a lexer built in now! Very cool. I noticed a few issues with how strings and text blocks are handled though.

This removes support for single-quoted strings from the Smithy lexer, as they were removed before Smithy hit 1.0. Additionally, this adds proper support for text blocks. Double quotes more or less worked for text blocks, except that they disallowed internal double quotes and escaping triple-double quotes.

### DIFF
--- a/tests/examplefiles/smithy/test.smithy
+++ b/tests/examplefiles/smithy/test.smithy
@@ -4,9 +4,12 @@ namespace test
 
 @documentation("""
 "Embedded quotes are permitted"
-Triple quotes may be escaped: \"""
 """)
 string TestTextBlock
+
+// This demonstrates how one might escape triple quotes.
+@documentation(""" \""" "\"" ""\" ""\"""")
+string TestTripleQuoteEscape
 
 /// Define how an HTTP request is serialized given a specific protocol,
 /// authentication scheme, and set of input parameters.
--- a/tests/examplefiles/smithy/test.smithy.output
+++ b/tests/examplefiles/smithy/test.smithy.output
@@ -9,12 +9,25 @@
 
 '@documentation' Name.Decorator
 '('           Text
-'"""\n"Embedded quotes are permitted"\nTriple quotes may be escaped: \\"""\n"""' Literal.String.Double
+'"""\n"Embedded quotes are permitted"\n"""' Literal.String.Double
 ')'           Text
 '\n'          Text.Whitespace
 
 'string'      Keyword.Declaration
 ' TestTextBlock' Name.Class
+'\n\n'        Text.Whitespace
+
+'// This demonstrates how one might escape triple quotes.' Comment
+'\n'          Text.Whitespace
+
+'@documentation' Name.Decorator
+'('           Text
+'""" \\""" "\\"" ""\\" ""\\""""' Literal.String.Double
+')'           Text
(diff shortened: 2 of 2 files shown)


JordonPhillips:
That CI failure seems unrelated - perhaps something flaky got merged elsewhere? I'll rebase to see if it has been fixed.

Anteru:
For `"""\\""""""`, is that supposed to be a string containing `"""`? I don't quite understand how the escaping works, and I'm inclined to believe you probably want to enter a new state when you find a `"""` to make it clearer as to what is happening.

JordonPhillips:
> For `"""\\""""""`, is that supposed to be a string containing `"""`?

Three consecutive un-escaped dquotes ends the text block, so you would need to escape the third quote: `""" ""\""""`

JordonPhillips:
I updated the tests to show all the different ways you might escape a set of triple quotes, and I think what exists now should be fine. If you'd still like me to move this to another state, I do have a question: how would I exit the state given that starting and stopping it are the same pattern?

Anteru:
Once you enter the state, only the rules in the state apply, so you'd leave by looking at the token that gets you out of it. If you look at the "triple quoted strings" in ``python.py``, you can see:

```python
        'tdqf': [
            (r'"""', String.Double, '#pop'),
            include('fstrings-double'),
            (r'\n', String.Double)
        ],
```
Meaning the only way to get out of that state is by using `"""`.

Anteru:
So I've been looking at this once more, and I'm confused how this is supposed to work. Looking at:
```
""" \\""" "\\"" ""\\" ""\\""""
^^^ This starts a triple-quoted string
    ^^ This is a backslash-escape, I presume?
      ^^^ This is a triple quoted string again
```

I assume that inside a triple quoted string, `\\` is not an escaped `\`, but there's another level of indirection happening here? I.e. `\\"` within a triple-quoted string yields `\"` or something?

JordonPhillips:
That bit you're quoting is from the test case output, so those extra back slashes are because they're embedded in a python string and thus need to be escaped for that context. The source is:

```
@documentation(""" \""" "\"" ""\" ""\"""")
               ^^^ starts the triple-quoted string
                   ^^ escapes the first dquote, which escapes the
                   triple-double
                         ^^ escapes the second dquote, which also escapes the
                         triple-double
                                    ^^ escapes the third dquote, again
                                     escaping the triple-double
                                      ^^^ ends the triple-double
```

pygments/lexers/smithy.py
```diff
@@ -71,8 +71,8 @@ class SmithyLexer(RegexLexer):
             (r'\{', Text, "#push"),
             (r'\}', Text, "#pop"),
             (r'"{3}(\\\\|\n|\\")*"{3}', String.Doc),
+            (r'"""(?:.|\n)*?[^\\]"""', String.Double),
```
Jean-Abou-Samra, line 74: Is an empty string not allowed? This regex always requires at least one character inside it.

Also, see `Contributing.md` about lexing strings: you probably want a dedicated state, so as to highlight string escapes.

PR #1934, opened 2021-10-28
Remove unnecessary assignments in Paraiso style

ThomsonTan:
Remove the duplicated assignments to `background_color` and `highlight_color` in both paraiso_light and paraiso_dark styles.


Anteru:
Nice catch, thanks!

PR #1935, opened 2021-10-31
Add a lexer for the sed scripting language

Mango0x45:
Sed is a stream editing scripting language forming part of the POSIX
standard. This means that sed can be found as part of any POSIX compliant
OS such as Linux, MacOS, and the BSDs. This lexer supports the complete
POSIX standard for sed as well as various GNU extensions.

The names 'gsed' and 'ssed' may sometimes be used as part of a file
extension to refer specifically to GNU sed or Super Sed, two popular
implementations of sed.

The sed language is rather simple, here is a quick overview:

- Comments begin with a '#' and continue to the end of the line.
- Commands can be prepended with a range that selects lines. This can be
  either a number ([0-9]+), a '$', or a regular expression. Regular
  expressions can either be enclosed in '/' such as /regex/ or they can
  be exclosed in any other character however the opening delimiter must
  be escaped, for example: \\@regex@.
- Some commands such as 'w', 'r', and 'b' take a string as a parameter,
  this string extends either to the next ';' or newline.
- Some commands such as 'a' can take multiline strings.
- Commands can be grouped with braces ('{' and '}').
- The 'y' command takes two strings as parameters, seperated by any
  delimiter ('y/abc/123/' or 'y|abc|123|')
- The 's' command is like the 'y' command but the fist parameter is a
  regular expression. Additionally it takes an optional 3rd parameter
  which are a sequence of flags ('s|regex|replace|flags').

I chose to add support for GNU sed extensions as it is by far the most
used implementation.


Anteru:

PR #1936, opened 2021-11-01
enhance: support vagrantfile

tzing:
[Vagrant](https://www.vagrantup.com/docs/vagrantfile) is a tool for managing the virtual environments.
It uses a hardcoded config file name `Vagrantfile` and uses syntax of Ruby.
Would like to support it through ruby's lexer.


Anteru:
Merged, thanks!

PR #1938, opened 2021-11-02
Make test_contrasts.py ignore token.Whitespace

not-my-profile:
With the VisibleWhitespaceFilter Pygments can be configured to replace
whitespace characters like spaces, tabs and newlines with Unicode
characters like ·, » and ¶ respectively.

The min_contrast metric is meant to represent the readability of a
style. Increasing it should mean that the style becomes more readable.

This commit therefore excludes the token.Whitespace contrast from the
min_contrast calculation, since giving whitespace characters a high
contrast would make them so intrusive that they would actually hinder
the readability of the source code.

## Screenshots for illustration

Current default style:

image

WCAG AA would look like (which IMO is clearly less readable):

image


Anteru:
Good catch. Another good reason to be more precise about whitespace handling, but that is in @blu-base capable hands.

not-my-profile:
Thanks :) Though I don't quite get your comment on whitespace precision.

blu-base:
In many lexers whitespaces are detected as the `Text` token, not the `Text.Whitespace` token.
It seems this has historical reasons, since the `Whitespace` token has only been introduced in the later development.
For this reason, there are many lexers which cannot make use of the `VisibleWhitespaceFilter`, yet - since there are no `Whitespace` token declared, or only partially.

I started to review most of them in order to fix that inhomogenity. Though i just scratched the surface (#1905)

PR #1942, opened 2021-11-03
Make pygments.org website responsive

### LINKED ISSUE #1927: Pygment code examples are displayed far lower than they should be
The pygment documentation has a very strange spacing.  At least on Version 94.0.4606.81 (Official Build) snap (64-bit)
It looks like 
2021-10-26-015129_1068x796_scrot

or without js:
2021-10-26-015148_1054x774_scrot

That makes it slightly harder to read than necessary as it seems some part of the page didn't load, something is missing, and the code is not naturally the next place to go in the document order.

A quick CSS search seems to indicate that the trouble comes from the clear:both; which is interpreted as meaning that the code should be below the left-side menu bar

not-my-profile:
Also fixes #1927.

### DIFF
--- a/doc/_templates/indexsidebar.html
+++ b/doc/_templates/indexsidebar.html
@@ -1,3 +1,4 @@
+<section>
 <h3>Download</h3>
 {% if version.endswith('(hg)') %}
 <p>This documentation is for version <b>{{ version }}</b>, which is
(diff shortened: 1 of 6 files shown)


birkenfeld:
Awesome! Looks good to me, and we can always improve it further when needed.

Anteru:
Indeed, very much appreciated. Thanks!

PR #1943, opened 2021-11-03
Add styles gallery to website

not-my-profile:
image


birkenfeld:
+1 for the idea! I was about to propose to put this in `pygments.sphinxext` with the rest of the custom code we have, but then this isn't exactly easy to do.

Can I request a different example snippet? Some programming language, probably Python as it is the most neutral here.

not-my-profile:
I changed the example to a Python snippet (which I specifically designed to include many token types).

birkenfeld:
Perfect, thanks!

PR #1944, opened 2021-11-04
Post Style gallery fixes

not-my-profile:
Forgot to test it with `make dirhtml` (I was using `make html`).

### DIFF
--- a/doc/_templates/styles.html
+++ b/doc/_templates/styles.html
@@ -12,6 +12,9 @@
 .style-gallery h2 {
     margin-bottom: 0.2em;
 }
+.style-gallery pre {
+    background-color: inherit;
+}
 </style>
 {{ body }}
 
--- a/doc/_themes/pygments14/static/pygments14.css_t
+++ b/doc/_themes/pygments14/static/pygments14.css_t
@@ -338,6 +338,7 @@ pre {
     padding: 0.5em;
     border: 1px solid #ccc;
     border-radius: 2px;
+    background-color: #f8f8f8;
 }
 
 pre a {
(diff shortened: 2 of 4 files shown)


birkenfeld:
So it only works with one or the other? That is unfortunate...

not-my-profile:
Yeah it is ... and I can't think of a good workaround.

* `make html` produces: `docs/styles.html` and `styles.html`, so `../styles.html` would be the right link
* `make dirhtml` produces `docs/styles/index.html` and `styles/index.html`, so `../../styles/` would be the right link

While for the other direction we can use the Sphinx `pathto` helper function, unless I am mistaken that isn't available from RST.

birkenfeld:
The workaround is to add a file `doc/styles.rst` with only `:orphan:` as its content and to use :doc:`link text here </styles>`.

not-my-profile:
Haha, right ... forgot that you are an expert on Sphinx^^ Now it's really fixed :)

not-my-profile:
Added another commit that fixes the missing background color in the sidebar (unintended side-effect by the styles gallery commit).

I get that you tend to merge PRs by squashing all commits together ... I don't really like that^^ ... if commits do separate things I prefer to keep them separate ... makes for a more readable history.

birkenfeld:
If you prefer that, no problem! Thanks for the attention to detail :)

birkenfeld:
Argh, github defeated me...

birkenfeld:
Ok, manually fetched and rebased master to your two commits.

Anteru:
Thanks!

PR #1946, opened 2021-11-04
Update debian control

sblondon:
This PR adds a test file for debian/control file lexer and fixes some fields. The changes are based on:

https://www.debian.org/doc/debian-policy/ch-controlfields.html
https://www.debian.org/doc/debian-policy/ch-relationships.html

### DIFF
--- a/pygments/lexers/installers.py
+++ b/pygments/lexers/installers.py
@@ -267,6 +267,9 @@ class DebianControlLexer(RegexLexer):
     """
     Lexer for Debian ``control`` files and ``apt-cache show <pkg>`` outputs.
 
+    Specification of `control`` files is available at
+    https://www.debian.org/doc/debian-policy/ch-controlfields.html
+
     .. versionadded:: 0.9
     """
     name = 'Debian Control file'
--- a/pygments/lexers/installers.py
+++ b/pygments/lexers/installers.py
@@ -280,7 +280,8 @@ class DebianControlLexer(RegexLexer):
         'root': [
             (r'^(Description)', Keyword, 'description'),
             (r'^(Maintainer)(:\s*)', bygroups(Keyword, Text), 'maintainer'),
-            (r'^((Build-)?Depends)', Keyword, 'depends'),
+            (r'^((Build-|Pre-)?Depends(-Indep|-Arch)?)', Keyword, 'depends'),
+            (r'^(Recommends|Suggests|Enhances)', Keyword, 'depends'),
             (r'^((?:Python-)?Version)(:\s*)(\S+)$',
              bygroups(Keyword, Text, Number)),
             (r'^((?:Installed-)?Size)(:\s*)(\S+)$',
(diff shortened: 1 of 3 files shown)


birkenfeld:
Unrelated to your change, but: in the test output I notice that field values (i.e. the part after `:`) are processed character by character.  This is not really efficient.  Do you want to have a go at fixing the regexes?

sblondon:
If I understand correctly, you would like such changes in `tests/examplefiles/control/control.output`:

From that
```
'Maintainer'  Keyword
': '          Text
'D'           Text
'e'           Text
'b'           Text
'i'           Text
'a'           Text
'n'           Text
' '           Text
'P'           Text
'y'           Text
't'           Text
'h'           Text
'o'           Text
'n'           Text
' '           Text
'T'           Text
'e'           Text
'a'           Text
'm'           Text
' '           Text
'<team+python@domain.tld>' Generic.Strong
```

To something like:
```
'Maintainer'  Keyword
': Debian Python Team'          Text
'<team+python@domain.tld>' Generic.Strong
```

If it's what you ask, I will  do it.
If not, could you give me an hint about your need?

birkenfeld:
Yes, exactly! Basically in the substates `'.'` like rules should be replaced by one that matches as many non-special characters as possible.

sblondon:
@birkenfeld I done it by hand (https://github.com/pygments/pygments/pull/1946/commits/a4adf637cc6d2c22ce2af14810c88ca3ec1084d4) but now the test fails because the output is not as expected.
Running `pytest tests/examplefiles --update-goldens` overwrites the changes done to `control.output`.

Do you know what I missed?

birkenfeld:
Ah - I think there was a slight misunderstanding. The regexes in the lexer need to be adapted, then the output changes automatically on update-goldens :)

If you don't want to worry about that, it's fine - I'll merge this (in its original state) and fix the regexes when I get to it.

sblondon:
I think I understand it better: I should modify the regexes like `(r'.', Text),` [here](https://github.com/pygments/pygments/blob/643a356c535efd35753614b843e9562c3a73a135/pygments/lexers/installers.py#L294). In such case, the regex should accepts every characters but `<` and `,` (because they need to be matched by previous lines).
The line should become something like:
```python
(r'[^,<]*', Text),
```

(I did not check the regex inside the file so it probably need some fixes.)
Do you confirm I am on the right path?

sblondon:
I removed the handmade changes on `control.output` and added the `Uploaders` field.

Then, I added a new commit to improve the regexes (see https://github.com/pygments/pygments/pull/1946/commits/329e273c33248806f7a87c00041cd894e0a3671a). The `control.output` is more compact after this change.

The `depends` block needs to get same changes. However, it's not obvious how to do it so I will not do it (because I think there are more risk I break it than I improve it).

birkenfeld:
Thanks a lot! I'll take it from here then.

PR #1947, opened 2021-11-07
Add kshrc to the shell lexer

eigilhs:


Anteru:
Merged, thanks a lot!

PR #1948, opened 2021-11-08
GDScript: Add void datatype

Morphclue:
# Context
The Lexer is missing the `void`-type. This type isn't listed in the [Grammar Docs](https://docs.godotengine.org/en/stable/development/file_formats/gdscript_grammar.html). The reason for that could be that the grammar is incosistent with GDScript (see note in Grammar Docs).

The type `void` should be added, because you can specify it as a return type in `GDScript`. 
Example from the [GDScript Basics](https://docs.godotengine.org/en/stable/getting_started/scripting/gdscript/gdscript_basics.html):
```python
func void_function() -> void:
    return # Can't return a value
```

## References
* [Grammar Docs](https://docs.godotengine.org/en/stable/development/file_formats/gdscript_grammar.html)
* [GDScript Basics](https://docs.godotengine.org/en/stable/getting_started/scripting/gdscript/gdscript_basics.html)


Anteru:
Merged, I'll update the tests manually.

PR #1955, opened 2021-11-14
Fix unreadable dark styles by adding missing colors

not-my-profile:
The first commit fixes some completely unreadable styles (due to missing colors):

Before:
image

After:
image

(same for stata-dark & rrt)

The second commit slightly changes the color of numbers for the sas style so that it meets WCAG AA. I'd post a screen but the change is so small that I don't think you can actually see any difference.

Ideally merge these without squashing.


Anteru:
Merged, thanks a lot!

PR #1956, opened 2021-11-14
Sort styles in gallery by background luminance

not-my-profile:
**Before**:

image

**After**:

image


Anteru:
Merged, thanks a lot!

PR #1957, opened 2021-11-14
Whitespace token modernization - [ef] lexers - regarding #1905

blu-base:
This PR is chunk of the effort (#1905) to insert the Whitespace token where ever it applies.

The lexer factor seems very inelegant from what i have seen...
I also skipped the foxpro lexer. That lexer needs much more work or more patience than I had today =)


Anteru:
Merged, once again, thanks a lot!

PR #1958, opened 2021-11-14
demo: Sort options in lexer select

not-my-profile:
Previously CSS+Django/Jinja, CSS+Ruby, and CSS+Genshi Text were all
before CSS because pygments.lexers.LEXERS contains the lexers ordered
alphabetically by their class name and CssDjangoLexer, CssErbLexer and
CssGenshiLexer come before CssLexer.

If users are however typing "CSS" while focusing the language select in
the demo, chances are that they actually want the base CSS lexer. The
same applies to HTML, JavaScript, XML, etc.


birkenfeld:
Thanks!

PR #1959, opened 2021-11-15
Improve CddlLexer performance

blu-base:
The `CddlLexer` always seemed incredibly slow for the tiny provided example file. Profiling showed that the re engine took most of the time. After some testing the `_re_id` expression proved to be the culprit.

An other minor change improved grouping of the `String.Single` Token. That's the only reason the examplefile output needs an update.

---- 

If you want, I can remove the modification for the `String.Single` token to only keep the important change to the `_re_id` expression

### DIFF
--- a/pygments/lexers/cddl.py
+++ b/pygments/lexers/cddl.py
@@ -104,7 +104,8 @@ class CddlLexer(RegexLexer):
 
     _re_id = (
         r"[$@A-Z_a-z]"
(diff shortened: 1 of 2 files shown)


blu-base:
The failed check seems to be a fluke?

birkenfeld:
Looks good to me (and I like the cleanup for the string case), thanks!

PR #1960, opened 2021-11-16
Add support for loading TrueType font files.

MarronEyes:
In the ImageFormatter class, we can load ttf fonts that already exists in the operating system.
However, the "font_name" argument does not accept a custom font (a filename or a file-like object containing a ttf font).
I hope that this pull request solves this problem.


birkenfeld:
I think this is a good feature to have, only two questions:

* how to handle different font styles? At the moment this is impossible.
* I think it's more idiomatic to check for `hasattr(obj, 'read')` to check for file-like objects, no?

MarronEyes:
- We can maybe ask the user to provide different styles of the font only if he wants to use a custom font.
- You are right. I will change it.

birkenfeld:
Thanks! Could you please update the documentation of the parameter (in the docstring of `ImageFormatter`)?

MarronEyes:
Yes, of course!
I also noticed that the "FreeTypeFont" font object can take multiple styles if the specified font is a variation font. (See https://pillow.readthedocs.io/en/stable/reference/ImageFont.html#PIL.ImageFont.FreeTypeFont.get_variation_names)
So we could associate each style (in the "fonts" dict) with its font object ?

MarronEyes:
I tested it and cannot instantiate new font objects with different styles because it throws an OSError.
The only solution I think is to use the "set_variation_by_name" method (https://pillow.readthedocs.io/en/stable/reference/ImageFont.html#PIL.ImageFont.FreeTypeFont.set_variation_by_name) to switch the style of the font, but it requires changing the code of the ImageFormatter class.

MarronEyes:
Sorry for the long inactivity, I wasn't available at the moment.
I added a get_style function to get the specified style of the font in get_font.
The code is tested and it works! The pr can be merged.

Anteru:
This also looks good to me, but there's no test for it (nor do I see how this can be tested easily.) @birkenfeld , @Jean-Abou-Samra , any concerns with that? I'm not super happy to merge code without tests but I don't see an easy way to test image based output.

pygments/formatters/img.py
```diff
@@ -219,14 +227,42 @@ def get_font(self, bold, oblique):
...
+        If not found, return the normal font.
+        """
+        font = self.fonts[style]
+        for style_name in STYLES[style]:
+            try:
+                font.set_variation_by_name(style_name)
+                return font
+            except ValueError:
+                pass
+            except OSError:
```
Jean-Abou-Samra, line 262: This indicates an invalid font, right? Shouldn't there be a warning or an error?

pygments/formatters/img.py
```diff
@@ -219,14 +227,42 @@ def get_font(self, bold, oblique):
...
+        Get the specified style of the font if it is a variable font.
+        If not found, return the normal font.
+        """
+        font = self.fonts[style]
+        for style_name in STYLES[style]:
+            try:
+                font.set_variation_by_name(style_name)
+                return font
+            except ValueError:
+                pass
```
Jean-Abou-Samra, line 261: What is this case for? I don't see `ValueError` mentioned at
https://pillow.readthedocs.io/en/stable/reference/ImageFont.html#PIL.ImageFont.FreeTypeFont.set_variation_by_name

pygments/formatters/img.py
```diff
@@ -254,6 +290,8 @@ class ImageFormatter(Formatter):
         The font name to be used as the base font from which others, such as
         bold and italic fonts will be generated.  This really should be a
         monospace font to look sane.
+        If a filename or a file-like object is specified, the user must
+        provide different styles of the font.
```
Jean-Abou-Samra, line 294: If I understand correctly, a typo in the file name will result in `font_name` being understood as the name of a font and not a path, and the errors will read 'No usable fonts named ...' or something like that depending on the system. Can we make it explicit rather than implicit and use a separate `font_path` parameter?

Jean-Abou-Samra:
> I'm not super happy to merge code without tests but I don't see an easy way to test image based output.

No, I don't see one either, especially that font stuff is so platform-dependent :( So I think it's OK if this doesn't have tests.

pygments/formatters/img.py
Anteru: Does it guarantee that no `ValueError` is raised? I'd guess some of the code could do that ... alternative would be to catch only `OSError` and pass in all other cases?

Anteru:
Merged for 2.16. Let's see if this works in the wild, thanks for the contribution!

Issue #1895, opened 2021-09-12
nasm: the ch of check highlights specially


Akuli:
![asm](https://user-images.githubusercontent.com/18505570/132998342-5bc9f49a-a697-492c-a32f-07afacacd9d5.png)

I'm new to assembly, but I'm pretty sure that the `ch` of `check_collision_main` is not supposed to be a separate token.

birkenfeld:
fixed by #1961

PR #1961, opened 2021-11-16
Asm lex bugfix #1895: register re check for boundary 

blu-base:
There were a few expressions left which did not check for word boundaries, which led to #1895.

This PR fixes these minor issues, and adds a snippet as shown in that issue.


birkenfeld:
Nice, thanks!

PR #1963, opened 2021-11-17
Update Alloy DSL for Alloy 6

hwayne:
Last Tuesday [Alloytools](http://alloytools.org/) released a new, [backwards-incompatible version](https://alloytools.org/alloy6.html). The biggest change is that `'` is now an operator and no longer a valid identifier token. The team is encouraging people to use `"` instead, which the current Pygments lexer assumes automatically starts a string.

This change adds the new keywords, modifies how the lexer handles `"`, and closes an edge case in using strings for `fact` definitions. See commit messages for further details on each.


hwayne:
Ooop, looks like I didn't notice the Alloy tests, will fix

hwayne:
Running `regexlint` locally doesn't seem to trigger an error. Is there a specific way I should be running it?

Anteru:
@birkenfeld Do you have an idea what might be going on here?

Jean-Abou-Samra:
I get the error locally with an up-to-date checkout of `regexlint`:

```
/home/jean/repos/pygments/pygments/lexers/dsls.py:641: (AlloyLexer:root:pat#17) E105: Potential out of order alternation between '\\.' and '\\.\\.'
              (r'!|#|&&|\+\+|<<|>>|>=|<=>|<=|\.|\.\.|->', Operator),
                                                ^ here
```

It seems sensible because this regex will never match `..` as the first `.` already matches `\.`.

hwayne:
Only a year late, but I fixed the regexlint issue, let's see if this passes the check!

birkenfeld:
Tests are fine. Since there were tons of unrelated changes and conflicts in the branch, I manually cherry-picked the changes and pushed them in f1283e65e34483658a89bc48a885f840ce6c8a34.

PR #1965, opened 2021-11-20
Add lexer for typeql

sitch:
Add lexer for typeql based on https://github.com/typedb-osi/typeql-plugin-vscode/blob/master/syntaxes/tql.tmLanguage.json


Anteru:
Thanks for preparing the PR. This is missing test files and has a few other issues.

PR #1967, opened 2021-11-21
Improve Brainfuck comment detection

Mango0x45:
In brainfuck, if you reach a loop using the '[' and ']' characters and
the current cell has a value of 0, the loop is skipped. This is
sometimes used as a trick to allow for comments containing punctuation
that conflicts with brainfuck commands. For example, you can put the
following at the very beginning of your brainfuck code:

[
	This is my comment! I can use all sorts of characters like , and
	. which I would not be able to use normally.
]

You can also set the value of the current memory cell before the loop to
0 to force the loop to be skipped. This can be done with the [-] or [+]
loops:

[-][This is another comment, and it has a period and comma.]

### DIFF
--- a/pygments/lexers/esoteric.py
+++ b/pygments/lexers/esoteric.py
@@ -8,7 +8,7 @@
     :license: BSD, see LICENSE for details.
 """
 
-from pygments.lexer import RegexLexer, include, words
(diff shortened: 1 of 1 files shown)


Anteru:
Thanks for preparing this! Could you please add an example snippet so this is actually tested?

PR #1973, opened 2021-11-26
Add links to lines with css-line numbers

dschwoerer:
I wanted to link to the lines, and was disappointed that I could not click on the numbers.

Only later I realised, I could by hand create the links. This patch allows to have links, and thus makes the references easily discoverable.

Resolves https://github.com/Fortran-FOSS-Programmers/ford/issues/369

### DIFF
--- a/pygments/formatters/html.py
+++ b/pygments/formatters/html.py
@@ -765,7 +765,8 @@ class HtmlFormatter(Formatter):
         for t, line in inner:
             if t:
                 i += 1
-                yield 1, '<a id="%s-%d" name="%s-%d"></a>' % (s, i, s, i) + line
+                href = "" if self.linenos  else ' href="%s-%d"' % (s, i)
+                yield 1, '<a id="%s-%d" name="%s-%d"%s></a>' % (s, i, s, i, href) + line
             else:
                 yield 0, line
 


Anteru:
Merged, after a lot of wrestling with it...

Anteru:
@dschwoerer How are you using this feature? There's a (valid) concern that the link itself is invisible, so there is no click area, meaning without CSS styling the anchor somehow this feature is useless. I'd like to understand how you're handling this and see if we can find a better solution.

dschwoerer:
See Fortran-FOSS-Programmers/ford#369

Unfortunately I don't have a full working example, maybe @ZedThree could provide one?

ZedThree:
We're not using this in production yet, but here's a MWE (pygments CSS file not included):


```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <link href="pygments.css" rel="stylesheet">

    <style>
      .hl pre {
          counter-reset: line-numbering;
      }

      .hl pre a::before {
          content: counter(line-numbering);
          counter-increment: line-numbering;
          margin-top: 4.5em;
          width: 20px;
          opacity: 0.7;
          display: inline-block;
          color: #aaa;
          background: #eee;
          border-right: 1px solid #ccc;
      }

      .hl pre a {
          display: inline-block;
          height: 4.5em;
          margin: -4.5em 0 0;
      }

      .hl pre a:hover::before {
          color: #888;
          background: #ccc;
      }
    </style>
  </head>

  <body>
    <div class="hl"><pre>
        <a id="ln-1" name="ln-1" href="#ln-1"></a> <span class="k">subroutine </span><span class="n">printSquare</span><span class="p">(</span><span class="n">x</span><span class="p">)</span>
        <a id="ln-2" name="ln-2" href="#ln-2"></a>   <span class="kt">integer</span><span class="p">,</span> <span class="k">intent</span><span class="p">(</span><span class="n">in</span><span class="p">)</span> <span class="kd">::</span> <span class="n">x</span>
        <a id="ln-3" name="ln-3" href="#ln-3"></a>   <span class="k">write</span><span class="p">(</span><span class="o">*</span><span class="p">,</span><span class="o">*</span><span class="p">)</span> <span class="n">x</span><span class="o">*</span><span class="n">x</span>
        <a id="ln-4" name="ln-4" href="#ln-4"></a> <span class="k">end subroutine </span><span class="n">printSquare</span>
    </pre></div>
  </body>
</html>
```

I've stripped down our existing CSS, and added styling on hover:

```css
      .hl pre a:hover::before {
          color: #888;
          background: #ccc;
      }
```

We use `HtmlFormatter(lineanchors="ln", cssclass="hl")` to format this table.

The result looks something like:

image

---

btw, I think the `href` in this PR is missing the initial `#` to correctly link to the anchor

Anteru:
Thanks a lot! The # has been fixed already (that was an oversight ...). I see you're using extra CSS to add the line numbers, I guess that might be worth documenting as the recommended CSS for this feature, because by default those links are invisible/unclickable.

ZedThree:
Yes, probably worth documenting, as it's a non-trivial amount of CSS needed

PR #1978, opened 2021-12-03
Whitespace token modernization - [gh] lexers w/o html - regarding #1905

blu-base:
This PR is chunk of the effort (#1905) to insert the Whitespace token where ever it applies.

From this chunk, i excluded the html lexer, which i'm going to prepare separately.
The html lexers affects a hugh number of tests...


Anteru:
Merged, thanks a lot! I made a minor change to the `Cypher` lexer to use `words`, hope you're ok if that gets folded into this PR.

blu-base:
> Merged, thanks a lot! I made a minor change to the `Cypher` lexer to use `words`, hope you're ok if that gets folded into this PR.

Yes sure! I just tried to maintain the original intend. But this resulted in that cumbersome section.

PR #1979, opened 2021-12-04
Refine the One Dark theme a bit for consistency with other popular projects (Atom, Vim, Alacritty, etc) using the same theme

ranelpadon:
@Anteru As per my comment in https://github.com/pygments/pygments/pull/1924#issuecomment-985987042. Just want to emphasize that the exact values are more reliable metric than the "contrast" thing since the latter is kinda subjective. Other users that don't like the contrast of One Dark could just use the other themes. But for fans of One Dark, I think they will value consistency.

Actually, it's not much different. Here's the current Pygment's version (just ignore the IPython prompt's/left side color which are custom, non-One Dark color):

Color - One Dark - With Custom Prompt Color and Characters__

and here's the expected version:

Color - One Dark - With Custom Prompt Color and Characters

Non-fans will not notice the subtle difference. And I noticed only since I'm using Alacritty with One Dark theme also (i.e. there are 2 flavors of "orange" color in the first photo above). Thanks :)

### DIFF
--- a/pygments/styles/onedark.py
+++ b/pygments/styles/onedark.py
@@ -12,14 +12,14 @@
 """
 
 from pygments.style import Style
-from pygments.token import Token, Punctuation, Keyword, Name, Comment, String, \
-     Number, Operator, Generic, Whitespace
+from pygments.token import (Comment, Generic, Keyword, Name, Number, Operator,
+                            Punctuation, String, Token, Whitespace)
 
 
 class OneDarkStyle(Style):
     """
     Theme inspired by One Dark Pro for Atom
-    
+
     .. versionadded:: 2.11
     """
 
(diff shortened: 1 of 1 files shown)


ranelpadon:
PS: I wrote also an article earlier featuring IPython/Pygments, and took a lot of screenshots, hence I noticed also the subtle differences:
https://medium.com/codex/pimp-your-python-and-django-shells-db9c995c3735

You might want to check it out also. :)

Anteru:
Merged, thanks. There's a warning that One Dark is not accessible and that bit doesn't change, so we might as well use the "exact" colors.

ranelpadon:
@Anteru Not really sure what you meant here:
> There's a warning that One Dark is not accessible and that bit doesn't change, so we might as well use the "exact" colors.

Although my PR evens out the intensity of some colors. Otherwise, some colors are too strong relative to others which cause imbalanced for the overall look-and-feel (as you could see in my screenshots above). There's a reason why their colors are "exact" so that all colors have "equal" perceived intensity. And thanks a lot for merging it!

Actually, we incorporated my PR also in the cool [Pastery](https://www.pastery.net/) site (implemented in Django). Turned out that Pastery is using Pygments too to highlight the rendered/embedabble code snippets. Pygments color scheme is being transpiled to CSS. I filed an issue to add the One Dark theme which was resolved few days ago :
https://gitlab.com/stavros/pastery/-/issues/14

So, now One Dark is available (here's my sample):

Screen Shot 2021-12-28 at 10 45 44 PM

Just thought that you might be interested to know as well. :)

PR #1985, opened 2021-12-10
Add Chinese documentation support.

xinetzone:


birkenfeld:
Thanks for the contribution. At the moment we don't have the resources to maintain the docs in different languages.

If you maintain and host them yourself, we'll be happy to add a link from the main docs though.

Issue #1353, opened 2020-01-09
HashiCorp Configuration Language (HCL) Support


karras:
The HashiCorp Configuration Language (HCL) is becoming very popular these days and also very handy to have in other projects such as MediaWiki, which make use of pygments.

karras:
Specification: https://github.com/hashicorp/hcl

@hairmare ^

appleshampooid:
pygments does support HCL but you have to label it as `terraform` which isn't 100% accurate, but seems to work pretty well for me.

karras:
Good catch, totally mist that one. Thanks!

lgiordani:
Can you please have a look at #1756? Feedback is welcome. Thanks!

lorengordon:
@lgiordani I have a block that looks like this, which is failing when using pygments 2.9.0 with `WARNING: Could not lex literal_block as "terraform". Highlighting skipped.` Strange thing is this syntax is valid with both HCL 1 and HCL 2, and it was working fine with pygments 2.8.1.

```
```terraform
provider "aws" {}

module "test-lx-instance" {
  source = "git::https://github.com/plus3it/terraform-aws-watchmaker//modules/lx-instance/"

  Name      = "tf-watchmaker-lx-autoscale"
  AmiId     = "__AMIID__"
  AmiDistro = "__AMIDISTRO__"
}
```

lgiordani:
@lorengordon I'll have a look as soon as possible, thanks for the test case

Anteru:
Can someone confirm this fixed by #1756? I see an error on the first `}` but that seems unrelated?

Jeinhaus:
I am using [terragrunt](https://terragrunt.gruntwork.io/) that also uses HCL for configuration. The changes from #1756 do not fix that. Should I open another issue for this (even though it's using HCL)?

An example for terragrunt's configuration is below:
```
terraform {
  source =
    "github.com:foo/infrastructure-modules.git//app?ref=v0.0.1"
}
inputs = {
  instance_count = 10
  instance_type  = "m4.large"
}
```

Anteru:
Given #1756 doesn't fix your issue, but is already merged, I'd suggest to open a new one. For my understanding: Is terragrunt using a dialect of HCL or is that a problem for "plain" HCL?

lorengordon:
[...]

PR #1987, opened 2021-12-10
Fix terraform blockname regex

dwc0011:
The regex for blockname does not allow empty provider blocks.
For example as mentioned by @lorengordon in #1353 the code snippet he provided had an empty provider block which is valid teraform. @Anteru also pointed out the issue with the first "}", at the time thought to be unrelated, however that is the reason why the snippet provided by @lorengordon was receiving the error.

The following valid terraform does not process correctly.
`provider "aws" {}`

I have updated the blockname regex to allow `{}|{`  this fixes the issue and passes all tests.  I have added a test to the snippets and ran the golden updates command on it.
`python -m pytest --update-goldens tests/snippets/terraform/test_types.txt`
I then ran `make test` and all tests passed.

Admittedly, this is an oversimplified fix and I am working on stronger regex to deal with blocknames that are not allowed to be empty.  But to get the ball rolling and ensure the tests are correct and heading down the right path I wanted to get this PR started.  

If there is any reason to put more complex or fuller terraform tests under examples, for this I just followed the instructions under contributing and added a snippets test, but am wondering terraform is at the point of needing fuller more complex test case.

Please let me know if you want me to create a separate issue or use #1353 as the issue this PR addresses.


birkenfeld:
Strange, I don't understand how the lexer leaves the `blockname` state without producing an error...

dwc0011:
> Strange, I don't understand how the lexer leaves the `blockname` state without producing an error...

Can you elaborate?  Are you saying with the fix in the regex, or before the fix in the regex?

birkenfeld:
For both cases, the `blockname` state has no "exit" condition.

I now re-read the code, and the lexer has an implicit "return to root state" if at line end and no rules in a state matched.  This is not quite clean though, and I wonder if it should be removed.

In any case, please add a `default('#pop')` tot he state.

dwc0011:
> For both cases, the `blockname` state has no "exit" condition.
> 
> I now re-read the code, and the lexer has an implicit "return to root state" if at line end and no rules in a state matched. This is not quite clean though, and I wonder if it should be removed.
> 
> In any case, please add a `default('#pop')` tot he state.

Ok, I didn't look at whether the lexer was correct, once I noticed the regex I adjusted that and then ran the tests.  I will look at it closer and add the default('#pop')  and test again.

lorengordon:
@birkenfeld Anything else needed here?

pygments/lexers/configs.py
```diff
@@ -726,8 +726,9 @@ def heredoc_callback(self, match, ctx):
         'blockname': [
             # e.g. resource "aws_security_group" "allow_tls" {
             # e.g. backend "consul" {
-            (r'(\s*)("[0-9a-zA-Z-_]+")?(\s*)("[0-9a-zA-Z-_]+")(\s+)(\{)',
+            (r'(\s*)("[0-9a-zA-Z-_]+")?(\s*)("[0-9a-zA-Z-_]+")(\s+)(\{\}|\{)',
```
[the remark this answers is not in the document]
dwc0011: Removed (\s+)(\{) fom the regex

pygments/lexers/configs.py
```diff
@@ -726,8 +726,28 @@ def heredoc_callback(self, match, ctx):
...
+                    Whitespace,
+                    Punctuation,
+                ),
+            ),
+            default("#pop"),
+=======
+            (r'(\s*)("[0-9a-zA-Z-_]+")?(\s*)("[0-9a-zA-Z-_]+")',
+	    bygroups(Whitespace, Name.Class, Whitespace, Name.Variable, Whitespace, Punctuation)),
+            default('#pop'),
+>>>>>>> Stashed changes
```
Jean-Abou-Samra, line 750: You have unresolved merge conflicts.

Jean-Abou-Samra:
Unfortunately, this PR still has merge conflicts with the master branch.

dwc0011:
> Unfortunately, this PR still has merge conflicts with the master branch.

I assumed we would deal with merge issue with master once everything was reviewed and agreed it is good.

I will address now.

Jean-Abou-Samra:
> I assumed we would deal with merge issue with master once everything was reviewed and agreed it is good.

Well, it happens that updates to the merge target completely change the game as to the way a change should be done. Plus, for such a small change, it shouldn't be a burden to rebase.

pygments/lexers/configs.py
```diff
@@ -729,6 +729,13 @@ def heredoc_callback(self, match, ctx):
                 heredoc_callback,
             )
         ],
+        'blockname': [
+            # e.g. resource "aws_security_group" "allow_tls" {
+            # e.g. backend "consul" {
+            (r'(\s*)("[0-9a-zA-Z-_]+")?(\s*)("[0-9a-zA-Z-_]+")',
+             bygroups(Whitespace, Name.Class, Whitespace, Name.Variable, Whitespace, Punctuation)),
```
Jean-Abou-Samra, line 736: Fails with `regexlint` in the CI, it complains that there are only 4 groups for 6 token types in bygroups(). I tend to agree :-)

dwc0011:
> I was surprised to see a fresh and unused state in the diff -- apparently the same issue has already been fixed by #2097 in the meantime?

I am glad theirs got merged, this one has been open since December and never was addressed until recently, such is the life of open source projects.  We will test the latest build and see if it fixes the issues we were addressed with the other fix.

lorengordon:
@Jean-Abou-Samra Any idea when the 2.12 release will be, so we can see if #2097 does indeed fix this issue?

Jean-Abou-Samra:
@lorengordon You'd have to ask @Anteru. However, there is no need to wait on the release. Just do

```
pip install --user https://github.com/pygments/pygments/archive/master.zip
```

and (provided I've got the command right) that should give you an installation with current master. For your information, this is the token output I get for the example from https://github.com/pygments/pygments/issues/1353#issuecomment-862414618:

```
Token.Keyword.Reserved	'provider'
Token.Text.Whitespace	' '
Token.Name.Variable	'"aws"'
Token.Text.Whitespace	' '
Token.Punctuation	'{'
Token.Punctuation	'}'
Token.Text.Whitespace	'\n'
Token.Text.Whitespace	'\n'
Token.Keyword.Reserved	'module'
Token.Text.Whitespace	' '
Token.Name.Variable	'"test-lx-instance"'
Token.Text.Whitespace	' '
Token.Punctuation	'{'
Token.Text.Whitespace	'\n'
Token.Text.Whitespace	'  '
Token.Name.Attribute	'source'
Token.Text.Whitespace	' '
Token.Operator	'='
Token.Text.Whitespace	' '
Token.Literal.String.Double	'"git::https://github.com/plus3it/terraform-aws-watchmaker//modules/lx-instance/"'
Token.Text.Whitespace	'\n'
Token.Text.Whitespace	'\n'
Token.Text.Whitespace	'  '
Token.Name.Attribute	'Name'
Token.Text.Whitespace	'      '
Token.Operator	'='
Token.Text.Whitespace	' '
Token.Literal.String.Double	'"tf-watchmaker-lx-autoscale"'
Token.Text.Whitespace	'\n'
Token.Text.Whitespace	'  '
Token.Name.Attribute	'AmiId'
Token.Text.Whitespace	'     '
Token.Operator	'='
Token.Text.Whitespace	' '
Token.Literal.String.Double	'"__AMIID__"'
Token.Text.Whitespace	'\n'
Token.Text.Whitespace	'  '
Token.Name.Attribute	'AmiDistro'
Token.Text.Whitespace	' '
Token.Operator	'='
Token.Text.Whitespace	' '
Token.Literal.String.Double	'"__AMIDISTRO__"'
Token.Text.Whitespace	'\n'
Token.Punctuation	'}'
Token.Text.Whitespace	'\n'
```

lorengordon:
Appreciate that, thank you!

Jean-Abou-Samra:
You're welcome. Do you confirm that the bug went away?

PR #1994, opened 2021-12-20
Python: lex soft keywords

### LINKED ISSUE #1797: support new Python "soft keywords": match, case and _
Python 3.10 introduces pattern-matching blocks, in which the `match`, `case` and `_` "soft keywords" are considered keywords only in specific parts of the syntax. See [the docs](https://docs.python.org/3.10/reference/compound_stmts.html#the-match-statement) for more details.

However, the list of Python keywords in `lexers/python.py` doesn't include these, nor does the Python lexer appear to have any other way of supporting these keywords.

amitkummer:
Some notes:

- This approach is very similar to what @taleinat did in python/cpython#25851 (thank you! this was a huge help).
  It is not perfect, but it's rather simple and I can't think of an edge case.

- I did not use the `words` function to create the regex matching the
  keywords list, because it returns a capturing group (`()`) and it
  needs to be non-capturing here (because of `bygroups` usage).

- I chose to go to the 'soft-keywords-inner' state after both
  `match` and `case`, even though it's unnecessary for `match`
  (the inner state catches the `_` wildcard keyword which appears
  only after a `case`).

  This is mostly harmless and saves us from writing the 'soft-keywords'
  regex twice each for `match` and `case` with the extra inner state
  just for `case`.

  The only piece of code this will lex incorrectly is `match _:`
  (`_` will be lexed as keyword). I doubt though that pattern matching
  will be used like this.

Fixes #1797.

### DIFF
--- a/pygments/lexers/python.py
+++ b/pygments/lexers/python.py
@@ -9,9 +9,10 @@
 """
 
 import re
+import keyword
 
 from pygments.lexer import Lexer, RegexLexer, include, bygroups, using, \
-    default, words, combined, do_insertions
+    default, words, combined, do_insertions, this
 from pygments.util import get_bool_opt, shebang_matches
 from pygments.token import Text, Comment, Operator, Keyword, Name, String, \
     Number, Punctuation, Generic, Other, Error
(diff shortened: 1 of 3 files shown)


Anteru:
Merged, thanks a lot!

PR #1998, opened 2021-12-23
Linkify GitHub issue/PR numbers in changelog

not-my-profile:
Stops with 2.4.2 because a different tracker was used before.

image


Anteru:
Thanks, that's a very nice way to handle this. Much appreciated!

PR #1999, opened 2021-12-26
Make the demo awesome

not-my-profile:
See my commit messages for the changes I made.

The demo should now be way more fun and useful (not just for users but also for developers).

(please don't squash my commits :)


Anteru:
That's quite a massive change, but I do like the direction. Thanks for putting this together!

PR #2001, opened 2021-12-28
Some LilyPond updates

Jean-Abou-Samra:
- Add pitch language names to builtins.  They are not highlighted
  specially in the default style, but can be in a custom style.

- Refactor matching of builtins in order to avoid confusion between
  dim chord modifier and \dim dynamic command.

- Support grob subproperties.

- Add segno repeat type.

- While at it, update builtins for latest version.


Jean-Abou-Samra:
The failure on `rita.py` is unrelated and fixed by #2003.

PR #2002, opened 2021-12-28
Fix single and multiline comment regex for GSQL

gregoryg:
Prior to this change, multiline comments would get highlighted as GSQL code

Test:
```
   /* I use the GRAPH to add a vertex
       now let's create the graph
   */
```

Likewise, a comment added to the end of a GSQL line would cause the entire line to be interpreted as a comment
Test:
```
USE GRAPH Patents # switch to our new graph
```

Note: this commit also removed trailing spaces from several other lines in `gsql.py`

### DIFF
--- a/pygments/lexers/gsql.py
+++ b/pygments/lexers/gsql.py
@@ -43,29 +43,29 @@ class GSQLLexer(RegexLexer):
             include('operators'),
         ],
         'comment': [
-            (r'.*\#.*\n', Comment.Single),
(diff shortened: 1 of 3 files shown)


Anteru:
Thanks! Would you mind adding this to the example file for GSQL so we can regression test it?

gregoryg:
> 

I'll do that - I'll add just a few lines to the existing example file

Anteru:
Perfect, thanks a lot!

Anteru:
Merged, thanks!

PR #2003, opened 2021-12-28
Fix rita.py header

Jean-Abou-Samra:


Anteru:
Thanks, was going to fix that manually but you've been faster :)

PR #2004, opened 2021-12-28
Update for new Python traceback locations

Jean-Abou-Samra:
Since https://github.com/python/cpython/pull/27037, they can include
tildes in addition to the carets.


Anteru:
Merged, thanks a lot!

Issue #1950, opened 2021-11-12
Is any benchmark/accurate for guess_lexer_for_filename available?


Zackhardtoname:
I have been wondering about the performance difference between Pygments and alternatives like [go-enry](https://github.com/go-enry/go-enry).

I am mainly concerned about the accuracy the [guess_lexer_for_filename](https://github.com/pygments/pygments/blob/3ef93d89ebb04373d2887b6be45b482c0887d9d2/pygments/lexers/__init__.py#L240) function in Pygments in identifying the correct programming languages by filename and file content. Is there any benchmark or general experiences from usage available? 

I also notice that there's not a build-in way to customize the matching rules so advanced users would need to write some custom code, which is fine.

Zackhardtoname:
[guesslang](https://guesslang.readthedocs.io/en/latest/contents.html#how-does-guesslang-guess) has a nice benchmark and open sourced [their benchmark tool](https://github.com/yoeo/guesslangtools).

Anteru:
No, there is no benchmark. If you go by filename it should be pretty accurate as there aren't that many languages colliding on the same ending, but "general" guessing based on the contents is not benchmarked. I think this is something we need to improve eventually, but it's a monumental undertaking and nobody volunteered yet to tackle this.

Zackhardtoname:
Actually I did find [a nice human-labeled dataset](https://github.com/smola/language-dataset) giving us a benchmark comparing it with other solutions like Linguist and go-enry.
The end result is that according to the human labels, **Pygments is right 76.18%**, Linguist is right 91.50% of the time. 2% of linguist’s error could be corrected by pygments

Sample code I used for getting the percentages:
```python
import os

import yaml
from collections import Counter, defaultdict

with open("language-dataset/data/dataset.yml", "r") as stream:
    try:
        res = yaml.safe_load(stream)
    except yaml.YAMLError as exc:
        print(exc)

num_total = 0
inconsistent = 0
allkey = {key for dictio in res["files"].values() for key in dictio["annotations"]}
all_linguist_keys = {key for key in allkey if key.startswith("linguist")}
all_diff_cases = defaultdict(list)
all_diff_lang = Counter()

for name, repo in res["files"].items():
    annotations = repo["annotations"]
    if "linguist" in annotations and "pygments-filename" in annotations:
        ground_truth = annotations.get("human-smola", annotations.get("human-ajnavarro"))
        if ground_truth and ground_truth != "Unknown":
            num_total += 1
[...]

Issue #1702, opened 2021-02-06
Refactor/Rewrite analyze_text


Anteru:
`analyze_text` is a source of problems and has an impossible task to do -- it's supposed to _locally_ decide how well a lexer matches a given piece of code, which is then used to make a global decision among all lexers. This is leading to tons of issues, and we have a large number out outstanding PRs related to it:

* #1610 
* #1593 
* #1607 
* #1619 
* #1692

There are two "right" solutions to this, and they can be implemented incrementally. The first one is to default `analyze_text` to only invoke lexers which support the given file extension. This will dramatically improve the accuracy, as there will be only 2-3 lexers that need to get coordinated instead of having any lexer participate (and some lexers do accept nearly everything.) This can be implemented right now, but it requires touching every single lexer and clean up the weights relative to other candidates.

The second step -- which is more involved -- is to turn to machine learning to train a model that can make global decisions by knowing all languages and being able to weight the probabilities globally. This would be the "endgame" solution for analyze_text, with the hope that we can have a reinforcement process in place to teach the model whenever an issue occurs. That's a much larger task though, similar in scope to https://github.com/github/linguist

alanhamlett:
Re-using linguist's training would be ideal for the machine learning second solution.

PR #2005, opened 2021-12-29
Better `Lexer.analyse_text`

not-my-profile:
While playing with the [new demo](https://pygments.org/demo/) I introduced in #1999 I noticed that the lexer guessing solely based on text very much leaves things to be desired. It works fine for clear cut cases like shebangs and DOCTYPES but when there isn't such a dead giveaway and you actually have to guess it's very often wrong.

`pygments.lexers.guess_lexer(text)` is currently powered by `Lexer.analyse_text(text)` which returns a float from 0 to 1 indicating how much the lexer thinks it recognizes its language (1 meaning it's certainly the right lexer and 0 meaning it's certainly not the right lexer).

IMO the clear problem with this is that most patterns can occur anywhere in the text and the position they occur in currently isn't used for lexer ranking. To illustrate this with an example:

* `#include <stdio.h>` is correctly recognized as C
* `<test></test>` is correctly recognized as XML
* The following is also correctly recognized as XML:
  ```xml
  <test>
  #include <stdio.h>
  </test>
  ```
* The following however is wrongly recognized as XML:
  ```c
  #include <stdio.h>
  
  void main(){
          printf("<test></test>");
  }
  ```
That is because the code currently looks like:

```python
class CLexer(CFamilyLexer):
    def analyse_text(text):
        if re.search(r'^\s*#include [<"]', text, re.MULTILINE):
            return 0.1

class XmlLexer(RegexLexer):
    def analyse_text(text):
        if looks_like_xml(text):
            return 0.45  # less than HTML
```

Since the `analyse_text` API doesn't propagate the positions of the matches only one of the last two examples will ever be guessed correctly.

Please note that C and XML were only examples. You can have the same problem with pretty much any pair of Pygment's >500 languages. Another example would be:

* ```python
  import sqlite3
  # ...
  cur.execute("INSERT INTO items (text) VALUES (?)", (text,))
  ```
  which should be recognized as Python, versus
* ```sql
  INSERT INTO items (text) VALUES ("import sqlite")
  ```
  which should be recognized as SQL.

So my idea is to improve the language guessing in Pygments by deprecating `Lexer.analse_text` in favor of a new method `Lexer.recognize_text` that yields the indexes of matches so that they can be factored into the scoring.

To give you two examples of how `analyse_text` would be converted to `recognize_text`:

```python
# before:
def analyse_text(text):
    if re.search(r'^\s*#include [<"]', text, re.MULTILINE):
        return 0.1
# after:
def recognize_text(text):
    yield re.search(r'^\s*#include [<"]', text, re.MULTILINE), 0.1
```

```python
# before:
def analyse_text(text):
    rv = 0
    if 'import ' in text:
        rv += 0.5
    if 'print ' in text:
        rv += 0.3
    return rv
# after:
def recognize_text(text):
    yield text.find('import '), 0.5
    yield text.find('print '), 0.3
```

Since there currently are about 150 `analyse_text` implementations, converting them all at once isn't feasible. This PR Draft demonstrates how we could introduce `recognize_text` in a backwards compatible manner, allowing us to convert the `analyse_text` methods one by one, improving lexer guessing incrementally.

What do you think?

### DIFF
--- a/pygments/lexer.py
+++ b/pygments/lexer.py
@@ -42,6 +42,8 @@ class LexerMeta(type):
     def __new__(mcs, name, bases, d):
         if 'analyse_text' in d:
             d['analyse_text'] = make_analysator(d['analyse_text'])
+        if 'recognize_text' in d:
+            d['recognize_text'] = staticmethod(d['recognize_text'])
         return type.__new__(mcs, name, bases, d)
 
 
--- a/pygments/lexers/__init__.py
+++ b/pygments/lexers/__init__.py
@@ -307,16 +307,34 @@ def guess_lexer(_text, **options):
         except ClassNotFound:
             pass
 
-    best_lexer = [0.0, None]
+    best_pos = float('inf')
+    best_score = 0.0
+    best_lexer = find_lexer_class('TextLexer')
+
     for lexer in _iter_lexerclasses():
-        rv = lexer.analyse_text(_text)
-        if rv == 1.0:
+        pos = float('inf')
+        if hasattr(lexer, 'recognize_text'):
+            score = 0
+            for match, s in lexer.recognize_text(_text):
+                if match is None or match == -1:
+                    continue
+
+                if isinstance(match, int):
+                    pos = min(pos, match)
+                else: # assuming match is re.Match
+                    pos = min(pos, match.start())
+
+                score += s
+        else:
+            score = lexer.analyse_text(_text)
+
+        if score == 1.0:
             return lexer(**options)
-        if rv > best_lexer[0]:
-            best_lexer[:] = (rv, lexer)
-    if not best_lexer[0] or best_lexer[1] is None:
-        raise ClassNotFound('no lexer matching the text found')
-    return best_lexer[1](**options)
+        if pos < best_pos or (pos == best_pos and score > best_score):
+            best_pos = pos
+            best_score = score
+            best_lexer = lexer
+    return best_lexer(**options)
 
 
 class _automodule(types.ModuleType):
--- a/pygments/lexers/algebra.py
+++ b/pygments/lexers/algebra.py
@@ -67,24 +67,18 @@ class GAPLexer(RegexLexer):
         ],
     }
 
-    def analyse_text(text):
-        score = 0.0
-
+    def recognize_text(text):
         # Declaration part
-        if re.search(
+        yield re.search(
             r"(InstallTrueMethod|Declare(Attribute|Category|Filter|Operation" +
             r"|GlobalFunction|Synonym|SynonymAttr|Property))", text
-        ):
-            score += 0.7
+        ), 0.5
 
         # Implementation part
-        if re.search(
+        yield re.search(
             r"(DeclareRepresentation|Install(GlobalFunction|Method|" +
             r"ImmediateMethod|OtherMethod)|New(Family|Type)|Objectify)", text
-        ):
-            score += 0.7
-
-        return min(score, 1.0)
+        ), 0.5
 
 
 class MathematicaLexer(RegexLexer):
--- a/pygments/lexers/c_cpp.py
+++ b/pygments/lexers/c_cpp.py
@@ -270,11 +270,9 @@ class CLexer(CFamilyLexer):
         ]
     }
 
-    def analyse_text(text):
-        if re.search(r'^\s*#include [<"]', text, re.MULTILINE):
-            return 0.1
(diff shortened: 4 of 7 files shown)


Anteru:
Yes, `analyse_text` is quite broken. I did discuss the issues here: https://github.com/pygments/pygments/issues/1702 and I think something like `guesslang` (mentioned https://github.com/pygments/pygments/issues/1950) might be a better direction by now. I'm not sure how much it's worth to "keep investing" in `analyse_text` improvements vs. "just" using `guesslang` and accepting the fact that most of the >500 languages we know about won't be reliably recognized. Maybe @birkenfeld has an opinion here?

not-my-profile:
>  I'm not sure how much it's worth to "keep investing" in analyse_text improvements

As far as I can see the `analyse_text` API hasn't been changed since 2006 when it was introduced in c9330c8a. Of course trying to get reasonable matches with an API that maps the existence of some pattern to a fixed floating point number between 0 and 1 is doomed to fail. I think factoring in match positions would lead to a major improvement of guess quality.

Yes guesslang is neat but it only supports ~50 languages, which is just 10% of the languages Pygments can highlight. While just adding guesslang would certainly improve the guessing for the popular languages, Pygments supports many more exotic languages for which getting the necessary amount of data for machine learning might be difficult. And I feel like one of the appeals of Pygments is that it's very easy to add support for another language.

Anteru:
Don't get me wrong, I do see your point, I'm just looking at the state we're in right now which is a "mostly useless" `analyse_text` implementation (judging by the number of bugs) which needs to get maintained separately from the lexer itself. There's also the problem that we're making a global guess which means stacking each lexer against each other instead of having a reasonable subset to select from (for instance, only disambiguating between all languages using `.cl` as the file ending.)

Looking a bit closer at the method you propose, I wonder if we can pick expressions stochastically from the lexer itself and use those for ranking. The main reason to not use the lexer itself and see if it produces errors is because it's too slow (is that still true if we abort at the first error token?), and maybe there's a way to pick a "reasonably selective" regex and use that for `recognize_text`?

Note that some languages don't do regex matches to decide what to do, though. I wrote a few `analyse_text` implementations which look at the distributions of characters (to recognize Brainfuck) or certain characters. How would those work with `recognize_text`?

not-my-profile:
> There's also the problem that we're making a global guess [..] instead of having a reasonable subset to select from (for instance, only disambiguating between all languages using .cl as the file ending.)

I'm confused ... isn't that what `pygments.lexers.guess_lexer_for_filename` is doing? (Having `pygments.lexers.guess_lexer` still makes sense for when you don't know the file extension.)

> I wonder if we can pick expressions stochastically from the lexer itself and use those for ranking.

Of course we could do that but I seriously doubt that it would result in high quality guesses. Because e.g. keywords are just a soft indicator ... they can always occur in strings. I think our best bet is to hand write regexes that are likely to occur at the top of the file (with some score) and then pick whichever lexer thinks it recognizes something first in the text (also factoring in said score).

>The main reason to not use the lexer itself and see if it produces errors is because it's too slow (is that still true if we abort at the first error token?)

```python
import pygments.lexers
import pygments.token

with open('pygments/lexer.py') as f:
    text = f.read()

error = []
ok = []

for cls in pygments.lexers._iter_lexerclasses():
    for (ttype, _) in cls().get_tokens(text):
        if ttype == pygments.token.Error:
            error.append(cls.__name__)
            break
    else:
        ok.append(cls.__name__)

print('ok', len(ok))
print('error', len(error))
```

Running that took 7 seconds for me and produced:

```
ok 125
error 397
```

meaning 125 lexers did not produce any error for `pygments/lexer.py` which IMO shows that we can forget that approach.

>maybe there's a way to pick a "reasonably selective" regex and use that for recognize_text?

I think good guessing requires knowledge about how files in the programming language are usually structured. That information cannot be extracted from lexers.

>I wrote a few analyse_text implementations which look at the distributions of characters (to recognize Brainfuck) or certain characters.

```python
>>> from pygments.lexers.esoteric import BrainfuckLexer
>>> BrainfuckLexer().analyse_text('<!DOCTYPE html>++++++++++[>+++++++>++++++++++>+++>+<<<<-]>++.>+.+++++++..+++.>++.<<+++++++++++++++.>.+++.------.--------.>+.>.')
1.0
```

That's exactly what I mean. `analyse_text` should not return 1 if it's not absolutely certain.

I think simply using a regex like `[-+<>[\]\s]{5,}` would be better. This way `guess_lexer` would have access to the start position and be able to give other lexers that recognized something earlier precedence.

PR #2006, opened 2021-12-29
GSQL lexer syntax changes

gregoryg:
`pygmentize` was hanging or taking minutes to parse certain GSQL code.

Syntax changes include: adding [], <>, and _ (square brackets, angle brackets and single underscore) to the Punctuation class.  
The test for that fix is now include in `tests/examplefiles/gsql/test.gsql` 

A one-liner to test for that fix:
```CREATE QUERY tg_astar (VERTEX source_vertex, VERTEX target_vertex, SET<STRING> e_type, STRING wt_type, STRING latitude, STRING longitude,```

The string regex was causing backtracking in some cases - simplified here.

In GSQL DDL, parameter such as `$0`, `$"fieldname"` and `_` to indicate empty position are possible.  They were being flagged as errors.   The test for that change is also added to the `test.gsql` file

A one-liner to test the DDL syntax fix:
```LOAD Inventory TO EDGE in_warehouse_inventory VALUES($"warehouse", $1, $_)```

Reusing edge names for multiple usage requires using pipe as an operator.  Pipe is now added to that section.

Test case was added to `test.gsql`

A one-liner test for edge name reuse : 
```  ADD UNDIRECTED EDGE in_code(FROM City, TO PostalCode | FROM Correspondence, TO PostalCode | FROM Address, TO PostalCode);```


Anteru:
Thanks, that was one nasty case of catastrophic backtracking you fixed there!

PR #2017, opened 2022-01-03
Update Agda’s reserved words

L-TChen:
A number of keywords are added and a deprecated keyword `unquoteGoal` is removed. For the latest set of keywords in Agda, see https://github.com/agda/agda/blob/master/src/full/Agda/Syntax/Parser/Lexer.x#L153-L193.


Anteru:
Merged, thanks!

PR #2021, opened 2022-01-05
Tango style: stop underlining whitespace

### LINKED ISSUE #2020: Whitespace in code receives underline for some themes, which can be confused with a `_` character
Hello, let me get this opportunity to first thank the authors and maintainers for the amazing package.
I also would like to apologize in advance if I am submitting the issue to the wrong repository.

Recently on setuptools, we received a bug report regarding our documentation: https://github.com/pypa/setuptools/issues/2999.

It seems that after the latest versions of `pygments` got released the default dark theme used in Sphinx for pygments started showing underlined whitespace, which is confusing for users (as reported in the linked issue):

image

After some quick investigation:

image

image

I think the idea here is that the colour of the whitespace (`.w` class, currently `#666666`) should be the same as the background (`#202020`), but it ended up being misconfigured - at least that is what happens in the light version of the theme (no idea why it is necessary to preserve the underline configuration instead of just removing it).

However I could not find out if this theme is something Sphinx itself is providing (in that case it would be better to open an issue there, right?), or if it comes directly from pygments.

My hypothesis is that this is a consequence of the whitespace improvements introduced in 2.11.0.

Do you guys have any advice/information about this?

amitkummer:
Fixes #2020.

To produce the output below I used:

```s
echo "int i = 1;" > test
pygmentize -l cpp -O full,style=tango -o test.html test
```

Before:

firefox_8L9qpHcQQa

After:

firefox_qyVVdgyvNK

I selected/marked the text with my cursor, so the underline difference is visible, otherwise the underline is invisible as it is the same color as the background (I hope this makes sense).

### DIFF
--- a/pygments/styles/tango.py
+++ b/pygments/styles/tango.py
@@ -55,7 +55,7 @@ class TangoStyle(Style):
     styles = {
         # No corresponding class for the following:
         #Text:                     "", # class:  ''
-        Whitespace:                "underline #f8f8f8",      # class: 'w'
+        Whitespace:                "#f8f8f8",                # class: 'w'
         Error:                     "#a40000 border:#ef2929", # class: 'err'
         Other:                     "#000000",                # class 'x'
 


Anteru:
Uh, looks like we both did the same change simultaneously, and because I can push directly I won ...: 05d1867453b93540344989e14cd2de27ccbe53de

I'm closing this but thanks for finding the issue and providing a PR!

PR #2022, opened 2022-01-05
CFamily: improve label detection

### LINKED ISSUE #1182: C/C++ conditional statement
(Original issue 1478 created by ashinkarov on 2018-10-18T03:45:03.514096+00:00)

In the following C/C++ program:
```c
int main() {
    return x?y:z;
}
```
`y` is still incorrectly identified as label.  I have:
```html
<span class="n">x</span><span class="o">?</span><span class="nl">y</span><span class="p">:</span><span class="n">z</span>
```
I am using `Pygments version 2.2.0, (c) 2006-2017 by Georg Brandl.`.

[The issue 436](https://bitbucket.org/birkenfeld/pygments-main/issues/436/the-c-lexer-may-parse-part-of-a-condition) is marked as resolved, but if I look at the code, I cannot wrap my head around why this should be the case.  
The existing rule for `Label` is this:
```python
(r'([a-zA-Z_]\w*)(\s*)(:)(?!:)', bygroups(Name.Label, Text, Punctuation))
```
So all it says is that `ID :<whatever>` is good, but `ID ::` is not.  However, this doesn't solve the original problem ---  in `x?y:z` `y` is still parsed as label.  I guess the right solution would be to enter a different state after we have seen `?` and within that state ignore the label rule.  As in C/C++ the `e1 ? e2 : e3` is an expression, therefore it cannot contain labels.

### LINKED ISSUE #1996: C++ lexer thinks that `:` can only follow a label name
This is a copy of some old issue: https://github.com/gpoore/minted/issues/195

Briefly: the token right before the comma is highlighted incorrectly in ternary operators (see image below) and range-based for loops. Right now the lexer is not aware of any other usage of `:` except indicating a label.

I have created a pull-request 3 years ago, but it has not been merged. All its details are lost, but as far as I remember, I have added a rule or two before or after this line: https://github.com/pygments/pygments/blob/8630e033313647d9579e314f3e8e2882f4558933/pygments/lexers/c_cpp.py#L85

It was probably not the best solution and just a workaround, but I didn't feel like digging in.

amitkummer:
Resolves #1996, resolves #1182.

### DIFF
--- a/pygments/lexers/c_cpp.py
+++ b/pygments/lexers/c_cpp.py
@@ -25,9 +25,6 @@ class CFamilyLexer(RegexLexer):
     definitions.
     """
 
-    #: optional Comment or Whitespace
-    _ws = r'(?:\s|//.*?\n|/[*].*?[*]/)+'
-
     # The trailing ?, rather than *, avoids a geometric performance drop here.
     #: only one /* */ style comment
     _ws1 = r'\s*(?:/[*].*?[*]/\s*)?'
--- a/pygments/lexers/c_cpp.py
+++ b/pygments/lexers/c_cpp.py
@@ -39,8 +39,8 @@ class CFamilyLexer(RegexLexer):
     _intsuffix = r'(([uU][lL]{0,2})|[lL]{1,2}[uU]?)?'
 
     # Identifier regex with C and C++ Universal Character Name (UCN) support.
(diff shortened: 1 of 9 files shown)


Anteru:
I can't claim I fully understood how you did it, but the tests look good :) Thanks a lot for working on this, merging for the 2.11.2 release!

Issue #1986, opened 2021-12-10
Custom tokens don't backtrack to the lower level


isidentical:
Normally styles backtrack to the lower level tokens when doing style resolution. For example if there is no custom styling for `Name.Function`, it will try to find the style of `Name` and use it. But this doesn't work with custom tokens:

```py
import pygments
from pygments.lexer import RegexLexer
from pygments.token import *
from pygments.style import Style
from pygments.formatters.terminal256 import Terminal256Formatter

class TestLexer(RegexLexer):
    tokens = {
        'root': [
            (r'A', Name.Function),
            (r'B', Name.Function.CUSTOM),
        ]
    }

class Style1(Style):
    styles = {
        Name: "#FA9BFA"
    }

class Style2(Style):
    styles = {
        Name.Function: "#FA9BFA",
        Name.Function.CUSTOM: "#FA9BFA"
    }


for style in [Style1, Style2]:
    print(pygments.highlight(
        code="AB",
        lexer=TestLexer(),
        formatter=Terminal256Formatter(style=style),
    ).strip())
```

I would expect both of these to work and output the same colors, but the initial style only highlights the `Name.Function`, and does not do anything for `Name.Function.CUSTOM`.
![image](https://user-images.githubusercontent.com/47358913/145557781-53c1ba9a-0deb-4ced-8a6d-615d047bcb99.png)

birkenfeld:
Do you mean "custom tokens"?

birkenfeld:
I agree.

Anteru:
Uh, that's actually a serious issue as we have a bunch of lexers relying on this...

Jean-Abou-Samra:
The `Terminal256Formatter` and `IRCFormatter` use the pattern `ttype = ttype[:-1]` instead of `ttype = ttype.parent`. Experimenting with it:

```
>>> from pygments.token import Token
>>> Token.Name.Custom[:-1]
('Name',)
```

So it makes `ttype` a tuple and lookups fail because they expect a token.

Is this the result of an earlier design of the `Token` object? Should we change it to make indexing return a token, or should we update the formatters to use the `parent` attribute?

There is also the `GroffFormatter`, which as far as I can see does not inherit styles at all.

Jean-Abou-Samra:
Thanks for the fix! I've opened #2024 about the GroffFormatter (Pygments is already two years ahead of its time, yay :-).

PR #2024, opened 2022-01-06
Let Groff formatter inherit token styles correctly

Jean-Abou-Samra:
Previously, Token.Name.Custom would raise KeyError on custom token
types not given rules in the style.

Fixes the rest of #1986.


Jean-Abou-Samra:
Gentle ping? This change is quite simple, review should normally take under 3 minutes :-)

birkenfeld:
Indeed, thanks!

PR #2026, opened 2022-01-06
Use `linespans` for `anchorlinenos`

mhils:
When using the HTML formatter with `linespans=foo`, `linenos=inline`, and `anchorlinenos=True`,
the generated anchor links should be `#foo-42` and not `#-42`.

Thanks for maintaining pygments, we are very happy users with [pdoc](https://pdoc.dev/)! 😃 🍰


Anteru:
Good catch, thanks!

PR #2031, opened 2022-01-11
Add .tpp as a c++ lexer file extension

HampusMat:
This PR will add the .tpp file extension to the c++ lexer.

If you're wondering, .tpp is a common file extension to use in c++ for header files that implement templates.
You can read more about it in [this relevant Stackoverflow post](https://stackoverflow.com/questions/495021/why-can-templates-only-be-implemented-in-the-header-file).


birkenfeld:
Thanks!

PR #2036, opened 2022-01-16
CI: Only deploy for upstream

hugovk:
Let's skip the deploy step for forks, it will fail because forks don't have `secrets.ACTIONS_DEPLOY_KEY`.

Also bump some versions.


birkenfeld:
@Anteru any objections?

Anteru:
No.

PR #2042, opened 2022-01-22
Highlight Clojure ratios as Numbers

TomFryers:
Currently, the numerator is highlighted differently from the slash and denominator, which looks slightly odd.


Anteru:
Merged, thanks!

PR #2043, opened 2022-01-22
Add support for .cljc clojure file extension

dannyfreeman:
.cljc files are clojure source files that run on both clojure and clojurescript runtimes. This PR adds support for formatting .cljc files by adding the .cljc extension to the ClojureLexer.

An example test file was created using this source file from the [re-frame](https://github.com/day8/re-frame) project: 
https://github.com/day8/re-frame/blob/7199496997cfb226311444f4402d1bda5798af60/src/re_frame/loggers.cljc
I generated the loggers.cljc.output file using `pytest tests/examplefiles --update-goldens` and visually checked the output with `python -m pygments tests/examplefiles/clojure/loggers.cljc`


dannyfreeman:
One of the [linters failed](https://github.com/pygments/pygments/runs/4918458883?check_suite_focus=true), and suggested I run `make mapfiles`. I did so and amended the commit.

birkenfeld:
Thanks! (The remaining check would have been fine after a rebase.)

PR #2046, opened 2022-01-23
Add the word But to the list of tokens

HelioGuilherme66:
Add the missing work But to the list of valid Gherkin tokens.


birkenfeld:
Thanks!

PR #2054, opened 2022-01-28
PHP: 8.x keywords

ancarda:
This is part of the work required to get PHP 8.x support in Pygments. There is something else that needs to be added onto this patch that I couldn't figure out.

The file `pygments/lexers/_php_builtins.py` needs to be regenerated as there are several new functions in the standard library, such as `str_starts_with` (8.0) and `array_is_list` (8.1). I tried to run it locally after installing `requirements.txt` with Pip, but I keep getting `ModuleNotFoundError`. I'm not a Python developer, what do I need to do to get this to rebuild?

### DIFF
--- a/pygments/lexers/php.py
+++ b/pygments/lexers/php.py
@@ -71,7 +71,7 @@ class ZephirLexer(RegexLexer):
             (r'(abstract|boolean|bool|char|class|const|double|enum|export|extends|final|'
              r'native|goto|implements|import|int|string|interface|long|ulong|char|uchar|'
              r'float|unsigned|private|protected|public|short|static|self|throws|reverse|'
-             r'transient|volatile)\b', Keyword.Reserved),
+             r'transient|volatile|readonly)\b', Keyword.Reserved),
             (r'(true|false|null|undefined)\b', Keyword.Constant),
             (r'(Array|Boolean|Date|_REQUEST|_COOKIE|_SESSION|'
              r'_GET|_POST|_SERVER|this|stdClass|range|count|iterator|'
@@ -221,7 +221,7 @@ class PhpLexer(RegexLexer):
              r'array|E_ALL|NULL|final|php_user_filter|interface|'
              r'implements|public|private|protected|abstract|clone|try|'
              r'catch|throw|this|use|namespace|trait|yield|'
-             r'finally)\b', Keyword),
+             r'finally|match)\b', Keyword),
             (r'(true|false|null)\b', Keyword.Constant),
             include('magicconstants'),
             (r'\$\{\$+' + _ident_inner + r'\}', Name.Variable),


birkenfeld:
A full error message would be helpful. Which module raises `ModuleNotFoundError`?

ancarda:
Sorry, that was awful of me for not including the error message.

I'm trying to execute the file directly. I don't know if this is the right thing to do or not.

```
% pip3 install -r requirements.txt 
Requirement already satisfied: pytest-cov in /opt/anaconda3/lib/python3.8/site-packages (from -r requirements.txt (line 1)) (3.0.0)
Requirement already satisfied: pytest-randomly in /opt/anaconda3/lib/python3.8/site-packages (from -r requirements.txt (line 2)) (3.11.0)
Requirement already satisfied: pytest>=6.0 in /opt/anaconda3/lib/python3.8/site-packages (from -r requirements.txt (line 3)) (6.2.3)
[...omitted...]

% python3 pygments/lexers/_php_builtins.py
  File "pygments/lexers/_php_builtins.py", line 4683, in <module>
    from urllib.request import urlretrieve
  File "/opt/anaconda3/lib/python3.8/urllib/request.py", line 88, in <module>
    import http.client
  File "/opt/anaconda3/lib/python3.8/http/client.py", line 71, in <module>
    import email.parser
ModuleNotFoundError: No module named 'email.parser'; 'email' is not a package
```

birkenfeld:
Oh right, this is due to the `pygments/lexers/email.py` being imported instead of the standard module. You can run it with `python3 -I` to avoid having the path of the script in the search path.

birkenfeld:
OK, since there were a few more snags, I've updated the file in https://github.com/pygments/pygments/pull/2056 - can you check if it matches what you'd expect?

pygments/lexers/php.py
```diff
@@ -221,7 +221,7 @@ class PhpLexer(RegexLexer):
              r'array|E_ALL|NULL|final|php_user_filter|interface|'
              r'implements|public|private|protected|abstract|clone|try|'
              r'catch|throw|this|use|namespace|trait|yield|'
-             r'finally)\b', Keyword),
+             r'finally|match)\b', Keyword),
```
ancarda, line 224: I'm pretty sure `readonly` is in the right place, however there's two places where `switch` is defined (`Keyword` and `Keyword.Declaration`). Is there a way I can test this to make sure it works? Match works just like switch:

```php
switch ($var) {
  case 'a':
    $result = "some value";
    break;
  case 'b':
    $result = "some other value";
    break;
}

// can be replaced with

$result = match ($var) {
  'a' => "some value",
  'b' => "some other value",
};
```

birkenfeld: I don't see the other place? In any case, this LGTM.

PR #2056, opened 2022-01-29
php: update builtin functions lists

birkenfeld:


ancarda (approved):
This looks good to me, thank you!

Despite being a large diff, this does include the 8.x functions like `array_is_list` and drops the removed functions like `import_request_variables`.

It seems a lot of old extensions I've never heard of were dropped too, like Hyperwave. These are still available via PECL, so shouldn't be included for highlighting like this. They aren't standard library functions anymore.

Anteru:
I carefully looked through all functions you've changed/added and I didn't see any issues.

Issue #2053, opened 2022-01-28
Elpi lexer has pathological backtracking


kurtmckee:
The Elpi lexer has pathological backtracking that can be triggered with the following minimal input:

```python
from pygments.lexers.elpi import ElpiLexer

list(ElpiLexer().get_tokens("a" * 30))
```

Only lowercase characters trigger pathological backtracking. Uppercase characters and digits will not.

The culprit is at [line 62](https://github.com/pygments/pygments/blob/a502df0a3ebcffb64125025a5e8a3452d8c63a17/pygments/lexers/elpi.py#L62).

@gares I'm not familiar with Elpi, so if you can jump in to address this, great! Otherwise I'll be able to take a crack at it after this weekend.

gares:
Thanks for pinning this down. What I'm trying to do is to make it so that, for `c` matched by `constant_re`:
- `c` takes `Text`; unless
- `c` starts with a capital, in which case it takes `Name.Variable`, line 61; or
- `c` is followed by `\`, in which case it takes `Name.Variable`, line 62.

My lexer is probably silly, but I don't know how to improve it. So help is welcome.

Examples:
- `foobar` is `Text`
- `Foobar` is `Name.Variable`
- `foobar\` is `Name.Variable`
- `Foobar\` is `Name.Variable` as well

birkenfeld:
The problem is that `constant_re` has an alternative `{}{}*` with `lcase_re` and `idcharstarns_re`, the latter of which contains repetitions as well.  So in effect you have `(something+)*` which has catastrophic backtracking properties.

Can the outer `*` be just removed here, or at least replaced by a `?`?

gares:
I would say it's legit, but I would put `something*` then, not `something+?`.

I read between the lines that these regexp are not compiled to an DFA. If they were all these expressions would collapse to the same. :-/

birkenfeld:
No; DFA regex engines don't support features like backreferences or look-behind assertions, which Python's re does.

Jean-Abou-Samra:
`idcharstarns_re` reads (reformatted):

```python
idcharstarns_re = rf"({idchar_re}+|\.{lcase_re}+)"
```

The relevant alternative of `constant_re` is

```
{lcase_re}{idcharstarns_re}*
```

This sounds like it's not matching what it intends. For example, it can match “`a.a5` where `5` matches `idchar_re`; the `.` works as long as there is a lowercase letter after it but it doesn't require more. What is the intended rule here?

[...]

PR #2058, opened 2022-01-31
Remove now redundant re.UNICODE and (?u)

Jean-Abou-Samra:
In Python 3, re.UNICODE, corresponding to the (?u) modifier,
is always used by default.  This removes its now unnecessary
explicit usage.  In lexers that had "flags = re.MULTILINE | re.UNICODE",
this line is just removed since (without lexer inheritance) re.MULTILINE
is the default.  Elsewhere, re.UNICODE is removed from the list
of flags.  In the CoqLexer, there was just re.UNICODE; it can be
removed altogether (bringing in re.MULTILINE) because the lexer
doesn't use ^ and $ anyway.

### DIFF
--- a/pygments/lexers/bare.py
+++ b/pygments/lexers/bare.py
@@ -26,8 +26,6 @@ class BareLexer(RegexLexer):
     filenames = ['*.bare']
     aliases = ['bare']
 
-    flags = re.MULTILINE | re.UNICODE
-
     keywords = [
         'type',
         'enum',
--- a/pygments/lexers/capnproto.py
+++ b/pygments/lexers/capnproto.py
@@ -26,8 +26,6 @@ class CapnProtoLexer(RegexLexer):
     filenames = ['*.capnp']
     aliases = ['capnp']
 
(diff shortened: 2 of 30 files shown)


birkenfeld:
LGTM, thanks!

Jean-Abou-Samra:
Ouch, GitHub tells me that this has caused a CI failure: https://github.com/Jean-Abou-Samra/pygments/actions/runs/1774370914. What I don't understand is why it has not been shown here -- everything was green. How comes that it failed later? (Sorry if this is a dumb question, I'm more used to GitLab and somewhat lost in GitHub's CI interface.) It seems to have failed with Python 3.9.10 on an infinite loop, but I cannot reproduce this locally -- the complete test suite passes for me with Python 3.9.7. Any clue?

birkenfeld:
Probably due to https://github.com/pygments/pygments/issues/2053 - don't worry about it.

Jean-Abou-Samra:
Alright, I see now the test with random input. Thank you for reassuring me; must have been unrelated indeed.

PR #2061, opened 2022-02-03
Elpi: fix catastrophic backtracking

### LINKED ISSUE #2053: Elpi lexer has pathological backtracking

Jean-Abou-Samra:
Closes #2053

### DIFF
--- a/pygments/lexers/elpi.py
+++ b/pygments/lexers/elpi.py
@@ -32,9 +32,9 @@ class ElpiLexer(RegexLexer):
     schar2_re = r"([+*^?/<>`'@#~=&!])"
     schar_re = r"({}|-|\$|_)".format(schar2_re)
     idchar_re = r"({}|{}|{}|{})".format(lcase_re,ucase_re,digit_re,schar_re)
-    idcharstarns_re = r"({}+|(?=\.[a-z])\.{}+)".format(idchar_re,idchar_re)
+    idcharstarns_re = r"({}*(\.({}|{}){}*)*)".format(idchar_re, lcase_re, ucase_re, idchar_re)
     symbchar_re = r"({}|{}|{}|{}|:)".format(lcase_re, ucase_re, digit_re, schar_re)
-    constant_re = r"({}{}*|{}{}*|{}{}*|_{}+)".format(ucase_re, idchar_re, lcase_re, idcharstarns_re,schar2_re, symbchar_re,idchar_re)
+    constant_re = r"({}{}*|{}{}|{}{}*|_{}+)".format(ucase_re, idchar_re, lcase_re, idcharstarns_re,schar2_re, symbchar_re,idchar_re)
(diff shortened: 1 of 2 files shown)


pygments/lexers/elpi.py
```diff
@@ -2,142 +2,172 @@
...
-    idcharstarns_re = r"({}+|(?=\.[a-z])\.{}+)".format(idchar_re,idchar_re)
-    symbchar_re = r"({}|{}|{}|{}|:)".format(lcase_re, ucase_re, digit_re, schar_re)
-    constant_re = r"({}{}*|{}{}*|{}{}*|_{}+)".format(ucase_re, idchar_re, lcase_re, idcharstarns_re,schar2_re, symbchar_re,idchar_re)
-    symbol_re=r"(,|<=>|->|:-|;|\?-|->|&|=>|\bas\b|\buvar\b|<|=<|=|==|>=|>|\bi<|\bi=<|\bi>=|\bi>|\bis\b|\br<|\br=<|\br>=|\br>|\bs<|\bs=<|\bs>=|\bs>|@|::|\[\]|`->|`:|`:=|\^|-|\+|\bi-|\bi\+|r-|r\+|/|\*|\bdiv\b|\bi\*|\bmod\b|\br\*|~|\bi~|\br~)"
-    escape_re=r"\(({}|{})\)".format(constant_re,symbol_re)
-    const_sym_re = r"({}|{}|{})".format(constant_re,symbol_re,escape_re)
+    schar_re = rf"({schar2_re}|-|\$|_)"
+    idchar_re = rf"([a-zA-Z]|{digit_re}|{schar_re})"
+    symbchar_re = rf"({idchar_re}|:)"
+    constant_re = rf"([A-Z]{idchar_re}*|[a-z]{idchar_re}*(\.[a-zA-Z]{idchar_re}*)*|{schar2_re}{symbchar_re}*|_{idchar_re}+)"
```
Jean-Abou-Samra, line 34: Note that this doesn't allow dots when the constant begins with a capital letter. I assume that is correct as it was already the case before.

gares: it is

gares (approved):

birkenfeld (changes requested):
The reformatting makes it basically impossible to see what has really changed.

Please don't apply `black` to the Pygments codebase, as you can see we have a pretty different style.

Jean-Abou-Samra:
> The reformatting makes it basically impossible to see what has really changed.
>
> Please don't apply black to the Pygments codebase, as you can see we have a pretty different style.

OK, I'll back the formatting out. Note, though, that it is in a separate commit for the purpose of review (I've been trained rigorously to use atomic commits, but maybe that is not usual here?).

birkenfeld:
As far as I can see, the formatting changes in the first commit ("minor changes") which also has a bunch of other (useful) things?

birkenfeld:
> I've been trained rigorously to use atomic commits, but maybe that is not usual here?

Of course it's useful, but even with atomic commits I wouldn't expect to see a wholesale reformatting in a PR to fix a bug :-)

Jean-Abou-Samra:
I have refactored this PR to include only the bugfix.

birkenfeld (approved):
Thanks a lot!

PR #2066, opened 2022-02-09
Add support for non-idiomatic single line comments in terraform

### LINKED ISSUE #2065: Terraform cannot lex terraform containing non-idiomatic single line comment.
Hello there, and thank your for Pygments!

The following valid terraform string cannot be lexed:

```terraform
// Non-idiomatic single-line comment.
```

See https://www.terraform.io/language/syntax/configuration#comments

This must not happens often since most terraform users will format their code with tools, and 

> Automatic configuration formatting tools may automatically transform // comments into # comments, since the double-slash style is not idiomatic.

cblegare:
Fixes #2065


birkenfeld:
Thanks for the fix!

PR #2067, opened 2022-02-19
Spice: Update some keywords

marcauberer:
- Add keywords `inline` and `public` and remove keyword `new`
- Update url to https and www


birkenfeld:
Thanks!

PR #2069, opened 2022-02-24
Fix catastrophic backtracking in HTML and XML

### LINKED ISSUE #2068: Exponential slowdown parsing Genshi templates with unfinished tags in comments
As of Pygments 2.11.0, when parsing an XML file using the Genshi lexer, comments that include an unclosed/unfinished tag can result in substantial slowdown.

Take the following sample XML document:

```xml
<?xml version="1.0"?>

<project>
  <!-- <move -->

  <target>
    <x />
    <x />
    <x />
    <x />
    <x />
    <x />
    <x />
    <x />
    <x />
    <x />
    <x />
    <x />
    <x />
    <x />
    <x />
    <x />
    <x />
    <x />
  </target>
</project>
```

Running this with `time pygmentize -l genshi test.xml` takes about 3 seconds locally.

Add one more `<x />` to the list, and that becomes 6 seconds. One more, 12 seconds. Then 24. 48. And so on.

This was first encountered in a real-world Ant buildfile. The above XML file is the simplest repro case I could devise.

If you change the comment to include an open or terminated tag (`<!-- <move> -->` or `<!-- <move/> -->`), the problem goes away.

I've bisected the Pygments code, and can verify that this regressed with commit 337dddef92b69d3a1e88fa450bef55149d24c1d9.

I've also verified that this is still an issue as of the latest commit on `master` (be3e35557cccf8c6477602f6341ba1b7ca468793).

Jean-Abou-Samra:
These lexers have re.DOTALL in their flags, so r"(.|\n)" is equivalent
to r".", except with catastrophic backtracking properties.

Closes #2068


birkenfeld:
Damn that's hard to spot! (Also would have assumed that this is optimized by the engine.... It's really about time to switch to regex.)

Anteru:
Aye :( Thanks for figuring this out. Yeah, if the maintainers can't be trusted to get the regex right there's no hope with the current regex engine.

birkenfeld:
regex recently got high praise from Tim Peters on python-dev, so I think we should be fine trusting it.

PR #2073, opened 2022-02-27
add K and Q lexers

psaris:
a jupyter notebook kernel is available for the Q language: https://github.com/KxSystems/jupyterq

syntax highlighting gets lost when using 'jupyter nbconvert' because there is no pygments lexer.

please accept the submission for a new Q lexer which is based on the K lexer (just as the Q programming language was created as a macro layer on the underlying K programming language).


pygments/lexers/q.py
```diff
@@ -0,0 +1,228 @@
...
+                bygroups(Name.Variable, Whitespace, Operator),
+            ),
+            # Functions
+            (r"\{", Punctuation, "functions"),
+            # Parentheses
+            (r"\(", Punctuation, "parentheses"),
+            # Brackets
+            (r"\[", Punctuation, "brackets"),
+            # Errors
+            (r"'`[a-zA-Z]*(\w|[.])*", Name.Exception),
```
Jean-Abou-Samra, line 100: Do you really mean the star on `[a-zA-Z]`? (Same below.)

As a mundane matter of style, `[\w.]` is a little easier to read than `(\w|[.])`.

pygments/lexers/q.py
```diff
@@ -0,0 +1,228 @@
...
+            # File Symbols
+            (r"`:[a-zA-Z/]*(\w|[./])*", String.Symbol),
+            # Symbols
+            (r"[`][a-zA-Z]*(\w|[.])*", String.Symbol),
```
Jean-Abou-Samra, line 104: Aren't the brackets around the backtick unneeded?

pygments/lexers/q.py
```diff
@@ -0,0 +1,228 @@
...
+            # Numbers
+            include("numbers"),
+            # Variable Names
+            (r"[a-zA-Z](\w|[.])*", Name),
+            # Operators
+            (r"[-=+*#$%@!~^&:.,<>'\\|/?_]", Operator),
+            # Punctuation
+            (r"[;]", Punctuation),
```
Jean-Abou-Samra, line 112: Why the brackets?

pygments/lexers/q.py
```diff
@@ -0,0 +1,228 @@
...
+            (
+                (
+                    r"[0-9]{4}[.][0-9]{2}"
+                    "(?:m|[.][0-9]{2}(?:T(?:[0-9]{2}:[0-9]{2}"
+                    "(?:[:][0-9]{2}(?:[.][0-9]*)?)?)?)?)"
+                ),
+                Literal.Date,  # datetime
+            ),
+            (
+                (r"[0-9]{2}:[0-9]{2}" "(?:[:][0-9]{2}" "(?:[.][0-9]{1,3})?)?"),
```
Jean-Abou-Samra, line 146: The implicit string concatenation on the same line seems coming from a reformatting.

pygments/lexers/q.py
```diff
@@ -0,0 +1,228 @@
...
+            (r"[0-9]+[jnp]", Number.Integer.Long),
+            (r"[0-9]+[ihtuv]", Number.Integer),
+            (r"([0-9]*[.]?[0-9]+|[0-9]+[.]?[0-9]*)[eE][+-]?[0-9]+[ef]?", Number.Float),
+            (r"([0-9]*[.][0-9]+|[0-9]+[.][0-9]*)[ef]?", Number.Float),
+            (r"[0-9]+[ef]", Number.Float),
+            (r"[0-9]+[c]", Number),
+            (r"[0-9]+", Number.Integer.Long),
+        ],
+        "comments": [
+            (r"[^\\]", Comment.Multiline),
```
Jean-Abou-Samra, line 163: This is a known antipattern: it generates one token per character, which makes the raw token output (such as in the regression test) much harder to read, and can slow down highlighting too. Please add a `+` quantifier.

Another way to do it would be `^/\s*\n(.|\n)*?^\\` to match a multiline comment and do without a dedicated state altogether.

pygments/lexers/q.py
```diff
@@ -0,0 +1,228 @@
...
+            (r"^\\", Comment.Multiline, "#pop"),
+            (r"[\\]", Comment.Multiline),
+        ],
+        "strings": [
+            (r"\\.", String.Double),
+            (r"[^\"]", String.Double),
```
Jean-Abou-Samra, line 169: Same remark here.

pygments/lexers/q.py
```diff
@@ -0,0 +1,228 @@
...
+    aliases = ["q"]
+    filenames = ["*.q"]
+
+    tokens = {
+        "root": [
+            include("whitespace"),
+            include("keywords"),
+            include("macros"),
+            include("declarations"),
+        ],
```
Jean-Abou-Samra, line 192: Maybe you can simplify that as

```
"root": [
  (words(...), Operator.Word),
  inherit,
]
```

?

pygments/lexers/q.py
```diff
@@ -0,0 +1,228 @@
...
+        "macros": [
+            (
+                words(
+                    ("aj", "aj0", "ajf", "ajf0", "all", "and")
+                    + ("any", "asc", "asof", "attr", "avgs")
```
Jean-Abou-Samra, line 197: `+` signs unneeded as above.

Jean-Abou-Samra (changes requested):
Thanks, this looks mostly good to me, with some unimportant remarks but a few important ones too.

psaris:
> 

thank you @Jean-Abou-Samra.  i appreciate your detailed and constructive suggestions.

pygments/lexers/q.py
```diff
@@ -0,0 +1,228 @@
...
+    """
+
+    name = "K"
+    aliases = ["k"]
+    filenames = ["*.k"]
+
+    tokens = {
+        "whitespace": [
+            # hashbang script
+            (r"^#!.*$", Comment.Hashbang),
```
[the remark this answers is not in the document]
psaris: thank you.  will remove the trailing `$`

pygments/lexers/q.py
```diff
@@ -0,0 +1,228 @@
...
+            # Comments
+            (r"^/\s*\n", Comment.Multiline, "comments"),
+            (r"[ \t]+/.*", Comment.Single),
```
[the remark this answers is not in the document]
psaris: yes, the syntax of k is quite interesting.  the `/` character is in-fact an operator when it comes directly after any character other than white space.  the following, for example, can compute 10 factorial
```
  */1+!10
3628800
```
but adding a space turns everything after the `*` into a comment
```
  * / 100
*
```

pygments/lexers/q.py
```diff
@@ -0,0 +1,228 @@
...
+            (r"^/.*\S.*", Comment.Single),
```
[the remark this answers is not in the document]
psaris: the distinction is that a line starting with a `/` followed by any non-whitespace is a single-line comment.  while a line that only has a single `/` is the start of a multi-line comment (which end with a line that only has `'`)

pygments/lexers/q.py
```diff
@@ -0,0 +1,228 @@
...
+        "root": [
+            include("whitespace"),
+            include("keywords"),
+            include("declarations"),
+        ],
+        "keywords": [
+            (
+                words(
+                    ("abs", "acos", "asin", "atan", "avg", "bin", "binr", "by")
+                    + ("cor", "cos", "cov", "dev", "delete", "div", "do")
```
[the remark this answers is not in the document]
psaris: if i remove the the bracketing and use of the `+` operator the [black formatter](https://black.readthedocs.io/en/stable/) puts every key word on a new line.  do you have a suggested way around this (other than wrapping the list with `#fmt: off` and `#fmt: on`)?

```
                words(
                    (
                        "abs",
                        "acos",
                        "asin",
                        "atan",
                        "avg",
                        "bin",
```

pygments/lexers/q.py
```diff
@@ -0,0 +1,228 @@
...
+                    + ("select", "setenv", "sin", "sqrt", "ss", "sum")
+                    + ("tan", "update", "var")
+                    + ("wavg", "while", "within", "wsum", "xexp"),
+                    suffix=r"\b",
+                ),
+                Keyword,
+            ),
+        ],
+        "declarations": [
+            (r"^\\t[s]?", Comment.Preproc),
```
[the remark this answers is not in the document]
psaris: yes, agreed.  thank you.

pygments/lexers/q.py
psaris: yes, this was needed to allow the string to spread across lines. i subsequently optimized the regex and the black formatter merged the shortened lines.  will fix.

pygments/lexers/q.py
psaris: thank you for the suggestion.  i hated the regression test output as well!  this is a result of following the [pygment documentation](https://pygments.org/docs/lexerdevelopment/#changing-states).  specifically:

```python
    tokens = {
        'root': [
            (r'[^/]+', Text),
            (r'/\*', Comment.Multiline, 'comment'),
            (r'//.*?$', Comment.Singleline),
            (r'/', Text)
        ],
        'comment': [
            (r'[^*/]', Comment.Multiline),
            (r'/\*', Comment.Multiline, '#push'),
            (r'\*/', Comment.Multiline, '#pop'),
            (r'[*/]', Comment.Multiline)
        ]
    }
```

pygments/lexers/q.py
psaris: will fix

pygments/lexers/q.py
psaris: again, do you have a clean way of preventing the black formatter from moving every tuple element onto it's own line?

pygments/lexers/q.py
psaris: if there are keywords in comments (or strings) , they would be tokenized first, and then the comments (or strings) would not highlight properly.  i'll add a test case to ensure this is caught.

psaris:
@Jean-Abou-Samra thank you for your constructive comments.  i've made the requested updates (and included one formatting question)

pygments/lexers/q.py
psaris: agreed, removed

pygments/lexers/q.py
psaris: agreed, removed

pygments/lexers/q.py
psaris: yes, i do require the `*` after `[a-zA-Z]` because the null symbol `` ` `` is a valid symbol.

thank you for the `[\w.]` suggestion. i didn't realize you could put `\w` within `[]`

pygments/lexers/q.py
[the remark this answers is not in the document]
birkenfeld: Simply, don't use `black` in Pygments code. (And please remove the formatting markers, they're not useful in our code base.)

pygments/lexers/q.py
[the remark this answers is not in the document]
birkenfeld: But the last expression doesn't match only non-whitespace after the `/`?

birkenfeld:
Thanks for the hint about the suboptimal example in the docs! This should be fixed with #2075.

pygments/lexers/q.py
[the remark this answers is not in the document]
Jean-Abou-Samra: OK. On the other hand, `[ \t]` doesn't include all possible whitespace characters -- doesn't it work with, say carriage returns? I think it would be easiest to use just two rules for comments:

```python
(r"^/\s*\n", Comment.Multiline, "comments"),
(r"(?<=^|\s)/.*", Comment.Single),
```

Jean-Abou-Samra:
One more thing: please add these languages to `doc/languages.rst`.

pygments/lexers/q.py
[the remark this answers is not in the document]
psaris: a combination of @Jean-Abou-Samra 's suggestion and a stackoverflow search resulted in the following two expressions.  in essence, a comment is either a line with only a leading ``/`` (and thus multi-line) or else it is a ``/`` that is not preceded by a non-whitespace character

```python
            (r"^/\s*\n", Comment.Multiline, "comments"),
            (r"(?<!\S)/.*", Comment.Single),
```

pygments/lexers/q.py
psaris: remove formatting comments for black

psaris:
@Jean-Abou-Samra @birkenfeld  have made the requested updates

pygments/lexers/q.py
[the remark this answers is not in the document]
psaris: the proposed ``(?<=^|\s)`` throws the following exception:

```
Error: error when loading custom lexer: uncompilable regex '(?<=^|\\s)/.*' in state 'whitespace' of <class 'QLexer'>: look-behind requires fixed-width pattern
```
negating the look-behind seems to do the trick:
```python
            (r"(?<!\S)/.*", Comment.Single),
```

pygments/lexers/q.py
```diff
@@ -0,0 +1,235 @@
...
+                     "signum", "ssr", "string", "sublist", "sums",
+                     "sv", "svar", "system", "tables", "til", "trim",
+                     "txf", "type", "uj", "ujf", "ungroup", "union",
+                     "upper", "upsert", "value", "view", "views", "vs",
+                     "where", "wj", "wj1", "ww", "xasc", "xbar",
+                     "xcol", "xcols", "xdesc", "xgroup", "xkey",
+                     "xlog", "xprev", "xrank"),
+                    suffix=r"\b",
+                ),
+                Operator.Word,
```
Jean-Abou-Samra, line 231: `Name.Builtin`, not `Operator.Word`?

Jean-Abou-Samra (commented):
One last remarks, generally looks good to me.

pygments/lexers/q.py
psaris: my goal is to have the K (C++ defined operators) in one  style (i chose `Keyword`) and the Q operators (defined as macros on top of K) in another style.  I chose `Operator.Word`  because i usually refer to them as operators and the documentation made sense to me.  
```
Operator.Word
    For any operator that is a word (e.g. not).
```
in fact, each of the Q operators can be infix just like a single character operator.  in fact many of the names (such as ``reverse`` are just a single character in K ``|`` but given a name in Q to make the language more palatable.

i'm having trouble intuiting the difference between your suggestion ``Name.Builtin`` and ``Operator.Word``.  how do you typically differentiate the two? i've tested the output with with html and instead of highlighting in purple, ``Name.Builtin``  is green as well (just not bold like ``Keyword``.  happy to switch if it makes more sense to you.

psaris:
@Jean-Abou-Samra have responded to your suggestion. gave some reasons why i picked ``Operator.Keyword`` but am open to your suggestion if it fits more naturally with ``Pygments``.

pygments/lexers/q.py
Jean-Abou-Samra: Typically, `Name.Builtin` is for built-in functions (think Python `print()`, `open()`, etc.) whereas `Operator.Word` is for infix operators that are not spelt with special characters but as a word, and usually are reserved names (can't be used as variables), which corresponds to Python's `and`, `or`, `not` and `is`. I hadn't realized these could be used with infix syntax. In that case, `Operator.Word` should be OK (it's not always possible to map concepts 1:1 between languages). By the way, note that the HTML output you tested uses the default style, there are many more. Have a look at https://pygments.org/styles/. There is no `Operator.Word` in the example, but if you want to see how it looks like, change `doc/examples/example.py` and build the docs locally.

psaris:
@Jean-Abou-Samra i've updated the K c++ defined keywords as Operator.Word and and Q K-defined keywords as Name.Builtin.

pygments/lexers/q.py
```diff
@@ -0,0 +1,235 @@
...
+            (r"[0-9]+[c]", Number),
+            # Integers
+            (r"[0-9]+[ihtuv]", Number.Integer),
+            # Long Integers
+            (r"[0-9]+[jnp]?", Number.Integer.Long),
+        ],
+        "comments": [
+            (r"[^\\]+", Comment.Multiline),
+            (r"^\\", Comment.Multiline, "#pop"),
+            (r"[\\]", Comment.Multiline),
```
Jean-Abou-Samra, line 180: This makes the CI fail as `regexlint` complains about the useless character class.

pygments/lexers/q.py
```diff
@@ -0,0 +1,235 @@
...
+        ],
+        "declarations": [
+            # Timing
+            (r"^\\ts?", Comment.Preproc),
+            (
+                r"^(\\\w\s+[^/\n]*?)([/].*)",
+                bygroups(Comment.Preproc, Comment.Single),
+            ),
+            # Generic System Commands
+            (r"^\\\w+.*$", Comment.Preproc),
```
Jean-Abou-Samra, line 81: While you are putting the final touch, please remove the useless `+` and `$` here.

pygments/lexers/q.py
psaris: i've corrected all the regexlint errors

pygments/lexers/q.py
psaris: i've cleaned this up as well

Jean-Abou-Samra:
Thanks for all the updates! Merged now.

birkenfeld:
Needs a changelog-update label?

Jean-Abou-Samra:
I thought this was implied by not setting the milestone? (CC @Anteru)

birkenfeld:
You're right, sorry, I misremembered the procedure :)

psaris:
@birkenfeld @Jean-Abou-Samra, i see that this lexer has been included in the 2.12.0 release.  i don't see it in the release notes however: https://github.com/pygments/pygments/releases/tag/2.12.0

Jean-Abou-Samra:
Well, so our procedure for CHANGES updates is not working as well as it could :-)

@Anteru Did you just not notice this PR among those without a milestone, or did I do something wrong?

How about simplifying our lives and always setting Changelog-update on PRs that will need a CHANGES update, independently from milestones?

psaris:
@Anteru  in addition the the procedure changes, will/can the release notes be updated?

Jean-Abou-Samra:
@psaris I think @Anteru has personal reasons not to be available at the moment, so please give him a little time.

Jean-Abou-Samra:
I'm doing the CHANGES update now. For the future I suggest always using the Changelog-update label as "single source of truth".

PR #2074, opened 2022-02-28
Update Stan to 2.29 standards

WardBrian:
Stan has undergone a lot of changes since the last time the lexer here was updated. A short summary:

1. Array types can now be declared with the syntax `array[2] int x` instead of `int x[2]`
2. Complex numbers have been added, with a `complex` type and support for imaginary literals like `3i`
3. Many functions and distributions have been added, alongside a couple operators
4. Most of the reserved words which were reserved for technical reasons are no longer reserved


pygments/lexers/_stan_builtins.py
```diff
@@ -436,122 +601,51 @@
...
     'uniform',
     'von_mises',
     'weibull',
     'wiener',
-    'wishart',
+    'wishart'
 )
 
 RESERVED = (
-    'alignas',
```
Jean-Abou-Samra, line 460: Are these words no longer special at all, or (as I suspect) are they just no longer reserved as identifiers? In the latter case, you should highlight them as `Token.Builtin`.

In any case, please make sure this list of words is covered by the test; I can't find `Keyword.Reserved` in the token output.

pygments/lexers/_stan_builtins.py
```diff
@@ -436,122 +601,51 @@
...
-    'STAN_MINOR',
-    'STAN_PATCH',
-    'static',
-    'static_assert',
-    'static_cast',
-    'struct',
-    'switch',
-    'template',
+    'until',
+    'if',
```
Jean-Abou-Samra, line 638: Seems duplicated with the keyword list?

pygments/lexers/_stan_builtins.py
```diff
@@ -253,102 +330,163 @@
...
+    'pareto_lupdf',
     'pareto_rng',
     'pareto_type_2_cdf',
     'pareto_type_2_lccdf',
     'pareto_type_2_lcdf',
     'pareto_type_2_lpdf',
+    'pareto_type_2_lupdf',
     'pareto_type_2_rng',
-    'Phi',
-    'Phi_approx',
```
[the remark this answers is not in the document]
WardBrian: These weren’t truly removed, the script I used to print the function names just put them at the start since they begin with capital letters. I’ll move them back and also add back `print` and `reject`

pygments/lexers/_stan_builtins.py
WardBrian: These genuinely are meaningless now. They were reserved previously because all Stan identifiers needed to be valid C++ identifiers, which has been relaxed. 

I can add them to the test file

pygments/lexers/_stan_builtins.py
WardBrian: There are duplicate entries; they were there previously as well. Generally speaking it is true that “if” is a reserved word in Stan, but I don’t know if it needs to be listed in both places in Pygments

pygments/lexers/_stan_builtins.py
Jean-Abou-Samra: Ah, OK. What I would like to see in the test file is something using one reserved keyword, so not one of those that are being removed but one that is staying.

pygments/lexers/_stan_builtins.py
Jean-Abou-Samra: In any case, they will be highlighted one way or the other. Looking at the lexer, it seems to prefer Keyword over Keyword.Reserved. It's not super important, but it would be nice to deduplicate the list, to make this less confusing.

Jean-Abou-Samra:
LGTM.

PR #2075, opened 2022-03-01
Add notes to Contributing.md about common mistakes in lexers

Jean-Abou-Samra:
While at it, fix some lexer examples, mostly to avoid emitting tokens
for single characters repeatedly.

### DIFF
--- a/Contributing.md
+++ b/Contributing.md
@@ -26,7 +26,7 @@ Contribution checklist
   [a new formatter](https://pygments.org/docs/formatterdevelopment/) or
   [a new filter](https://pygments.org/docs/filterdevelopment/)
 
-* Make sure to add a test for your new functionality, and where applicable, 
+* Make sure to add a test for your new functionality, and where applicable,
   write documentation.
 
 * When writing rules, try to merge simple rules. For instance, combine:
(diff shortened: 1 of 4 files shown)


doc/docs/lexerdevelopment.rst
```diff
@@ -376,20 +376,20 @@ There are a few more things you can do with states:
       class ExampleLexer(RegexLexer):
           tokens = {
               'comments': [
-                  (r'/\*.*?\*/', Comment),
+                  (r'/\*(.|\n)*?\*/', Comment),
```
birkenfeld, line 379: Maybe it's more foolproof to set `(?s)` for the pattern after the bug we had here lately :)

Contributing.md
```diff
@@ -53,6 +53,68 @@ Contribution checklist
   instead of matching ``@first@`` and ``@second@``. You can use ``@.*?@`` in
   this case to stop early. The ``?`` tries to match _as few times_ as possible.
 
+* Beware of so-called "catastrophic backtracking".  Here is a badly written
+  regular expression to match strings:
+
+  ```python
+  r'"(\\?.)*?"'
```
birkenfeld, line 60: This is a more non-obvious example, which is good, but you could additionally give the "easy one" with repeated repetition, like `(a*)+`.

Contributing.md
```diff
@@ -53,6 +53,68 @@ Contribution checklist
...
+
+  ```python
+  'comment': [
+      (r'\*/', Comment.Multiline, '#pop'),
+      (r'.', Comment.Multiline),
+  ]
+  ```
+
+  This generates one token per character in the comment, which makes the
+  raw token output (and in particular the test output) hard to read.
```
birkenfeld, line 107: It's also making everything slower.

birkenfeld (approved):
Nice idea, only a few suggestions to consider.

birkenfeld:
Very nice!

Issue #2077, opened 2022-03-02
ExtendedRegexLexer causes exception due to typo in code


brunosmmm:
I'm using a subclass of ExtendedRegexLexer for highlighting a custom DSL. In some obscure cases, I've discovered that the lexer fails, throwing an exception due to what I believe to be a typo in the source here:
https://github.com/pygments/pygments/blob/231366c7106f66f67f3f27c86ee51434419fcda2/pygments/lexer.py#L741
which should read
```python
del ctx.stack[1:]
```

PR #2078, opened 2022-03-02
Fix typo that prevents ExtendedRegexLexer from working properly

brunosmmm:
Please see #2077


birkenfeld:
Wow, that was long undiscovered! If you have a case that failed there, is it possible to make it into a test case?

brunosmmm:
Hi, I'm not sure I'll be able to accomplish that since the case appears in
a custom DSL that I have which is complex and I'm trying to highlight in a
semantically correct way. Any ideas?

birkenfeld:
Don't worry about it too much :)

PR #2080, opened 2022-03-03
Bump pytest requirement

Jean-Abou-Samra:
We require pytest 7 for the new from_parent() signature.


birkenfeld:
Ah, didn't realize that.

PR #2081, opened 2022-03-04
Make Formatter generic for the benefit of typeshed

not-my-profile:
The type stubs for Pygments in the typeshed have made `pygments.formatter.Formatter` generic in https://github.com/python/typeshed/pull/6819 so that an `outfile` passed to `pygments.highlight()` must be opened in binary mode when the formatter was created with a given `encoding`, for example:

```python
import pygments
import pygments.lexers
import pygments.formatters

lexer = pygments.lexers.get_lexer_by_name('python')
formatter = pygments.formatters.html.HtmlFormatter(encoding='utf-8')

with open('out.html', 'wb') as f: # mypy reports a type error if opened with 'w'
    pygments.highlight('test', lexer, formatter, outfile=f)
```

This does however come with the disadvantage that subclassing `Formatter` no longer type checks with `mypy --strict`:

```python
class Foo(pygments.formatter.Formatter):
     pass
```

because mypy complains with `Missing type parameters for generic type "Formatter"`. The problem with that is that fixing this by specifying the type parameter would result in a runtime error because the real `pygments.formatter.Formatter` is not generic and thus not subscriptable.

Making the actual `Formatter` generic improves the situation for pygments users using static type checkers and should not change anything for the pygments users that don't.

### DIFF
--- a/pygments/formatter.py
+++ b/pygments/formatter.py
@@ -9,6 +9,7 @@
 """
 
 import codecs
+from typing import Generic, TypeVar
 
 from pygments.util import get_bool_opt
 from pygments.styles import get_style_by_name
@@ -21,8 +22,11 @@ def _lookup_style(style):
         return get_style_by_name(style)
     return style
 
+_T = TypeVar('_T', str, bytes)
+# Formatter is only generic for the benefit of the Pygments type stubs
+# living in the typeshed (https://github.com/python/typeshed).
 
-class Formatter:
+class Formatter(Generic[_T]):
     """
     Converts a token stream to text.
 
--- a/pygments/formatters/img.py
+++ b/pygments/formatters/img.py
@@ -228,7 +228,7 @@ class FontManager:
             return self.fonts['NORMAL']
 
 
-class ImageFormatter(Formatter):
+class ImageFormatter(Formatter[bytes]):
     """
     Create a PNG image from source code. This uses the Python Imaging Library to
     generate a pixmap from the source code.


birkenfeld:
I'm not very keen on merging this; it would be an isolated case of strange typing stuff without any benefit for us.

If providing types via typeshed cannot handle such cases, it looks like it's buggy and should be fixed.

not-my-profile:
Right, I understand.

Dreamsorcerer:
Providing types via typeshed is already a suboptimal solution and generally considered to be temporary. The design of that class in pygments makes it difficult to type correctly, and impossible to not introduce these discrepancies between stubs and code. Ideally, the typing information should be moved here and supported as part of the project itself.

If that is an option and someone wants to start on it, I'm happy to help with code reviews.

birkenfeld:
I cannot say I like how the "optional" typing is creeping in everywhere.

If the other @pygments/maintainers are strongly in favor, I won't veto it, but I won't do the work.

Anteru:
I'd only do typing if we do it a) properly and b) once a newer version of Python which makes it easier to express the information becomes our baseline, most likely 3.10. Till then I see limited value -- and it will lead the codebase in a "half done" state which I really don't look forward to.

Dreamsorcerer:
Sounds reasonable. If there are things missing in the typing module from newer versions, you can always just import them from typing_extensions in older python versions, in case you want to do it a bit earlier.

Jean-Abou-Samra:
I'm totally new as a committer here, but I concur with @birkenfeld and @Anteru. One point I'd like to raise is that, by nature, Pygments is a project attracting many different one-time contributors adding and updating lexers for their pet languages, who are fluent in that language and may not be fluent in Python. Judging from the concepts proposed in typing PEPs posted on python-dev, it seems to me that writing typed Python code still poses a number of challenges, and thus makes for a steeper learning curve for modifying Pygments code than I would prefer.

danieleades:
> I'm totally new as a committer here, but I concur with @birkenfeld and @Anteru. One point I'd like to raise is that, by nature, Pygments is a project attracting many different one-time contributors adding and updating lexers for their pet languages, who are fluent in that language and may not be fluent in Python. Judging from the concepts proposed in typing PEPs posted on python-dev, it seems to me that writing typed Python code still poses a number of challenges, and thus makes for a steeper learning curve for modifying Pygments code than I would prefer.

strong disagree.

making small contributions in an unfamiliar repo is *dramatically* easier to get right when you have type annotations and the IDE support they enable providing a guardrail

Issue #2076, opened 2022-03-02
case value not highlighting in C++


Craie:
When listing multiple case values that aren't literals but constants on a single line in C++ (perhaps others?), the value is not highlighted as it is when listing each case and value on a separate line.

An MWE:

```
    switch ( CE )
    {
        case value:
        {
            // code for this value
        } break;
        case value: case value:
        {
            // code for these values
        } break;
        default:
        {
            // code for any other values
        } break;
    }
```

Note:  I'm using pygmentize 2.11.2 under minted 2.5 and tcolorbox 4.22 in LaTeX.  But the situation is the same from the command-line.

amitkummer:
Thank you for the report! This is probably my fault. I introduced some changes to how labels are handled in #2022.

I am going to further investigate this, and also see if I can add some comments explaining the current `label` rule I introduced, as it's a little cryptic.

PR #2084, opened 2022-03-08
CFamily: Lex identifiers after `case` as constants

amitkummer:
Add a state for marking identifiers preceded by  a `case` keyword as constants.

Additionally, refactor the `label` rule to no longer permit a `case` keyword before a label.

Consequentially, identifiers after a `case` keyword (like `foo` in `case foo:`) are no longer wrongly lexed as `Name.Label`, but as
`Name.Constant`.

In addition, this fixes https://github.com/pygments/pygments/issues/2076, as multiple `case` keywords in one line are lexed the same.

### DIFF
--- a/pygments/lexers/c_cpp.py
+++ b/pygments/lexers/c_cpp.py
@@ -52,8 +52,11 @@ class CFamilyLexer(RegexLexer):
              bygroups(using(this), Comment.Preproc), 'if0'),
             ('^(' + _ws1 + ')(#)',
              bygroups(using(this), Comment.Preproc), 'macro'),
-            (r'(^[ \t]*)(?!(?:public|private|protected|default)\b)(case\b\s+)?(' + _ident + r')(\s*)(:)(?!:)',
-             bygroups(Whitespace, using(this), Name.Label, Whitespace, Punctuation)),
+            # Labels:
+            (r'(^[ \t]*)'                                  # Line start and possible indentation.
+             r'(?!(?:public|private|protected|default)\b)' # Not followed by keywords which can be mistaken as labels.
+             r'(' + _ident + r')(\s*)(:)(?!:)',            # Actual label, followed by a single colon.
+             bygroups(Whitespace, Name.Label, Whitespace, Punctuation)),
             (r'\n', Whitespace),
             (r'[^\S\n]+', Whitespace),
             (r'\\\n', Text),  # line continuation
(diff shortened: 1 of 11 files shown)


pygments/lexers/c_cpp.py
```diff
@@ -174,6 +178,13 @@ class CFamilyLexer(RegexLexer):
             # template specification
             (r'\s*(?=>)', Text, '#pop'),
             default('#pop')
+        ],
+        # Mark identifiers preceded by `case` keyword as constants.
+        'case-value': [
+            (r'(:)(?!:)', Punctuation, '#pop'),
```
Jean-Abou-Samra, line 184: Sadly, it's a bit more complicated than that:

```c
  const int a = 1;
  const int b = 2;

  switch (4)
  case (a ? a : b):
  ...
```

amitkummer: Thank you for the quick review!

While this approach isn't perfect, it will rarely fail, with convoluted code that is rarely written. I think that in languages that are hard to lex, like C++, a simple and decent solution should be considered, even if it's not perfect.

I searched for `case` keyword usage in [Juce's source code](https://github.com/juce-framework/JUCE/) (big open source C++ GUI framework, with it's own standard library). After looking at half of all the usages, I failed to see one which is not handled correctly by the proposed solution.

This is just my opinion though, and I am happy to make modifications.

Jean-Abou-Samra: In this case, I think it should not require very complex logic to fix the issue, with something like (untested)

```python
'case-value': [
    (r':(?!:)', Punctuation, '#pop'),
    (r'\(', Punctuation, 'case-expr'),
],
'case-expr': [
    (r'\(', Punctuation, '#push'),
    (r'\)', Punctuation, '#pop'),
    (_namespaced_ident, Name.Constant),
    include('whitespace'),
    include('statements'),
]
```

I agree that lexers needn't be perfect, the fix for this issue just seemed acceptably simple (and the languages common enough) for it to be worth it (maybe I'm missing something though).

amitkummer: That's also not perfect, as the ternary operator doesn't have to be enclosed with parentheses  -- see [this godbolt link](https://godbolt.org/z/YxzY4TxsM) for an example.

I appreciate the help and direction given towards a solution. Unfortunately, a stray colon can appear after a `case` in many ways. This is because since C++17, lambdas are constant expressions, so an immediately invoked lambda can follow a `case` keyword, and inside this lambda, pretty much every syntax can appear.

I created a [short example](https://godbolt.org/z/r4jEavz3j) showcasing this. Colons appear in:

- Line 6 (class definition).
- Line 7 (access modifier).
- Line 8 (bit field).
- Line 12 (ranged for loop).

I think this is a lost battle :(

Jean-Abou-Samra: Ugh, OK. Somehow I was under the impression that the case statement required parentheses (not a C/C++ expert). LGTM then.

Jean-Abou-Samra:
I'll leave this for a few days in case other maintainers would like to take a look.

amitkummer:
I made a small oversight, so I pushed a couple of commits with a fix. I am sorry for the inconvenience, but I think it should be all good now.

Instead of lexing `foo::bar`, as `Name.Constant`, it's now lexed as:

```
foo  Name.Constant
:    Operator
:    Operator
bar  Name.Constant
```

Which is more correct and also consistent with how the lexer treats the `::` operator.

Anteru:
This looks good to me. Thanks for working on this!

PR #2085, opened 2022-03-10
Fix parsing of Isabelle symbols.

relatev:
This PR resolves parsing of Isabelle operators having a ^ in their name, like \\<^sub>. Previously they were not understood as a single token, which prevented the symbols filter from replacing these operators with Unicode.


Jean-Abou-Samra (changes requested):
For tests to pass, you need to regenerate the token output using `pytest --update-goldens tests/examplefiles/isabelle`. Please also add a test or change an existing one to cover this.

Anteru:
This has been fixed by #2159 .

PR #2086, opened 2022-03-12
Add lexer for parsing Uxntal assembly code.

non:
Uxn is a virtual machine which represents a personal computing
playground, and is described here:

    https://100r.co/site/uxn.html

Uxntal assembly langauge is described here:

    https://wiki.xxiivv.com/site/uxntal.html

The demo code piano.tal is used with permission.

### DIFF
--- a/pygments/lexers/tal.py
+++ b/pygments/lexers/tal.py
@@ -1,6 +1,6 @@
 """
-    pygments.lexers.uxntal
-    ~~~~~~~~~~~~~~~~~~~~~~
+    pygments.lexers.tal
+    ~~~~~~~~~~~~~~~~~~~
 
     Lexers for Uxntal
 
--- a/pygments/lexers/tal.py
+++ b/pygments/lexers/tal.py
@@ -4,6 +4,8 @@
 
     Lexers for Uxntal
 
+    .. versionadded:: 2.12
+
     :copyright: Copyright 2006-2022 by the Pygments team, see AUTHORS.
     :license: BSD, see LICENSE for details.
 """
--- a/pygments/lexers/tal.py
+++ b/pygments/lexers/tal.py
@@ -13,7 +13,7 @@
 import re
 import keyword
 
-from pygments.lexer import RegexLexer
+from pygments.lexer import RegexLexer, words
 from pygments.token import Comment, Keyword, Name, String, Number, Punctuation, Whitespace, Literal
 
 __all__ = ['TalLexer']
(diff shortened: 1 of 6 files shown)


pygments/lexers/tal.py
```diff
@@ -0,0 +1,58 @@
+"""
+    pygments.lexers.uxntal
+    ~~~~~~~~~~~~~~~~~~~~~~
```
Jean-Abou-Samra, line 3:
```suggestion
    pygments.lexers.tal
    ~~~~~~~~~~~~~~~~~~~
```

This should in theory make `make check` happy, but it also apparently can't find `pyflakes`, which seems odd …

pygments/lexers/tal.py
```diff
@@ -0,0 +1,58 @@
...
+class TalLexer(RegexLexer):
+
+    name = 'Tal'
+    aliases = ['tal', 'uxntal']
+    filenames = ['*.tal']
+    mimetypes = ['text/x-uxntal']
+
+    tokens = {
+        'comment': [
+            (r'(?:(?<=\s)|^)\((?:(?=\s)|$)', Comment.Multiline, '#push'), # nested comments
```
Jean-Abou-Samra, line 28: You can simplify this regex to `r'(?<!\S)\((?!\S)`. Same on the line below.

pygments/lexers/tal.py
```diff
@@ -0,0 +1,58 @@
...
+            (r'(?:(?<=\s)|^)\)(?:(?=\s)|$)', Comment.Multiline, '#pop'), # nested comments
+            (r'\S+', Comment.Multiline), # comments
+            (r'\s+', Comment.Multiline), # comments
```
Jean-Abou-Samra, line 31: `r'[^()]+'` will be easier to understand and generate less tokens.

pygments/lexers/tal.py
```diff
@@ -0,0 +1,58 @@
...
+        ],
+        'root': [
+            (r'\s+', Whitespace), # spaces
+            (r'(?:(?<=\s)|^)\((?:(?=\s)|$)', Comment.Multiline, 'comment'), # comments
+            (r'(?:(?<=\s)|^)(BRK|LIT|INC|POP|DUP|NIP|SWP|OVR|ROT|EQU|NEQ|GTH|LTH|JMP|JCN|JSR|STH|LDZ|STZ|LDR|STR|LDA|STA|DEI|DEO|ADD|SUB|MUL|DIV|AND|ORA|EOR|SFT)2?k?r?(?:(?=\s)|$)', Keyword.Reserved), # instructions
```
Jean-Abou-Samra, line 36: `words(('BRK', 'LIT', ...), prefix=r'(?<!\S)'`

We have a `words()` function for these kinds of lists, optimizing the regex.

pygments/lexers/tal.py
```diff
@@ -0,0 +1,58 @@
...
+            (r'(?:(?<=\s)|^)[][{}](?:(?=\s)|$)', Punctuation), # delimiers
```
Jean-Abou-Samra, line 37: As above and elsewhere, I would find this simpler with `(?<!\S)` and `(?!\S)`.

pygments/lexers/tal.py
```diff
@@ -0,0 +1,58 @@
...
+            (r'#([0-9a-f]{2}){1,2}', Number.Hex), # integer
+            (r'"\S+', String), # raw string
```
Jean-Abou-Samra, line 39: So there is a delimiter but no string content?

In any case, please make sure this is covered in the test file. Also, normally you'd use a lookahead assertion for the whitespace, to keep it highlighted as `Whitespace` without merging it with the `String` token.

pygments/lexers/tal.py
```diff
@@ -0,0 +1,58 @@
...
+            (r'\'\S', String.Char), # raw char
+            (r'([0-9a-f]{2}){1,2}(?:(?=\s)|$)', Literal), # raw integer
+            (r'\|[0-9a-f]{1,4}(?:(?=\s)|$)', Keyword.Declaration), # abs pad
+            (r'\$[0-9a-f]{1,4}(?:(?=\s)|$)', Keyword.Constant), # rel pad
```
Jean-Abou-Samra, line 43: These two are very similar, please merge them with `[|$]...` (having less rules makes lexers easier to understand).

Jean-Abou-Samra (changes requested):

pygments/lexers/tal.py
```diff
@@ -0,0 +1,58 @@
+"""
+    pygments.lexers.uxntal
+    ~~~~~~~~~~~~~~~~~~~~~~
+
+    Lexers for Uxntal
+
+    :copyright: Copyright 2006-2022 by the Pygments team, see AUTHORS.
```
Jean-Abou-Samra, line 7:
```suggestion
    .. versionadded:: 2.12

    :copyright: Copyright 2006-2022 by the Pygments team, see AUTHORS.
```

Jean-Abou-Samra (changes requested):
One more thing: please add the language to `doc/languages.rst`. You can also add your name in `AUTHORS`.

pygments/lexers/tal.py
non: That's a good point. I'll go through a simplify a bunch of these using this pattern.

pygments/lexers/tal.py
non: Yeah I can improve this to generate fewer tokens.

pygments/lexers/tal.py
non: I'll take a look at that, thanks!

pygments/lexers/tal.py
non: Good catch! This is a bug, it should be `"\s+` (i.e. a quote followed by one or more non-space characters).

non:
@Jean-Abou-Samra Thanks for your thorough review. I think I've addressed all the issues you found (and updated `languages.rst` and `AUTHORS`).

pygments/lexers/tal.py
Jean-Abou-Samra: Ah, sorry -- the `S` character looked close to an `s` in my browser.

pygments/lexers/tal.py
```diff
@@ -0,0 +1,68 @@
...
+    ]
+
+    tokens = {
+        # the comment delimiters must not be adjacent to non-space characters.
+        # this means ( foo ) is a valid comment but (foo) is not. this also
+        # applies to nested comments.
+        'comment': [
+            (r'(?<!\S)\((?!\S)', Comment.Multiline, '#push'), # nested comments
+            (r'(?<!\S)\)(?!\S)', Comment.Multiline, '#pop'), # nested comments
+            (r'(\S[()]|[()]\S|[^()])+', Comment.Multiline), # comments
```
Jean-Abou-Samra, line 42: This is going to fail on the comment

```
( ((( )
```

Please either use lookbehind/lookahead assertions, or (likely simpler) do two rules:

```python
(r'(<?!\S)\((?!\S)', Comment.Multiline, '#push'),
(r'(<?!\S)\)(?!\S)', Comment.Multiline, '#push'),
(r'[^()]+', Comment.Multiline),
(r'[()]', Comment.Multiline),
```

Jean-Abou-Samra (changes requested):

pygments/lexers/tal.py
```diff
@@ -0,0 +1,68 @@
...
+            (r"'\S(?!\S)", String.Char), # raw char
+            (r'([0-9a-f]{2}){1,2}(?!\S)', Literal), # raw integer
+            (r'[|$][0-9a-f]{1,4}(?!\S)', Keyword.Declaration), # abs/rel pad
+            (r'%\S+', Name.Decorator), # macro
+            (r'@\S+', Name.Function), # label
+            (r'&\S+', Name.Label), # sublabel
+            (r'/\S+', Name.Tag), # spacer
+            (r'\.\S+', Name.Variable.Magic), # zero page addr
+            (r',\S+', Name.Variable.Instance), # rel addr
+            (r';\S+', Name.Variable.Global), # abs addr
```
Jean-Abou-Samra, line 60: A question for @birkenfeld and @Anteru: is this kosher? The purpose is to have different colors for the types of addresses, but there's no clean mapping between the usual token types and these ...

pygments/lexers/tal.py
```diff
@@ -0,0 +1,68 @@
...
+"""
+
+import re
+
+from pygments.lexer import RegexLexer, words
+from pygments.token import Comment, Keyword, Name, String, Number, Punctuation, Whitespace, Literal
+
+__all__ = ['TalLexer']
+
+class TalLexer(RegexLexer):
```
Jean-Abou-Samra, line 20: Oh, and the warning when building the docs on your branch reminds me that the lexer class needs a docstring, for https://pygments.org/docs/lexers/. Look at existing docstrings for examples (and don't forget the `.. versionadded::`).

non: Added!

Jean-Abou-Samra:
Thanks for the update. LGTM, except I am still wondering about the remark in the open thread above.

pygments/lexers/tal.py
birkenfeld: Yes, this happens from time to time, and usually I don't think it's a problem.

Jean-Abou-Samra: OK, thanks.

PR #2087, opened 2022-03-13
else|break|next|repeat|in as Name.Function instead of Name

vladbazon:


Jean-Abou-Samra (changes requested):
Since this changes the output, you have to regenerate the tests for these languages with, e.g., `pytest --update-goldens tests/examplefiles/splus`.

pygments/lexers/r.py
```diff
@@ -89,7 +89,7 @@ class SLexer(RegexLexer):
             (r'\[{1,2}|\]{1,2}|\(|\)|;|,', Punctuation),
         ],
         'keywords': [
-            (r'(if|else|for|while|repeat|in|next|break|return|switch|function)'
+            (r'(if|for|while|return|switch|function)'
```
Jean-Abou-Samra, line 92: I am confused. This rule is apparently broken, since `include('valid_name')` comes before `include('keywords')` in the `'statements'` state. Do you really want to highlight these as functions and not as keywords? How about simply moving `include('valid_name')` after `include('keywords')`? Could you explain the reason why `else`, `repeat` etc. are different from `if`, `for`, etc.?

Anteru: I've been looking at the R specification and it does claim that things like "break" are indeed functions in https://cran.r-project.org/doc/manuals/r-release/R-lang.html#Flow-control-elements, but, in other sections they're considered keywords:

* https://cran.r-project.org/doc/manuals/r-release/R-lang.html#Looping
* https://cran.r-project.org/doc/manuals/r-release/R-lang.html#Reserved-words

I'm inclined to not change them to `Function` because they seem quite different from actual functions like `mean` etc.

vladbazon:
>> why else, repeat etc. are different from if, for, etc.

In R 'else'  is a function, but it is not called with "else()" but simply
"else" (without final '(', as recognized by Keyword.Reserved)

for(i in 1:10) {
    if(i < 3) next  # or, next() - which will be then a Keyword.Reserved
    print(i)
    if(i >= 5) break  # or, break()
}

For another example, instead of

repeat({     ### here, 'repeat' will be Keyword.Reserved

    # commands
})

is normally written:
repeat{     ### here, 'repeat' will be a 'Name' token  (and with my
modification, a 'Name.Function' -- which is desired)
    # commands
}


In SLexer  ("r.py") we modified:

        'root': [
            # calls:
            *(r'\b(else|next|break|repeat|in)\b', Name.Function),*
            (r'(%s)\s*(?=\()' % valid_name, Name.Function),
            include('statements'),### ...
        ],

reducing Keyword.Reserved (without 'else', 'repeat' etc.)

        'keywords': [
            (r'(if|for|while|return|switch|function)'
             r'(?![\w.])',
             Keyword.Reserved),
        ],


vladbazon

Leonel020681:
​name​: ​Pygments 
  
 ​on​: ​[push, pull_request] 
  
 ​jobs​: 
 ​  ​build​: 
 ​    ​runs-on​: ​${{ matrix.os }} 
 ​    ​strategy​: 
 ​      ​matrix​: 
 ​        ​os​: ​[ubuntu-latest, windows-latest] 
 ​        ​python-version​: ​["3.6", "3.7", "3.8", "3.9", "3.10"] 
 ​      ​max-parallel​: ​4 
  
 ​    ​steps​: 
 ​    - ​uses​: ​actions/checkout@v2 
 ​    - ​name​: ​Set up Python ${{ matrix.python-version }} 
 ​      ​uses​: ​actions/setup-python@v2 
 ​      ​with​: 
 ​        ​python-version​: ​${{ matrix.python-version }} 
 ​    - ​name​: ​Install package 
 ​      ​run​: ​| 
 ​        python -m pip install --upgrade pip 
 ​        pip install -r requirements.txt 
 ​        pip install . 
 ​    - ​name​: ​Test package 
 ​      ​run​: ​make test TEST=-v 
 ​      ​if​: ​runner.os == 'Linux' 
 ​    - ​name​: ​Test package 
 ​      ​run​: ​pytest 
 ​      ​if​: ​runner.os != 'Linux' 
  
 ​  ​check​: 
 ​    ​runs-on​: ​ubuntu-latest 
 ​    ​steps​: 
 ​    - ​uses​: ​actions/checkout@v2 
 ​    - ​uses​: ​actions/setup-python@v2 
 ​    - ​name​: ​Run make check 
 ​      ​run​: ​make check 
 ​    - ​name​: ​Fail if the basic checks failed 
 ​      ​run​: ​make check 
 ​      ​if​: ​runner.os == 'Linux' 
  
 ​  ​check-mapfiles​: 
 ​    ​runs-on​: ​ubuntu-latest 
 ​    ​steps​: 
 ​    - ​uses​: ​actions/checkout@v2 
 ​    - ​uses​: ​actions/setup-python@v2 
 ​    - ​name​: ​Regenerate mapfiles 
 ​      ​run​: ​make mapfiles 
 ​    - ​name​: ​Fail if mapfiles changed 
 ​      ​run​: ​| 
 ​        if git ls-files -m | grep mapping; then 
 ​          echo 'Please run "make mapfiles" and add the changes to a commit.' 
 ​          exit 1 
 ​        fi 
  
 ​  ​lint​: 
 ​    ​runs-on​: ​ubuntu-latest 
 ​    ​steps​: 
 ​    - ​uses​: ​actions/checkout@v2 
 ​    - ​uses​: ​actions/setup-python@v2 
 ​      ​with​: 
 ​        ​python-version​: ​3.8 
 ​    - ​name​: ​Check out regexlint 
 ​      ​run​: ​git clone https://github.com/pygments/regexlint 
 ​    - ​name​: ​Run regexlint 
 ​      ​run​: ​make regexlint REGEXLINT=`pwd`/regexlint

Jean-Abou-Samra:
@Leonel020681 That's our CI workflow definition. Not sure what you mean by copying it in your comment?

Anteru:
@Jean-Abou-Samra Probably a bot/spam

pygments/lexers/r.py
vladbazon: >>> I'm inclined to not change them to Function because they seem quite different from actual functions like mean etc.
> library(highr)   ## in the R console
> highr:::cmd_latex
                         cmd1 cmd2
COMMENT              \\hlcom{    }
FUNCTION             \\hlkwa{    }
IF                   \\hlkwa{    }
ELSE                 \\hlkwa{    }
WHILE                \\hlkwa{    }
FOR                  \\hlkwa{    }
IN                   \\hlkwa{    }
BREAK                \\hlkwa{    }
REPEAT               \\hlkwa{    }
NEXT                 \\hlkwa{    }
NULL_CONST           \\hlkwa{    }
LEFT_ASSIGN          \\hlkwb{    }

The same highlighting LaTeX command 
\hlkwa{ ...  }
is used for FUNCTION as for ELSE, BREAK, NEXT, IN, REPEAT 

But in SLexer ("r.py") actually highlight only function names (if followed by '(' evidently), not the keyword 'break', 'else' etc.
(except when they are followed by '(' - which is correct, but not usual; we write "for i in 1:5", not "for i in() 1:5")

--vladbazon

Mike86947963 (approved):

PR #2099, opened 2022-03-30
LilyPond: fix 'maybe-subproperties' state for properties containing dashes

Jean-Abou-Samra:
I saw this nasty case in our docs today.


birkenfeld (approved):

Jean-Abou-Samra:
Thanks for reviewing!

PR #2101, opened 2022-04-03
Fix #632.

### LINKED ISSUE #632: code-block table should be wrapped in a `<div>` tag
(Original issue 925 created by None on 2013-11-11T02:14:41.552192+00:00)

In the document of [HtmlFormatter](http://pygments.org/docs/formatters/#htmlformatter), there is an example output snippt when `linenos` option is set to `"table"`. In this example, the `<table>` tag is wrapped in a `<div>` tag with `"highlight"` class.

But current implementation doesn't output the wrapper `<div>`. It generates a bare `<table>` tag with class `"<cssclass>table"`.

Without the wrapping `<div>`, it's hard to styling the table. Say, you cannot easily limit the height of table and make a scroll bar when the number of lines is too large.

I hope that the wrapper `<div>` could be, at least, generated by some configuration.

Anteru:
The doc string indicates that the linenos table is wrapped in `<div class="highlight">`,
but the actual implementation puts the `<div>` inside the table cell containing the code.
This seems to cause issues as explained in #632, and given it doesn't match the
documentation, this PR restores the original behavior.


pygments/formatters/html.py
```diff
@@ -933,16 +933,26 @@ def _highlight_lines(self, tokensource):
             else:
                 yield 1, value
 
-    def wrap(self, source, outfile):
+    def wrap(self, source, *, include_div):
```
Jean-Abou-Samra, line 936: I see `outfile` in a subclass in the documentation example at

https://pygments.org/docs/formatters/#formatter-classes

Is this something to be concerned about?

Anteru: I don't think it should be there given it's unused. I'll update the docs though to match.

Anteru: It looks like it's only used there, and even in the sample code, `outfile` was unused. That said, looking at the wrapping code, I think I'll start a `Html5Formatter` or similar and redo all of this, because the whole table business shouldn't be necessary at all ... but for now the comments are now consistent with the code, and it should only affect users who have subclassed the `HtmlFormatter` _and_ reimplemented `wrap`.

Jean-Abou-Samra:
I can't claim I understand everything in the HtmlFormatter, but what I understand in this patch LGTM.

Anteru:
I think the logic was actually quite broken :/ Second try (I noticed some unit tests would look different by default.)

Jean-Abou-Samra:
Mmmh, I replied to a GitHub notification by email yesterday: "Sorry for dropping the ball. I'll try to review this again tomorrow". Not sure why it didn't reach this PR, maybe it doesn't work for requests for review? Nevertheless, I'm OK with the merge.

Anteru:
:( I didn't get notified, and I figured I'll finally set aside some time to make a release so I didn't want to delay this even further. I think the logic is right now and if not I'll do a 2.12.1 soon.

Anteru:
Glad to hear you got it sorted out, and thanks for taking care of the related issue. If that's all the fallout we'll get I'll be quite happy.

PR #2102, opened 2022-04-03
Improvements to the Savi lexer

jemc:
This PR has several improvements to the Savi lexer:
- Update Savi language example.
- Fix Savi highlighting for underscore/private identifiers.
- Add string interpolation to the Savi lexer. 

See individual commits for details.

Each commit is a self-contained unit of change.

### DIFF
--- a/tests/examplefiles/savi/example.savi
+++ b/tests/examplefiles/savi/example.savi
@@ -9,10 +9,10 @@
   :const describes: "Addition"
 
   :: Return the number 2 (written in hex, just for fun).
-  :const two U64'val: 0x02
+  :const _two U64'val: 0x02
 
   :it "adds two twos"
-    assert: @two + '\x02' == 4
+    assert: @_two + '\x02' == 4
 
   :: Raise an error if the argument is positive.
   :fun non add_overflow!(a U64'val, b U64'val): a +! b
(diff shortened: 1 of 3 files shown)


tests/examplefiles/savi/example.savi
```diff
@@ -23,6 +21,6 @@
     integers Array(U64)'val = [99, 100, 101]
 
     // Check addition overflow for various pairs of addends.
-    @assert = try (add_overflow!(U64.max_value, 1), False | True)
-    @assert = try (add_overflow!(U64.max_value, 0), True | False)
-    @assert = try (add_overflow!(integers[0]!, 1), True | False)
```
Jean-Abou-Samra, line 28: This makes `True` and `False` no longer tested. I think it's more important to test all lexer rules than to have an example that is well-written according to current language standards.

jemc: Please note that `True` and `False` are highlighted according to the same rules as all other capitalized identifiers (such as `Main`, `Spec`, `U64`, and many other such identifiers in this example) - they are not special-case tokens as they are in some languages.

Hence, there is no decrease here in the amount of lexer rules being tested.

Jean-Abou-Samra: Ah, sorry for not looking closely enough.

Jean-Abou-Samra:
Thanks!

PR #2103, opened 2022-04-04
Register .j2 and .jinja2 file suffixes to relevant lexers

### LINKED ISSUE #2100: Support for matching *.j2 filenames
I request that the `.j2` and the `.jinja2` filename suffixes are added to the existing jinja syntax definitions.
In most non-django, non salt-stack systems i've worked with, there is the convention of using `filename.html.j2` or `filename.html.jinja2` for HTML+Jinja2 templates. 

To back my claim: [here](https://github.com/danielchatfield/atom-jinja2/tree/master/grammars) are the language definitions for Atom, and [here](https://github.com/Martin819/sublime-jinja2/tree/45355633d17ee562481ca1d2ad0e3502c238f58a/Syntaxes) are ones for Sublime(whose definitions are used by many other downstream applications), and [here](https://github.com/samuelcolvin/jinjahtml-vscode/blob/master/package.json) are the ones for VSCode.

Here is the current suffixes being, and what I request be added:

* `YAML+Jinja` [matches](https://github.com/pygments/pygments/blob/64e8e05307689fe726ce9df2d5907e9c2fb67405/pygments/lexers/templates.py#L1873) `.sls`, but not `.{yaml,yml}.{j2,jinja2}`
* `HTML+Django/Jinja` [matches](https://github.com/pygments/pygments/blob/64e8e05307689fe726ce9df2d5907e9c2fb67405/pygments/lexers/templates.py#L1254) `.{html,htm,xhtml}` but not `.{html,htm,xhtml}.{j2,jinja2}`
* `XML+Django/Jinja` [matches](https://github.com/pygments/pygments/blob/64e8e05307689fe726ce9df2d5907e9c2fb67405/pygments/lexers/templates.py#L1275) `.xml`, but not `.xml.{j2,jinja2}`
* `CSS+Django/Jinja` [matches](https://github.com/pygments/pygments/blob/64e8e05307689fe726ce9df2d5907e9c2fb67405/pygments/lexers/templates.py#L1296) `.css` but not `.css.{j2,jinja2}`
* `JavaScript+Django/Jinja` [matches](https://github.com/pygments/pygments/blob/64e8e05307689fe726ce9df2d5907e9c2fb67405/pygments/lexers/templates.py#L1315) `.js` but not `.js.{j2,jinja2}`

The current state means that the base HTML, XML, CSS and JS grammars will match first, shadowing the jinja variants, making them something you always have to specify explicitly by other means than through filename-suffixes. 

Downstream issue: https://github.com/Textualize/rich/issues/2018

pbsds:
Fixes #2100


pygments/lexers/templates.py
```diff
@@ -1251,7 +1251,8 @@ class HtmlDjangoLexer(DelegatingLexer):
 
     name = 'HTML+Django/Jinja'
     aliases = ['html+django', 'html+jinja', 'htmldjango']
-    alias_filenames = ['*.html', '*.htm', '*.xhtml']
+    filenames = ['*.html.j2', '*.htm.j2', '*.xhtml.j2', '*.html.jinja2', '*.htm.jinja2', '*.xhtml.jinja2']
+    alias_filenames = ['*.html.j2', '*.htm.j2', '*.xhtml.j2', '*.html.jinja2', '*.htm.jinja2', '*.xhtml.jinja2']
```
Jean-Abou-Samra, line 1255: - Why the duplication between `filenames` and `alias_filenames`?
- Do you really want to remove the `*.html` pattern? Right now, it tries to guess a lexer with `analyse_text` methods. This will make all `*.html` files highlighted by the `HtmlLexer` exclusively.

Jean-Abou-Samra:
Thanks, LGTM.

Jean-Abou-Samra:
Whoops, it appears that I rebased this instead of squash-merging ... Sorry folks.

PR #2106, opened 2022-04-06
Highlight python 3.10's EncodingWarning

kevinastock:
EncodingWarning was added in 3.10: https://docs.python.org/3/library/exceptions.html#EncodingWarning


Jean-Abou-Samra:
Sounds good, thanks!

PR #2107, opened 2022-04-07
Initial commit for MCFunction Lexer + tests

RitikShah:
Hi, I'm a newer open-source contributor so hopefully I've dotted my eyes and crossed my ts.

I'm committing 2 lexers for the `mcfunction` language (and the snbt data format) used in the popular game, Minecraft. This lexer was loosely based off of work done by @Arcensoth [here](https://github.com/Arcensoth/language-mcfunction), which is currently being used to syntax highlight `mcfunction` on Github (and most text editors). This lexer mostly strays away from the line-based lexing approach that's currently on Github, which allows it to work for commands and data to span multiple lines cleanly.

This lexer is designed for the Java Edition of the game (you can read more about the language on the [wiki](https://minecraft.fandom.com/wiki/Commands)), however, it should also function for the Bedrock Edition.

I've included 1 full examplefiles test alongside some key snippets to help cover the entire lexer. Potentially, we can add more to help cover the lexer if needed.

---

I've tried my best to keep a approach to all of the regex, but I'm a little unsure on some of the detailed cases. Particularly, I spun out the `snbt` data format (read more [here](https://minecraft.fandom.com/wiki/NBT_format)), since it's quite useful to highlight this data on it's own, but I ran into some weird handling trying to reuse the same lexer when needed inside the language. This is mostly due to the fact that both JSON and SNBT are very similar, yet, distinct data formats that the game uses (sometimes, inside of each other).

I also had some struggles with the `"selector"` state management, I feel like it could be simplified somewhat. Hopefully, someone else can take a look and provide suggestions.

I appreciate anyone taking a look at this PR! Since Github has `mcfunction` support, it's common to see it used in codeblocks in documentation. I'm hoping this lexer will be useful for those using Sphinx alongside their minecraft project.

Thanks!


pygments/lexers/mcfunction.py
```diff
@@ -0,0 +1,272 @@
...
+
+class SNBTLexer(RegexLexer):
+    """Lexer for stringified NBT, a data format used in Minecraft
+
+
+    .. versionadded:: ?
+    """
+
+    name = "SNBT"
+    url = ""
```
Jean-Abou-Samra, line 28: Could you put an URL here, for consumption by https://pygments.org/languages/? (My first search result is https://minecraft.fandom.com/wiki/NBT_format.)

pygments/lexers/mcfunction.py
```diff
@@ -0,0 +1,272 @@
...
+    aliases = ["snbt"]
+    filenames = ["*.snbt"]
+    mimetype = ["text/snbt"]
+
+    tokens = {
+        "root": [
+            # We only look for the open bracket here since square bracket
+            #  is only valid in NBT pathing (which is a mcfunction idea).
+            (r"(?=\{)", Punctuation, "compound"),
```
Jean-Abou-Samra, line 37: You can just leave this as

```
default("compound"),
```

This is clearer in my opinion: took me a second to realize that the Punctuation token type was not significant as matches are always empty.

pygments/lexers/mcfunction.py
```diff
@@ -0,0 +1,272 @@
...
+
+        "operators": [
+            (r"[,:;]", Punctuation),
+        ],
+
+        "literals": [
+            (r"(true|false)", Keyword.Constant),
+            (r"(-?\d*\.\d+[fFdD]?)", Number.Float),
+            (r"(-?\d+[bBsSlLfFdD]?)", Number.Integer),
+            (r"(-?\d+[eE]-?\d+)", Number.Float),
```
Jean-Abou-Samra, line 52: Does the language not support `1.5e5`?

Jean-Abou-Samra, line 52: Does the language not support `1.5e5`?

pygments/lexers/mcfunction.py
```diff
@@ -0,0 +1,272 @@
...
+            # Seperate states for both types of strings so they don't entangle
+            (r'"', String.Double, "literals.string_double"),
+            (r"'", String.Single, "literals.string_single"),
+        ],
+        "literals.string_double": [
+            (r"\\.", String.Escape),
+            (r'[^\\"\n]+', String.Double),
+            (r'"', String.Double, "#pop"),
+            (r"[$\n]", Error, "#pop"),
```
Jean-Abou-Samra, line 61: So literal strings can't contain a dollar character? Don't you rather mean that they must be matched and not run till the end of the line (`$`)? A `$` character in a character class is escaped, it has the literal meaning of a dollar.

pygments/lexers/mcfunction.py
```diff
@@ -0,0 +1,272 @@
...
+            include("operators"),
+            include("whitespace"),
+            include("literals"),
+            (r"\{", Punctuation, "#push"),
+            (r"\[", Punctuation, "list"),
+            (r"\}", Punctuation, "#pop"),
+        ],
+
+        "list": [
+            (r"(?<=\[)[BIL](?:\;)", Keyword.Type),
```
Jean-Abou-Samra, line 83: What is the (non-capturing) group for?

Jean-Abou-Samra, line 83: What is the (non-capturing) group for?

pygments/lexers/mcfunction.py
```diff
@@ -0,0 +1,272 @@
...
+            (r"[A-z_]+", Name.Attribute),
+            include("literals"),
+            include("operators"),
+            include("whitespace"),
+            (r"\{", Punctuation, "#push"),
+            (r"\{", Punctuation, "compound"),
```
Jean-Abou-Samra, line 89: This rule never matches? Didn't you mean to use a bracket in the previous rule?

Jean-Abou-Samra, line 89: This rule never matches?

pygments/lexers/mcfunction.py
```diff
@@ -0,0 +1,272 @@
...
+        ],
+    }
+
+
+class MCFunctionLexer(RegexLexer):
+    """Lexer for the mcfunction scripting language used in Minecraft
+
+    Modelled somewhat after the Github mcfunction grammar:
+    - "https://github.com/Arcensoth/language-mcfunction
+
```
Jean-Abou-Samra, line 100:
```suggestion
    Modelled somewhat after the `GitHub mcfunction grammar <https://github.com/Arcensoth/language-mcfunction>`_.
```

Jean-Abou-Samra, line 100:
```suggestion
    Modelled somewhat after the `GitHub mcfunction grammar <https://github.com/Arcensoth/language-mcfunction>`_.
```

pygments/lexers/mcfunction.py
```diff
@@ -0,0 +1,272 @@
...
+
+    .. versionadded:: ?
+    """
+
+    name = "SNBT"
+    url = ""
+    aliases = ["snbt"]
+    filenames = ["*.snbt"]
+    mimetype = ["text/snbt"]
```
Jean-Abou-Samra, line 31: Did you mean `mimetypes`? Ditto in the other lexer.

Jean-Abou-Samra, line 31: Did you mean `mimetypes`? Ditto in the other lexer.

pygments/lexers/mcfunction.py
```diff
@@ -0,0 +1,272 @@
...
+            include("operators"),
+            include("delimiters"),
+            include("selectors"),
+        ],
+
+        "names": [
+            # The start of a command (either beginning of line OR after the run keyword)
+            #  We don't encode a list of keywords since mods, plugins, or even pre-processors
+            #  may add new commands, so we have a 'close-enough' regex which catches them.
+            (r"(?=^\s*)([a-z]+)", Name.Builtin),
```
Jean-Abou-Samra, line 129: Just to confirm, a command must be lowercase, right?

The group is unneeded.

pygments/lexers/mcfunction.py
```diff
@@ -0,0 +1,272 @@
...
+            (r"(?<=run\s)([a-z]+)", Name.Builtin),
```
Jean-Abou-Samra, line 130: Not `\s+`?

Jean-Abou-Samra, line 130: Not `\s+`?

pygments/lexers/mcfunction.py
```diff
@@ -0,0 +1,272 @@
...
+            # UUID
+            (
+                r"\b([0-9a-fA-F]+(?:(-)[0-9a-fA-F]+){4})\b",
```
Jean-Abou-Samra, line 133:
```suggestion
                r"\b[0-9a-fA-F]+(?:-[0-9a-fA-F]+){4}\b",
```

pygments/lexers/mcfunction.py
```diff
@@ -0,0 +1,272 @@
...
+        ],
+        "names.fakeplayer": [(r"[\.A-z0-9_]+", Name.Variable.Magic), default("#pop")],
+        "names.maybe-nbt": [
+            (r"\[", Punctuation, "nbt.list"),
+            default("#pop"),
+        ],
+
+        "resource-name": [
+            (
+                r"(#?)([a-z_][a-z_\.\-]*)(\:)([a-z0-9_\.\-\/]+)",
```
Jean-Abou-Samra, line 150: Unneeded groups..

Jean-Abou-Samra (commented):
Here's a first round of review, will take another look later. This looks quite good-quality overall.

pygments/lexers/mcfunction.py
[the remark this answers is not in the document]
RitikShah: Hm, this isn't a default though (I can change Punctuation to Text like the normal Default), I only want the Lexer to continue if it finds a bracket. Or should i use `include("compound")` here? I don't want to lex the other aspects of a compound until we find the very first bracket. Maybe I'm misunderstanding.

pygments/lexers/mcfunction.py
RitikShah: Language does support this, I just completely forgot! Only supports it in valid SNBT though.

pygments/lexers/mcfunction.py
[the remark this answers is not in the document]
RitikShah: Ah, I misunderstood this, yea I need to use something like `r"(\n|$)"` though perhaps `$` is sufficient for making the string error if there is no matching quotation.

pygments/lexers/mcfunction.py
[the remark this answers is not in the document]
RitikShah: Theoretically, commands can be uppercase but there are no situations where that exists (via vanilla or modded). I ended up matching with the Github parser which assumes that commands are lowercase except I should maybe add a `_` here.

```mcfunction
# success
execute as @a run ...

# failure
Execute as @a run ...
```

When you say groups are needed, do u mean the `()`'s?

pygments/lexers/mcfunction.py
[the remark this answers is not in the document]
Jean-Abou-Samra: The way it is currently written, the lexer expects the file to start with a bracket. Otherwise, it will emit error tokens. I thought that brace was a requirement of the language while reading the code. If you want to start at the first brace, you should do something like

```
"root": [
    (r"[^\{]+", Text),
    (r"\{", Punctuation, "compound"),
] 
```

 By the way, can you add some tests for SNBT?

pygments/lexers/mcfunction.py
Jean-Abou-Samra: That's not a big deal -- usually, Pygments does not attempt to validate the input, just to lex it correctly.

pygments/lexers/mcfunction.py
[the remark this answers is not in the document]
Jean-Abou-Samra: Actually, you can leave this rule out completely. If the string is unterminated, no rule will match, and the lexer will start emitting error tokens.

pygments/lexers/mcfunction.py
RitikShah: Ah yes, I should maybe take it out of the group so that It's not included in the look-behind assertion.

pygments/lexers/mcfunction.py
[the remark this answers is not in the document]
Jean-Abou-Samra: Yes.

Jean-Abou-Samra:
There are a bunch of linting errors in the CI (https://github.com/pygments/pygments/runs/5868270378?check_suite_focus=true), be sure to check and fix them (run the tool locally with `REGEXLINT=/path/to/regexlint make regexlint` after cloning https://github.com/pygments/regexlint).

RitikShah:
Thanks so much for these comments. Since you mentioned you wanted to take a closer look, I wasn't sure if I should implement your first-look suggestions or leave it static.

Also, I had some issues with literals in the MCFunctionLexer (particularly floats and numbers). In the commands syntax, we have a coordinate concept which allows you to specify different aspects of relative through preceding operators (the `~` and `^`).

```mcfunction
tp @s 100.5 80 -100.5
tp @s 10 ~ -10
tp @s ~10 ~ ~-10
tp @s 10 ^0.5 -10
tp @s 10 ^-0.5 -10
```

For some reason, my literals and operators doesn't play nice with each other, but I'm not entirely sure why!

Jean-Abou-Samra:
> Since you mentioned you wanted to take a closer look, I wasn't sure if I should implement your first-look suggestions or leave it static.

It'd be helpful for me (and likely for other maintainers if they want to review this as well) if you started doing the updates now.

> For some reason, my literals and operators doesn't play nice with each other, but I'm not entirely sure why!

Will take a look when I review this again.

pygments/lexers/mcfunction.py
```diff
@@ -0,0 +1,272 @@
...
+            (r"\\.", String.Escape),
+            (r"[^\\'\n]+", String.Single),
+            (r"'", String.Single, "#pop"),
+            (r"[$\n]", Error, "#pop"),
+        ],
+
+        "compound": [
+            # this handles the unquoted snbt keys
+            #  note: stringified keys still work
+            (r"[A-z_]+", Name.Attribute),
```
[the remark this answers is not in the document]
RitikShah: Language does support unicode, I'm a bit confused by using `\W` here, doesn't that match non-words?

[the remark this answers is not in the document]
Jean-Abou-Samra: Yeah, it's contorted logic. The standard Python re module does have a direct way to say 'a unicode alphabetic character, excluding numbers and underscores'. All it has is 'unicode alphabetic character or number or underscore' (`\w`). However, you can get 'unicode alphabetic character only' with this trick: 'neither (something else than an alphabetic character or a number or an underscore), nor a number, nor an underscore', i.e. `[^\W\d_]`. To get an underscore with it: `[^\W\d]`. But if the language doesn't support Unicode, you're fine not bothering with this.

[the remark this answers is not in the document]
RitikShah: Mhm, this is tricky. theoretically, "yes", but it must be quoted (i think). U can even have stuff like `.` and `:` as a key here, but you must quote the key, which makes it almost JSON-esque. However, if something is quoted, it'll match as a literal anyways so it should be fine here.

RitikShah (commented):
I think I accidentally started a review, but I have to submit review to get those comments posted?

RitikShah:
I'm working on refactoring the NBT aspect of the lexer. When I use the `SNBTLexer` inside of the `MCFunctionLexer`, I'm unsure how to properly "pop" out.

In my `MCFunctionLexer`, the `root` state includes a `nbt` state which has the following rules:
```py
"nbt": [
    (r"\{", Punctuation, "nbt.compound"),
    (r"\[", Punctuation, "nbt.list"),
],
"nbt.compound": [
    (r".+", using(SNBTLexer, state=("compound",))),
    default("#pop"),
],
"nbt.list": [
    (r".+", using(SNBTLexer, state=("list",))),
    default("#pop"),
],
```

And then my `SNBTLexer` `root` state was changed from a suggestion earlier to:
```py
"root": [
    (r"\{", Punctuation.SNBT.Start, "compound"),   # The token has a special name for debugging
    (r"[^\{]+", Text.TEST),
],
```

With `using`, I pass in a state which starts the SNBTLexer off at a specific state (which works). The issue occurs when the secondary state pops since it puts me back in the `SNBTLexer` `root` state instead of the original `MCFunctionLexer` `nbt.compound` state (which would default pop). I took a look at the method and it looks like if you pass in a string, it automatically puts in the `root` state (this was odd, I feel like it should be in the docstring), so I passed in a tuple with 1-element to try to avoid that but it still occured.

I think I'm not understanding how `using` works which is why I'm unable to solve this. I feel like I want to be able to just push the `compound` state from `SNBTLexer` onto the stack (`(r"\{", Punctuation, SNBTLexer.push_state("compound"))`, like this) but the `using` helper only allows me to substitute a token where I can classify subtokens within one regex rule.

Jean-Abou-Samra:
`using()` is basically intended for parsing embedded code in another language within code in a main language (think <script> in HTML and the like). It's not intended for use cases where you have two similar languages that should share definitions. For that, the primary tool is subclassing. See https://pygments.org/docs/lexerdevelopment/#subclassing-lexers-derived-from-regexlexer.

RitikShah:
Just pushed my latest, this has generic property which treats keys and values the same across every interface of mapping (and lists).

I think this is the best way to handle the various property maps across this strange language. I think I might use the subclassing method to handle the literals the same across the `SNBTLexer` and the `MCFunctionLexer`.

tests/snippets/mcfunction/commenting.txt
```diff
@@ -0,0 +1,293 @@
...
+' '           Text.Whitespace
+'player'      Literal.String.Doc
+'\n'          Text
+
+'#'           Comment.Multiline
+'\n'          Text
+
+'#'           Comment.Multiline
+' '           Text.Whitespace
+'S'           Comment.Multiline
```
birkenfeld, line 66: This looks like a rule could use some repeats.

pygments/lexers/mcfunction.py
```diff
@@ -0,0 +1,326 @@
...
+            (r"#.*$", Comment.Single),
+        ],
+        "comments.block": [
+            (rf"^\s*#{_block_comment_prefix}", Comment.Multiline, "comments.block.emphasized"),
+            (r"^\s*#", Comment.Multiline, "comments.block.normal"),
+            default("#pop"),
+        ],
+        "comments.block.normal": [
+            include("comments.block.special"),
+            (r"\S+?", Comment.Multiline),
```
birkenfeld, line 178: That non-greedy match matches a single letter each time, which is quite inefficient. Is the `?` really necessary?

pygments/lexers/mcfunction.py
```diff
@@ -0,0 +1,326 @@
...
+        ],
+
+        "compound": [
+            # this handles the unquoted snbt keys
+            #  note: stringified keys still work
+            (r"[A-Z_a-z]+", Name.Attribute),
+            include("operators"),
+            include("whitespace"),
+            include("literals"),
+            (r"\{", Punctuation.Compound.Push, "#push"),
```
birkenfeld, line 77: Are these specialized token names really useful?

Also, semantically, there is no difference between `Compound.Push` and `Compound.New`, the distinction only an artifact of how Pygments works.

pygments/lexers/mcfunction.py
```diff
@@ -0,0 +1,326 @@
...
+        "operators": [
+            (r"[,:;]", Punctuation),
+        ],
+
+        "literals": [
+            (r"(true|false)", Keyword.Constant),
+            (r"-?\d*\.\d+[fFdD]?", Number.Float),
+            (r"-?\d+[bBsSlLfFdD]?", Number.Integer),
+            (r"-?\d+[eE]-?\d+", Number.Float),
+            # Seperate states for both types of strings so they don't entangle
```
birkenfeld, line 55:
```suggestion
            # Separate states for both types of strings so they don't entangle
```

pygments/lexers/mcfunction.py
```diff
@@ -0,0 +1,326 @@
...
+        ],
+
+        "operators": [
+            (r"[,:;]", Punctuation),
+        ],
+
+        "literals": [
+            (r"(true|false)", Keyword.Constant),
+            (r"-?\d*\.\d+[fFdD]?", Number.Float),
+            (r"-?\d+[bBsSlLfFdD]?", Number.Integer),
```
birkenfeld, line 53: This will catch the initial part of `10e5` before the rule below does; they must be switched.

pygments/lexers/mcfunction.py
```diff
@@ -0,0 +1,326 @@
...
+        ],
+    }
+
+
+class MCFunctionLexer(RegexLexer):
+    """Lexer for the mcfunction scripting language used in Minecraft
+    
+
+    Modelled somewhat after the `GitHub mcfunction grammar <https://github.com/Arcensoth/language-mcfunction>`_.
+    .. versionadded:: 2.12.0
```
birkenfeld, line 100: an empty line is required before the `.. versionadded`.

pygments/lexers/mcfunction.py
```diff
@@ -0,0 +1,326 @@
...
+            # normal command names and scoreboards
+            #  there's no way to know the differences unfortuntely
+            (r"[A-Za-z_][A-Za-z0-9_.#%$]+", Keyword.Constant),
+            (r"[#%$][A-Za-z0-9_.#%$]+", Name.Variable.Magic),
+        ],
+
+        "resource-name": [
+            (
+                # resource names have to be lowercase
+                r"#?[a-z_][a-z_\.\-]*\:[a-z0-9_\.\-\/]+",
```
birkenfeld, line 145: No need to escape `.` and `/` in charsets, and `-` can be placed at the end. `:` is never special either.
```suggestion
                r"#?[a-z_][a-z_.-]*:[a-z0-9_./-]+",
```
Same below.

birkenfeld:
Just a few more comments - also the mapfile seems to be out of date, so please run `map mapfiles` once more. Thanks!

pygments/lexers/mcfunction.py
RitikShah: This is an artifact of my debugging process (I figured out a better method later), somehow got left in there.

tests/snippets/mcfunction/commenting.txt
RitikShah: I'm not sure what this means, should the tokens for commenting be larger (most text being coupled under a single token versus whitespace separating the words)?

pygments/lexers/mcfunction.py
```diff
@@ -175,13 +172,13 @@ class MCFunctionLexer(RegexLexer):
         ],
         "comments.block.normal": [
             include("comments.block.special"),
-            (r"\S+?", Comment.Multiline),
+            (r".+", Comment.Multiline),
```
birkenfeld, line 175: Hm, now the "comments.block.special" and "whitespace" includes will not match within the comment.

Usually you can handle it like you did in strings: `[everything that doesn't start a special token]+`

RitikShah: Yea, I was thinking about how to get the comment tokens to capture more than just a word, but I forgot to revert it properly (took a lunch break 😅).

RitikShah:
I'm getting closer to a completed lexer here (looks like I still failed tests?). At the moment, I'm mostly worried about my handling of some literals.

I think the route I went with a generic property map was successful, hopefully I did fine with my usage of state (I tried to avoid state inflation by ensuring I popped as I went). I actually used a small debug script in the `get_tokens_unprocessed` to append the state stack to each token (literally just directly using `__getattr__`) so it would be easier to debug. I'm not sure if there was an easier way to trace through the lexer, perhaps it could be interesting to have a debug mode where metadata is attached to the lexer (the current stack perhaps).

RitikShah:
This time, I made sure to pass all of the regex tests and ran `make all`. Hopefully everything turns right this time 😅

triplay-9:
I just cloned `pygments:master` and found that `pygments/lexers/mcfunction.py` is missing. The file disappears in [2.14.0](https://github.com/pygments/pygments/tree/2.14.0/pygments/lexers), but is present in [2.13.0](https://github.com/pygments/pygments/tree/2.13.0/pygments/lexers).

PR #2110, opened 2022-04-12
Fix nested type name highlighting in Savi lexer.

jemc:
Nested type names were being highlighted as if they were function names.

Now a phrase like `Spec.Process.run` is highlighted properly as:
- `Name.Class`
- `Punctuation`
- `Name.Class`
- `Punctuation`
- `Name.Function`

Instead of wrongly highlighted as it was before this commit:
- `Name.Class`
- `Punctuation`
- `Name.Function`
- `Punctuation`
- `Name.Function`

### DIFF
--- a/pygments/lexers/savi.py
+++ b/pygments/lexers/savi.py
@@ -58,8 +58,11 @@ class SaviLexer(RegexLexer):
       # Single-Char String
       (r"'", String.Char, "string.char"),
 
-      # Class (or other type)
+      # Type Name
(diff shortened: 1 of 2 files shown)


jemc:
Not sure what's wrong with the windows-latest 3.10 check, but it would appear to be an unrelated CI failure, because the ubuntu-latest 3.10 check passed, and this PR doesn't have anything platform-specific in it.

Jean-Abou-Samra:
Yeah, it succeeded this time. Merging, thank you.

PR #2112, opened 2022-04-14
Add lexer for colon-separated value config files like /etc/passwd, /etc/shadow and /etc/group

Leistungsabfall:
Hi,

I created two lexers (PasswdLexer, ShadowLexer) for lexing linux config files /etc/passwd and /etc/shadow.
I don't know if they are of much use for pygments though, so feel free to reject this PR if they are too specific.

/etc/passwd:
passwd

/etc/shadow:
shadow


pygments/lexers/configs.py
```diff
@@ -1127,3 +1127,99 @@ def analyse_text(text):
             result += 0.49
 
         return result
+
+
+class PasswdLexer(RegexLexer):
+    """
+    Lexer for /etc/passwd files.
```
Jean-Abou-Samra, line 1134:
```suggestion
    Lexer for ``/etc/passwd`` files.
    
    .. versionadded:: 2.12
```

pygments/lexers/configs.py
```diff
@@ -1127,3 +1127,99 @@ def analyse_text(text):
...
+    """
+
+    name = 'PasswdLexer'
```
Jean-Abou-Samra, line 1137: The name should be the name of the language, not that of the lexer class. I'd use "/etc/passwd".

pygments/lexers/configs.py
```diff
@@ -1127,3 +1127,99 @@ def analyse_text(text):
...
+    aliases = ['passwd']
+    filenames = ['passwd']
```
Jean-Abou-Samra, line 1139: I find this a bit problematic. It will assume any file called "passwd" is in /etc/passwd format. I think "passwd" might be too generic a name to make that assumption. Also, is it really useful? I mean, I'd expect a program that uses Pygments to highlight a file without providing a language explicitly to pass a file that is part of some kind of project (LaTeX document, Git repository, Sphinx documentation or such), not part of the system itself. Could we leave `filenames` empty and let the user explicitly specify to use the `/etc/passwd` lexer on the command line or via the tool they are using?

pygments/lexers/configs.py
```diff
@@ -1127,3 +1127,99 @@ def analyse_text(text):
...
+              bygroups(
+                  Text,
+                  Punctuation,
+                  String,
+                  Punctuation,
+                  Number,
+                  Punctuation,
+                  Number,
+                  Punctuation,
+                  Text,
```
Jean-Abou-Samra, line 1166: Maybe `Comment`?

pygments/lexers/configs.py
```diff
@@ -1127,3 +1127,99 @@ def analyse_text(text):
...
+                  String,
+                ),
+            ),
+        ],
+    }
+
+
+class ShadowLexer(RegexLexer):
+    """
+    Lexer for /etc/shadow files.
```
Jean-Abou-Samra, line 1179:
```suggestion
    Lexer for /``etc/shadow`` files.
    
    .. versionadded:: 2.12
```

pygments/lexers/configs.py
```diff
@@ -1127,3 +1127,99 @@ def analyse_text(text):
...
+    """
+
+    name = 'ShadowLexer'
```
Jean-Abou-Samra, line 1182:
```suggestion
    name = '/etc/shadow'
```

Jean-Abou-Samra (commented):
> I don't know if they are of much use for pygments though, so feel free to reject this PR if they are too specific.


They don't seem egregious to me, but I would prefer if they didn't match files named `passwd` and `shadow`.

birkenfeld:
It seems to me that they could be combined into a single lexer, with highlighting determined not by position but by content...

Leistungsabfall:
Thank you both for your valuable feedback.
I applied the suggested changes and merged both lexers into a generic `UnixConfigLexer` that is able to parse any config file using colons (``:``) as value separators (e.g. ``/etc/shadow``, ``/etc/passwd``, ``/etc/group``).

I'd appreciate if you could take another look at the changes :)

pygments/lexers/configs.py
```diff
@@ -1127,3 +1127,28 @@ def analyse_text(text):
...
         return result
+
+
+class UnixConfigLexer(RegexLexer):
+    """
+    Lexer for Unix/Linux config files using colon-separated values, e.g.
+        * ``/etc/group``
+        * ``/etc/passwd``
+        * ``/etc/shadow``
+
```
Jean-Abou-Samra, line 1138: This should be dedented and separated from the previous paragraph by a blank line in order not to be interpreted as a block quote (reST syntax quirks).

Jean-Abou-Samra (commented):
Sorry that I forgot to say it the first time: this PR is missing some tests. Please add them as `tests/snippets/unixconfig/test-name.txt` and run `pytest --update-goldens tests/snippets/unixconfig` to generate golden test output.

tests/snippets/unixconfig/etc_group.txt
```diff
@@ -0,0 +1,42 @@
...
+':'           Punctuation
+'\n'          Text.Whitespace
+
+'sudo'        Text
+':'           Punctuation
+'x'           Literal.String
+':'           Punctuation
+'1'           Literal.Number
+':'           Punctuation
+'syslog,user\nsyslog' Text
```
Jean-Abou-Samra, line 24: This seems to indicate that the lexer is not working as intended ...

pygments/lexers/configs.py
```diff
@@ -1146,6 +1147,7 @@ class UnixConfigLexer(RegexLexer):
     tokens = {
         'root': [
             (r'^#.*', Comment),
+            (r'\n', Whitespace),
             (r':', Punctuation),
             (r'[0-9]+', Number),
             (r'[a-zA-Z0-9\_\-\s\(\),]{2,}', Text),
```
Jean-Abou-Samra, line 1153: ... because of this, which matches `\n` in `\s`. You probably want `((?!\n)[big group]){2,}`.

Leistungsabfall: Good catch! Looks better now.

Jean-Abou-Samra (commented):
Looks OK to me. Leaving this around in case @birkenfeld wants to take another look.

birkenfeld:
Thanks for the contribution!

PR #2114, opened 2022-04-16
Added a new language "Pseudocode".

xtay2:
This is not added support for generic pseudocode, but rather my own language. See the website here: https://pseudocode.site/

Also I've added the "yield"-keyword for switch-case-labeled-statements in java.


Jean-Abou-Samra:
We discussed this PR between maintainers, and we're sorry to say that we would like to reject it. The reason is twofold:

- "Pseudocode" is an unfortunate name: it's easy to confuse it with actual pseudocode, and also we might want to add actual support for pseudocode in the future,
- This language seems very young and in alpha stage. While rare languages do have a place in Pygments, we expect a language to have a community around it.

I encourage you to submit the commit for Java 'yield' in a separate PR. For 'Pseudocode', a plugin would be better. I think you could make a new PR for 'Pseudocode' with a lexer called something else (pseudocodelang?) when your language has gained some traction. Thanks for your understanding.

PR #2118, opened 2022-04-23
Futhark: add missing tokens

athas:
Adds recognition of 'def'/'type^'/'type~' keywords and adds some
missing characters to the regexes for identifiers and punctuation.


Anteru:
Merged, thanks!

Issue #2119, opened 2022-04-24
Error reading page: format() argument must be a formatter instance, not a class


Lucas-C:
Hi

Since the release of pygments 2.12.0, my `mkdocs`-based documentation build is failing:
https://github.com/PyFPDF/fpdf2/runs/6149548887

Detailed stacktrace:
```
ERROR    -  Error reading page 'Development.md': format() argument must be a formatter instance, not a class
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.10.4/x64/lib/python3.10/site-packages/pygments/__init__.py", line 61, in format
    formatter.format(tokens, realoutfile)
  File "/opt/hostedtoolcache/Python/3.10.4/x64/lib/python3.10/site-packages/pygments/formatter.py", line 94, in format
    return self.format_unencoded(tokensource, outfile)
  File "/opt/hostedtoolcache/Python/3.10.4/x64/lib/python3.10/site-packages/pygments/formatters/html.py", line 981, in format_unencoded
    source = self.wrap(source)
TypeError: BlockHtmlFormatter.wrap() missing 1 required positional argument: 'outfile'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.10.4/x64/bin/mkdocs", line 8, in <module>
    sys.exit(cli())
  File "/opt/hostedtoolcache/Python/3.10.4/x64/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/opt/hostedtoolcache/Python/3.10.4/x64/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/opt/hostedtoolcache/Python/3.10.4/x64/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/opt/hostedtoolcache/Python/3.10.4/x64/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/opt/hostedtoolcache/Python/3.10.4/x64/lib/python3.10/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/opt/hostedtoolcache/Python/3.10.4/x64/lib/python3.10/site-packages/mkdocs/__main__.py", line 192, in build_command
[...]

PR #2123, opened 2022-04-29
Improve heuristic to warn about passing lexer/formatter class

Jean-Abou-Samra:
Don't rely on the error message since 'missing 1 required positional
argument' can give false positives.  Instead, use issubclass().

Addresses part of #2119.


birkenfeld (approved):
Thanks, that's much cleaner.

PR #2124, opened 2022-04-29
Fix most warnings in Sphinx build

Jean-Abou-Samra:
Note that some invalid references are simply removed because the
lexers are all in the same section, so the link is already easy to
find and there is no need to repeat it over and over.

### DIFF
--- a/doc/languages.rst
+++ b/doc/languages.rst
@@ -1,0 +1,5 @@
+:orphan:
+
+This file is overridden by _templates/languages.html and just exists to
+allow the language list to be reliably linked from the documentation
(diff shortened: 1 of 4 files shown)


birkenfeld:
Yeah that's a bit sad - the workaround is to keep `languages.rst` around, not connected to a toctree, and put `:orphan:` into it to shut up the warning about the latter.

Jean-Abou-Samra:
Thanks for the tip! There should be no warnings at all now.

Jean-Abou-Samra:
I move to merge this, since the fixes are rather trivial.

birkenfeld:
Indeed!

PR #2128, opened 2022-05-02
HTMLFormatter: Allow inline line numbers with nowrap (#2127)

### LINKED ISSUE #2127: HTML formatter: `nowrap` should not clobber `linenos=inline` option
Hi there, I noticed that using the HTML formatter with `linenos=inline` and `nowrap` is not possible with `pygmentize`.

It seems like this shouldn't be the case because there's nothing about about the line numbers that justify needing to be wrapped in a `pre` and `div` tag. I'm writing a static site generator in Clojure that allows the user to interop with `pygmentize`, but I don't want the outer `pre` and `div` elements in `pygmentize`'s output.

For example, Chroma allows the user to include inline line numbers without enclosing with a `pre`: `chroma --lexer=clojure --formatter=html --html-only --html-prevent-surrounding-pre --html-inline-styles --style=algol_nu --html-lines /tmp/foo.clj`

I'd be happy to try to tackle this if it seems reasonable.

### Minimal reproduction
`foo.py`
```python
print("hi")
for x in range(5):
    print(x)
```
```
❯ pygmentize -f html -O "nowrap,linenos=inline" -l python /tmp/foo.py
<span class="nb">print</span><span class="p">(</span><span class="s2">&quot;hi&quot;</span><span class="p">)</span>
<span class="k">for</span> <span class="n">x</span> <span class="ow">in</span> <span class="nb">range</span><span class="p">(</span><span class="mi">5</span><span class="p">):</span>
    <span class="nb">print</span><span class="p">(</span><span class="n">x</span><span class="p">)</span>
```
### Version
```
❯ pygmentize -V
Pygments version 2.12.0, (c) 2006-2022 by Georg Brandl, Matthäus Chajdas and contributors.
```

stelcodes:
This patch allows the nowrap option to be used with linenos=inline.
Previously, when nowrap was set linenos=inline was ignored.

Fixes #2127

### DIFF
--- a/pygments/formatters/html.py
+++ b/pygments/formatters/html.py
@@ -967,7 +967,7 @@ class HtmlFormatter(Formatter):
 
         # As a special case, we wrap line numbers before line highlighting
         # so the line numbers get wrapped in the highlighting tag.
-        if not self.nowrap and self.linenos == 2:
+        if self.linenos == 2:
             source = self._wrap_inlinelinenos(source)
 
         if self.hl_lines:
--- a/tests/test_html_formatter.py
+++ b/tests/test_html_formatter.py
@@ -118,6 +118,13 @@ def test_lineanchors_with_startnum():
     html = outfile.getvalue()
     assert re.search("<pre>\\s*<span>\\s*</span>\\s*<a id=\"foo-5\" name=\"foo-5\" href=\"#foo-5\">", html)
 
+def test_linenos_inline_with_nowrap():
+    optdict = dict(linenos="inline", nowrap=True)
+    outfile = StringIO()
+    fmt = HtmlFormatter(**optdict)
+    fmt.format(tokensource, outfile)
+    html = outfile.getvalue()
+    assert re.search("^<span class=\"linenos\">  1", html)
 
 def test_valid_output():
     # test all available wrappers


stelcodes:
@Jean-Abou-Samra that's a great question. I can only speculate, but I think the `nowrap` check was put there because the `linenos=table` method of generating line numbers produces must wrap the code in a `pre` element so line numbers in general were deemed incompatible with the `nowrap` option.
```python
        if not self.nowrap:
            if self.lineanchors:
                source = self._wrap_lineanchors(source)
            if self.linespans:
                source = self._wrap_linespans(source)
            source = self.wrap(source)
            if self.linenos == 1:
                source = self._wrap_tablelinenos(source)
            source = self._wrap_div(source)
            if self.full:
                source = self._wrap_full(source, outfile)
```
This chunk in `HtmlFormatter.format_unencoded` right below the line kind of shows this. The `source = self.wrap(source)` wraps the code in a `pre` tag, then the `table` element is wrapped around that. Finally the whole thing is wrapped in a div, and then optionally a full HTML document.

So I think that there has been an assumption that line numbers and `nowrap` just don't play well together, but with `linenos=inline` this should no longer be the case. I'm wondering if the above chunk could be altered to allow inline line number anchors and spans by further reducing the effect of `nowrap`:
```python
        if self.lineanchors:
            source = self._wrap_lineanchors(source)
        if self.linespans:
            source = self._wrap_linespans(source)
        if not self.nowrap:
            source = self.wrap(source)
            if self.linenos == 1:
                source = self._wrap_tablelinenos(source)
            source = self._wrap_div(source)
            if self.full:
                source = self._wrap_full(source, outfile)
```

Anteru:
Truth to be told, I don't fully understand the logic there either. We have no tests for this so it's hard to tell what the optimal behavior is, but the explanation makes sense, and I don#t see why we should not allow this combination. What I'd like to ask though is to add a test to https://github.com/pygments/pygments/blob/master/tests/test_html_formatter_linenos_elements.py to cover this case as well, so we have "golden test output" I can look at and inspect quickly myself, this has been a life safer already for those kinds of things. It also needs a longer changelog entry to explain in which case the output will differ because from my experience someone somewhere will be relying on this already.

Jean-Abou-Samra:
Ping @stelcodes, do you need help with this?

stelcodes:
Hey @Jean-Abou-Samra, I got distracted with other stuff but I want to keep working on this. I think I understand what I need to addl. I should have some time this weekend.

axljoker03 (approved):

PR #2132, opened 2022-05-09
java multiline highlighting based on issue 1858

VishalN7:
We have implemented what we thought to be multiline support, and would appreciate any feedback regarding implementation/commenting and any supporting documentation you would like from us.


Jean-Abou-Samra (changes requested):
On the code level this looks good to me, but you need to write a test. Please add a file under `tests/snippets/java/` containing a small sample of Java code with a text block. Then use `pytest --update-goldens tests/snippets/java/` in order to generate the output appended to the file. After that, reopen the file and read the token output to check it's correct. If all's well, commit it and run `pytest` to perform a full round of testing. Thanks.

Jean-Abou-Samra:
Thanks! I love this test :-)

Issue #1893, opened 2021-09-08
Lexer request: flex


seshaljain:
[flex](http://dinosaur.compilertools.net/) is a fast scanner generator which accepts regular expressions written in C. The C lexer (almost) works, but some escapes are marked as erroneous syntax.

dman000:
Hello, I'm a student from the University of Adelaide. I’m working with a group of six students looking to contribute towards this project. We are interested in working on this issue if it is still available.

Jean-Abou-Samra:
Sure. If nobody commented on the issue, it is still available. Feel free to take it.

mattv52:
Hi, is there a way to have the lexer not start in the root state?
I would like our new lexer (inheriting from the existing C lexer) to start in a new state and then go to the root state later. Is this possible?
Thanks

Jean-Abou-Samra:
I don't believe this exists, but there is a way around. Look at `SchemeLexer` and `LilyPondLexer` to see how this works. Basically, you can do

```
# C lexer:
tokens: {
    'root': [
        default('actual-root'),
    ],
    'actual-root': [
        # Fill in with what is currently in the 'root' state.
    ],
    # All instances of 'root' in the lexer should become 'actual-root'.
}

# Flex lexer

tokens: {
    'root': [
        # This is the start state of your new lexer.
    ],
    # To 'go to the root state', go to the 'actual-root' state.
}
```

That said, is Flex not grammar rules mixed with C code? Wouldn't `using()` be more appropriate for your purpose?

PR #2134, opened 2022-05-10
Adding Flex Lexer based on issue #1893

dman000:
We have added support for a Flex Lexer based on the issue https://github.com/pygments/pygments/issues/1893.
Any feedback regarding our implementation and correctness of our code would be much appreciated also. Multiple tests are also available for these purposes.


examplec.html
```diff
@@ -0,0 +1,114 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.01//EN"
```
Jean-Abou-Samra, line 1: This PR contains a number of files you used for testing.

- `examplec.html`
- `exampleflex.html`
- `pygments/lexers/aainput.txt`
- `pygments/lexers/out.txt`
- `test.c`
- `test.exe`

They should not be part of the main repository; please remove them from this PR (`git rm --cached examplec.html exampleflex.html ...; git commit`; this leaves them untouched in your working tree, they just disappear from what will eventually be merged).

pygments/lexers/c_like.py
```diff
@@ -1,9 +1,7 @@
 """
     pygments.lexers.c_like
     ~~~~~~~~~~~~~~~~~~~~~~
-
```
Jean-Abou-Samra, line 4: Where does this change and the ones below come from? Is your editor stripping empty lines automatically? Please add these empty lines back; some of them are required by reST syntax.

pygments/lexers/c_like.py
```diff
@@ -663,3 +649,56 @@ class OmgIdlLexer(CLexer):
             include('annotation_appl'),
         ],
     }
+class FlexLexer(CLexer):
+    """
+    Lexer for Flex <http://dinosaur.compilertools.net/flex/index.html>,
```
Jean-Abou-Samra, line 654:
```suggestion
    Lexer for `Flex <http://dinosaur.compilertools.net/flex/index.html>`_, 
```

to make it a real hyperlink.

pygments/lexers/c_like.py
```diff
@@ -663,3 +649,56 @@ class OmgIdlLexer(CLexer):
             include('annotation_appl'),
         ],
     }
+class FlexLexer(CLexer):
+    """
+    Lexer for Flex <http://dinosaur.compilertools.net/flex/index.html>, 
+    A fast scanner generator.
```
Jean-Abou-Samra, line 655:
```suggestion
    a fast scanner generator.
```

pygments/lexers/c_like.py
```diff
@@ -663,3 +649,56 @@ class OmgIdlLexer(CLexer):
...
+    .. versionadded:: 2.5
```
Jean-Abou-Samra, line 656:
```suggestion

    .. versionadded:: 2.5
```

Here you need an empty line, or the directive block won't be recognized.

You can check the output by running `make html` in the `doc/` directory and opening `_build/html/docs/lexers.html`.

tests/examplefiles/flex/example.l
```diff
@@ -0,0 +1,79 @@
+/* scanner for a toy Pascal-like language */
```
Jean-Abou-Samra, line 1: For all these tests, you also need to add test output. This is how we catch regressions: the testing infrastructure compares the output of Pygments against the expected output. To generate this, run `pytest --update-goldens tests/examplefiles/flex` and commit the resulting `.output` files.

Jean-Abou-Samra (changes requested):
Some preliminary comments; I need to go do something else, so I'll review the actual lexer later today.

pygments/lexers/c_like.py
```diff
@@ -663,3 +649,56 @@ class OmgIdlLexer(CLexer):
...
+    """
+    name = 'Flex'
+    aliases = ['flex']
+    filenames = ['*.lex', '*.l']
+    mimetypes = ['text/x-lex']
+
```
Jean-Abou-Samra, line 662: By the way: you also need

```
url = 'http://...'
```

pygments/lexers/c_like.py
```diff
@@ -663,3 +649,56 @@ class OmgIdlLexer(CLexer):
             include('annotation_appl'),
         ],
     }
+class FlexLexer(CLexer):
```
birkenfeld, line 652: leave 2 empty lines between classes

pygments/lexers/c_like.py
birkenfeld: Actually the hyperlink should go to the `url` attribute of the class. `https://github.com/westes/flex` is probably a better choice, since it's not from 1995 :)

pygments/lexers/c_like.py
birkenfeld: also it would be 2.13 now.

pygments/lexers/c_like.py
```diff
@@ -663,3 +649,56 @@ class OmgIdlLexer(CLexer):
...
+        # rules for rules section
+        'rules': [
+            include('whitespace'),
+            (r'%%', String.Delimiter, '#pop'),
+            (r'\\[abfnrtv]', String.Char),
+            include('root')
+        ],
+        'keywords': [
+            (words((
+                'yytext','yylmax','yyleng','yyin','yylineno','yy_current_buffer', 'yy_break','yy_buffer_state','yy_flex_debug','yy_end_of_buffer_char','yy_user_action','yy_user_init','yy_act','yy_interactive','yy_start','yy_decl','yy_flex_major_version', 'yy_flex_minor_version',
```
birkenfeld, line 680: please break lines at 80-120 chars and keep spaces after comma.

erickhartawan (commented):
Based on the PR comment by Jean-Abou-Samra

Jean-Abou-Samra:
> Based on the PR comment by Jean-Abou-Samra

What do you mean? When there are several reviews, you need to address all of them :-)

examplec.html
Jean-Abou-Samra: @erickhartawan I see a commit from you entitled "[Removed local testing files](https://github.com/pygments/pygments/pull/2134/commits/0f1f927265258a0b47bbd36ee5625e5661b2f064)", but it doesn't actually remove the files.

pygments/lexers/c_like.py
```diff
@@ -663,3 +649,53 @@ class OmgIdlLexer(CLexer):
...
+    .. versionadded:: 2.5
+    """
+    name = 'Flex'
+    aliases = ['flex']
+    filenames = ['*.lex', '*.l']
+    mimetypes = ['text/x-lex']
+
+    tokens = {
+        # root will be for definitions section and user code (through inherit mainly)
+        'root': [
```
Jean-Abou-Samra, line 665: I would recommend restructuring this lexer. It is using the `CLexer` base class to highlight most things, which means this for example will unavoidably be highlighted incorrectly

```
%%

while { rule_action(); }
```

`while` is a C language keyword, but here it should be recognized as a pattern (`String.Regex`). 

Instead, I would use the trick at https://github.com/pygments/pygments/issues/1893#issuecomment-1111764791 (for lack of something better for the time being; it would be nice ultimately if we had a way to say "start using this other lexer, and hand back to me when it pops all of its stack").

mattv52: So you would suggest to have `FlexLexer` inherit from `RegexLexer` and then use `using()` to handle the C code sections?

pygments/lexers/c_like.py
```diff
@@ -663,3 +649,53 @@ class OmgIdlLexer(CLexer):
...
+            )), Name.Functions),
+            inherit,
+        ],
+        'statements': [
+            (r'\[:(alnum|alpha|blank|cntrl|digit|graph|lower|print|punct|space|upper|xdigit):\]', Name.Builtin),
+            (r'\n', Whitespace, '#pop'),
+            (r'[a-zA-Z]\-[a-zA-Z]', String),
+            (r'\\[ntr]',String.Escape),
+            (words((
+                'if','then','begin','end','procedure','function'), suffix=r'\b'),
```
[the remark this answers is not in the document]
mattv52: Does the `RegexLexer` actually correctly highlight regular expressions or does it just provide us with the ability to use regular expressions while lexing?

pygments/lexers/c_like.py
```diff
@@ -663,3 +649,53 @@ class OmgIdlLexer(CLexer):
...
+            (r'\\[abfnrtv]', String.Char),
+            include('root')
+        ],
+        'keywords': [
+            (words((
+                'yytext','yylmax','yyleng','yyin','yylineno','yy_current_buffer', 'yy_break','yy_buffer_state','yy_flex_debug','yy_end_of_buffer_char','yy_user_action','yy_user_init','yy_act','yy_interactive','yy_start','yy_decl','yy_flex_major_version', 'yy_flex_minor_version',
+            )), Name.Builtin),
+            (words((
+                'yylex', 'yymore', 'yyterminate', 'yy_input', 'yyless', 'yycopy', 'yyrestart', 'yy_scan_string', 'yy_scan_bytes', 'yy_scan_buffer', 'yywrap',  'yy_set_interactive','yy_set_bol','yy_at_bol','yy_create_buffer','yy_delete_buffer','yy_flush_buffer', 'yy_flex_debug','yy_init_buffer','yy_flush_buffer','yy_load_buffer_state','yy_switch_to_buffer','yy_pop_state','yy_push_state','yy_top_state','yyFlexLexer'
+            )), Name.Functions),
```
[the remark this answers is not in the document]
erickhartawan: Sorry for the late reply. Name.Functions was a typo.  The rationale behind the differentiation is Name.Builtin -> for builtin attributes and Name.Function is for builtin function. Do you think we should just put everything under Name.Builtin? Thanks so much for the guidance and comment. Really appreciate it and we will try to fix everything.

examplec.html
erickhartawan: I think I made a mistake. will get it fixed soon!

PR #2136, opened 2022-05-11
nim: types shouldn't be highlighted as keywords

narimiran:
This was requested by the author of Nim himself: @araq.

### DIFF
--- a/pygments/lexers/nimrod.py
+++ b/pygments/lexers/nimrod.py
@@ -87,10 +87,10 @@ class NimrodLexer(RegexLexer):
             (r'(%s)\b' % underscorize(opWords), Operator.Word),
             (r'(p_?r_?o_?c_?\s)(?![(\[\]])', Keyword, 'funcname'),
             (r'(%s)\b' % underscorize(keywords), Keyword),
-            (r'(%s)\b' % underscorize(['from', 'import', 'include']),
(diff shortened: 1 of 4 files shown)


Jean-Abou-Samra:
Could you explain the rationale behind this change? Keyword.Type isn't highlighted the same as Keyword in the default style, e.g.:

image


If you are using another style that doesn't have a rule for Keyword.Type, it might be a matter of adding it.

Now, another reason for changing to Name would be if these aren't really reserved keywords (like type names are in, e.g., Python -- they're just built-in objects in the default namespace, they can be rebound if you really want to), but in that case wouldn't Name.Builtin be a better fit?

Jean-Abou-Samra:
Could you explain the rationale behind this change? Keyword.Type isn't highlighted the same as Keyword in the default style, e.g.:

image


If you are using another style that doesn't have a rule for Keyword.Type, it might be a matter of adding it.

Now, another reason for changing to Name would be if these aren't really reserved keywords (like type names are in, e.g., Python -- they're just built-in objects in the default namespace, they can be rebound if you really want to), but in that case wouldn't Name.Builtin be a better fit?

birkenfeld:
I'd like to add that you can see in the diff view that Github also colors the types...

PR #2140, opened 2022-05-17
Fix comments within function declarations in C (#1891)

### LINKED ISSUE #1891: C-language code-block lexing error with specific comment syntax "//@"
I'd reported this originally here - https://github.com/sphinx-doc/sphinx/issues/9548 - but it turns out that Sphinx is using pygments highlighter which independently shows the same behaviour I reported there. So copying the issue over.

## Describe the bug

The following code block causes a C-language "lexing error" to be reported and highlighting skipped.

```
.. code-block :: c

    int div(int x, int y)
    //@requires y >= 0;
    {
        return x / y;
    }
```

On the other hand, if I just remove the trailing semicolon on the comment line, the error goes away. i.e. the below version lexes fine.

```
.. code-block :: c

    int div(int x, int y)
    //@requires y >= 0
    {
        return x / y;
    }
```

.. and the following lexes fine too -

```
.. code-block :: c

    int div(int x, int y)
    // requires y >= 0;
    {
         return x / y;
    }
```

Since the line in question is a comment line, what characters feature within the comment line should not be of any consequence for lexing the block as "C".

## How to Reproduce

Include the example code blocks shown in any rst documentation processed by Sphinx. The following note by @jakobandersen shows how to reproduce it using `pygmentize`.

> Thanks for reporting, though the lexing and highlighting is under the hood done with Pygments. It seems that running 
> `pygmentize -l c -F raiseonerror test.c` emulates what Sphinx does, and for your first example I indeed get an error:

```
*** Error while highlighting:
pygments.filters.ErrorToken: @
   (file "/home/jla/.local/lib/python3.8/site-packages/pygments/filters/__init__.py", line 782, in filter)
*** If this is a bug you want to report, please rerun with -v.
int div(int x, int y)
```

## Expected behavior

All the indicated C-language code blocks should lex fine since the line in question is a comment line.

lambda-karlculus:
Fixes #1891 and other combinations of comments within function declarations in C.

Comments were not properly detected between the parts of function declarations. This was fixed by detecting those comments and passing them to a minimal lexer. A new test snippet file was created to test the fix.


pygments/lexers/c_cpp.py
```diff
@@ -19,6 +19,21 @@
 __all__ = ['CLexer', 'CppLexer']
 
 
+class CFamilyComments(RegexLexer):
```
Jean-Abou-Samra, line 22: No need for a separate class; just do `using(this, state='comments')`, and add this as `'comments'` state in the `CFamily` lexer. That way, some things like subclassing are easier.

pygments/lexers/c_cpp.py
```diff
@@ -118,19 +133,25 @@ class CFamilyLexer(RegexLexer):
             include('keywords'),
             # functions
             (r'(' + _namespaced_ident + r'(?:[&*\s])+)'  # return arguments
+             r'((?:(?://.*?\n)|(?:/[*][\w\W]*?[*]/)|\s*?)*)'    # possible comments
```
Jean-Abou-Samra, line 136: This regex has catastrophic backtracking. Try matching input like

```
int
                                                                                    // lots of spaces
+++ // whoops, invalid
```


It should take a loooong time. See `CONTRIBUTING.md` for more explanation. Then change this to

```
r'((?:((?://.*?\n)|(?:/[*](.|\n)*?[*]/))\s*)*)'
```

Also, please use a variable to avoid repeating this regex below.

Jean-Abou-Samra (changes requested):
Sorry for the delay. Sounds good, a few comments and we should be good to go.

pygments/lexers/c_cpp.py
amitkummer: +1 for using a variable. Additionally, @lambda-karlculus, is there a reason for not using the existing regular expressions for comments?
https://github.com/pygments/pygments/blob/81f1669f43cf24a33e720317b3f9b60991636d4e/pygments/lexers/c_cpp.py#L78-L79

amitkummer:
Also, I think that fixing this minor issue by adding this much logic to an already complex part of the lexer is uncalled for, and can introduce maintenance problems. Please simplify this @lambda-karlculus.

Jean-Abou-Samra:
> Also, I think that fixing this minor issue by adding this much logic to an already complex part of the lexer is uncalled for, and can introduce maintenance problems. Please simplify this @lambda-karlculus.

IMHO, with `using(this, 'comments')` and a variable for the regex, it will be OK.

pygments/lexers/c_cpp.py
lambda-karlculus: I made this change (and the changes requested below). However now the tests are failing, with error:
`TypeError: get_tokens_unprocessed() got an unexpected keyword argument 'stack'.`

I suspect it is because the possible comment regex may receive no input (empty string). I cannot seem to match to the empty string in a sensible way. Any advice on how to resolve this would be greatly appreciated.

Jean-Abou-Samra:
OK, I went ahead and changed your PR directly (don't expect me to do that every time :smile:).

- Fixed the basic failure: `using(this, state=...)` requires that the `get_tokens_unprocessed` method accept a `stack` optional argument defaulting to `('root',)`. (Adjusted in subclasses too.)
- Fixed some test failures in `tests/examplefiles/freefem`. This was super tricky. The comment regexes used things like `[\w\W]*?` for the inner content of the comment. For a *standalone* regex, that is fine: if you run it on a comment, it will stop at the end of the comment because `*?` is non-greedy. But *embedded* within another regex, it can cause trouble, because if the part of the larger regex after this one doesn't match with the smallest possible comment, it backtracks and `*?` starts matching more. It was marking huge swaths of code as comments …
- The catastrophic backtracking in `_possible_comments` was still there! :smile: I've fixed it.
- Also simplified: the `'comments'` state was actually useless because it was a superset of the `'whitespace'` state.

@amitkummer I think this is good to go. You have more experience than me tinkering with the `CFamilyLexer`, so I'll leave this for a while in case you want to take another look.

lambda-karlculus:
Thank you! That's amazing!

- Thanks for explaining the backtracking and errors in the regex. I tried to reuse the existing regexes, but understand now why that does not work.
- I agree that the `'comments'` state is just a subset of `'whitespace'` and should be removed.
- The first point (that `using(this, state=...)` requires `get_tokens_unprocessed` accept the `stack` argument) is interesting. It is not clear from the documentation that the former requires the latter. Unless I'm wrong here I may send a new pull request to add that to the docs.
- I noticed you made a change to stop "separating newline tokens". This was the existing behavior. Should this be changed in the `'whitespace'` state?

Jean-Abou-Samra:
> * 
> 
> * I noticed you made a change to stop "separating newline tokens". This was the existing behavior. Should this be changed in the `'whitespace'` state?

I did that one as a simplification before I realized that the `'comments'` state could be dropped entirely. As far as I can see, the `'whitespace'` state does this in order to be able to use `^` (it needs to stop at newlines for that). Well, it looks like it could also be refactored in a simpler way not requiring this, but let's not mix too many changes in this PR.

amitkummer:
Looks excellent now @Jean-Abou-Samra, Thanks!

Jean-Abou-Samra:
Thanks for reviewing, merging then.

PR #2142, opened 2022-05-20
Add value modifier for Kotlin

dellisd:
This just adds the [`value` modifier](https://kotlinlang.org/docs/inline-classes.html) for Kotlin.


Jean-Abou-Samra:
Thanks!

PR #2143, opened 2022-05-23
Fix pr

mattv52:


Anteru:
This is missing the golden tests outputs which you can generate using `pytest --update-goldens`. That said, there are error tokens generated as it looks like the transition to the `CLexer` is not correct and the `CLexer` fails to classify some tokens. I didn't have time to investigate this further, but that'll be the next problem you'll run into once you try to create the golden test files. Sorry it took so long to get to this :/

axljoker03 (approved):

Issue #2147, opened 2022-05-25
SQL+Jinja lexer


edgarrmondragon:
For things like Airflow, dbt, etc.

PR #2148, opened 2022-05-25
feat: add a new SqlJinjaLexer

edgarrmondragon:
I tried following the instructions in https://pygments.org/docs/lexerdevelopment/. Let me know if I'm missing anything 🙂 

Closes https://github.com/pygments/pygments/issues/2147

### DIFF
--- a/pygments/lexers/templates.py
+++ b/pygments/lexers/templates.py
@@ -18,6 +18,7 @@ from pygments.lexers.python import PythonLexer
 from pygments.lexers.perl import PerlLexer
 from pygments.lexers.jvm import JavaLexer, TeaLangLexer
 from pygments.lexers.data import YamlLexer
+from pygments.lexers.sql import SqlLexer
 from pygments.lexer import Lexer, DelegatingLexer, RegexLexer, bygroups, \
     include, using, this, default, combined
 from pygments.token import Error, Punctuation, Whitespace, \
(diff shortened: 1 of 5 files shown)


pygments/lexers/templates.py
```diff
@@ -2264,3 +2266,18 @@ class Angular2HtmlLexer(DelegatingLexer):
...
+class SqlJinjaLexer(DelegatingLexer):
+    """
+    Templated SQL lexer.
+
+    .. versionadded:: 2.13
+    """
+
+    name = 'SQL+Jinja'
+    aliases = ['sql+jinja']
+    filenames = ['*.sql']
```
Jean-Abou-Samra, line 2280: This is problematic because the plain `SqlLexer` already declares `*.sql`. If you want this, you need to add an `analyse_text()` method and check the one defined by `SqlLexer` so that they will distinguish.

Also, the other Jinja2 lexers have `*.xxx.j2` and `*.xxx.jinja2` filenames declared. I'd add these.

edgarrmondragon: Hi @Jean-Abou-Samra! Thanks for the review and suggestion. I've implemented `analyse_text` and added some unit tests too.

Jean-Abou-Samra:
Fails CI, you need to redo ‘make mapfiles’.

edgarrmondragon:
@Jean-Abou-Samra I've updated the map files

Jean-Abou-Samra:
Thanks.

PR #2152, opened 2022-05-29
Merge mapping file generation scripts

Jean-Abou-Samra:
Use a unified script, to reduce code duplication and in preparation
for doing a similar thing with styles and filters.  The new script
also uses a bit more modern Python APIs (e.g., pathlib).

Unlike the previous scripts, it does not replace replace CRLF with LF
because Git should do that itself.

--- 

I'd like to introduce a new plugin API, and the patch is getting somewhat large, so I'm trying to get some parts of it out of the way.


scripts/gen_mapfiles.py
```diff
@@ -0,0 +1,47 @@
...
+import re
+import sys
+
+top_src_dir = Path(__file__).parent.parent
+pygments_package = top_src_dir / 'pygments'
+sys.path.insert(0, pygments_package.parent.resolve())
+
+from pygments.util import docstring_headline
+
+for key in ['lexers', 'formatters']:
```
Anteru, line 22: I'd prefer to move everything here into a function and then do the usual `if __name__ == '__main__':` check, any reason why you're doing this all of this directly?

scripts/gen_mapfiles.py
```diff
@@ -0,0 +1,47 @@
...
+    for file in (pygments_package / key).glob('[!_]*.py'):
+        module_name = '.'.join(file.relative_to(pygments_package.parent).with_suffix('').parts)
+        print(module_name)
+        module = import_module(module_name)
+        for obj_name in module.__all__:
+            obj = getattr(module, obj_name)
+            desc = (module_name, obj.name, tuple(obj.aliases), tuple(obj.filenames))
+            if key == 'lexers':
+                desc += (tuple(obj.mimetypes),)
+            else: # key == 'formatters'
```
Anteru, line 33: I'd prefer a `elif key == 'formatters'` and explicit error handling otherwise. Yes it's stupid because there are only two types here, but if someone ever adds something to the list, this branch will cause trouble instead of failing early.

Also, shouldn't this be `if` in both cases, i.e.:
```python
if key == 'lexers':
  # add lexer specific data to desc
if key == 'formatters':
  # add formatter specific data to desc
```

(Also also, I wish we could require Python 3.10 here and use `match` ...)

Anteru (approved):
This is a good change to remove duplicated code and consolidate functionality. I have only very minor comments, feel free to merge without changes if you don't agree.

Jean-Abou-Samra:
Thanks for the quick review! I've addressed your comments.

scripts/gen_mapfiles.py
Jean-Abou-Samra: Done, but with `elif` since otherwise I don't have an `else` to raise an error.

Anteru (approved):

Anteru:
Thanks for the quick changes. Feel free to merge.

birkenfeld:
Very nice simplification, thanks!

PR #2153, opened 2022-05-29
Remove filename pattern caches

Jean-Abou-Samra:
These are irrelevant because the fnmatch module implementation already
uses functools.lru_cache.

(More diff size reduction towards a new plugin system.)


birkenfeld (approved):
Nice!

PR #2154, opened 2022-05-29
Support building docs when some plugins are installed

Jean-Abou-Samra:
Pass plugins=False to get_all_lexers, otherwise we try to document
them, which fails. For styles, use STYLE_MAP directly; given that I'm
trying to introduce a new plugin interface, I'd rather not change
get_all_styles to support plugins=False right now.


Jean-Abou-Samra:
@Anteru @birkenfeld thoughts on this one? I'd like to merge it because it fixes the doc build for anyone having plugins installed in their environment.

birkenfeld:
LGTM.

PR #2156, opened 2022-05-29
Update token docs

Jean-Abou-Samra:
Update status of Whitespace, which we're now trying to use for all
whitespace.  Explain that a token name must start with a capital
letter.  Also remove a few references to ancient versions.


birkenfeld (approved):
LGTM!

PR #2157, opened 2022-06-02
Add Python 3.11 in CI

Jean-Abou-Samra:
Python 3.11 is in the beta phase (3.11b3 released yesterday, although this is conceptually b2 because they expedited b3 due to an incompatibility with pytest -- https://mail.python.org/archives/list/python-dev@python.org/thread/VZ222BFFAWW73DYFCR7NZR2C5M3TXD65/). I propose to hit the ground running by adding it to the CI already now.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10", "3.11"]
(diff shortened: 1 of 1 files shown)


Anteru:
This seems to be too close to the bleeding edge as CI fails?

Jean-Abou-Samra:
Looks like it's not finding it, I suspect it wants a non-beta version. I've pushed a tentative fix, let's see how that goes. (I tested 3.11.0b3 locally and it works just fine.)

Jean-Abou-Samra:
Ah, this still fails because `lxml` doesn't have pre-built wheels for 3.11 yet, so it needs a compiler to build them ... Closing, I guess it's a bit early for this.

PR #2160, opened 2022-06-12
Detect true-color terminal support for pygmentize

micahcochran:
For the `pygmentize` command line, this pull request adds detection of true-color support on the terminal and makes that the default formatter.  It looks up the `COLORTERM` environment variable having a value of `truecolor` or `24bit`. If that is the case, it sets the default formatter to use the `TerminalTrueColorFormatter`.

The `COLORTERM` environment variable is a convention that has been adopted by some terminals for specifying true-color is supported.

### DIFF
--- a/doc/docs/cmdline.rst
+++ b/doc/docs/cmdline.rst
@@ -13,6 +13,18 @@ You can use Pygments from the shell, provided you installed the
 will print the file test.py to standard output, using the Python lexer
 (inferred from the file name extension) and the terminal formatter (because
 you didn't give an explicit formatter name).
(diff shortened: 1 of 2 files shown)


Jean-Abou-Samra (changes requested):
Sounds good, but you need to update the documentation in `doc/docs/cmdline.rst`.

micahcochran:
I added a second commit, which documents how the terminal determines the maximum number of colors supported.

Jean-Abou-Samra (approved):

Jean-Abou-Samra:
Thanks!

PR #2165, opened 2022-06-19
Teraterm updates

kurtmckee:
This patch:

* Tokenizes whitespace as `Text.Whitespace`, to help address #1905.
* Tokenizes strings in their entirety, rather than one character at a time.
* Removes backslash escape sequences for string literals. They're not special in Tera Term string literals and I have no idea why I added those in the first place!
* Updates tox.ini to drop testing for CPython 3.5 and PyPy3. This now matches CI.

Related to #1905


birkenfeld:
Why did you drop PyPy from tox.ini?

PR #2168, opened 2022-06-29
Add StarOfficeStyle

flywire:
StarOffice style is a theme used in the user interface and carries through to OpenOffice and LibreOffice, and the documentation.

Reference:
* [Color Config Entry](https://github.com/LibreOffice/core/blob/master/svtools/source/config/colorcfg.cxx#L447-L453)
* [Color Types](https://github.com/LibreOffice/core/blob/e4a57dcdabc9ae7d381025e008b90635c1b7b10c/include/tools/color.hxx#L448)

### DIFF
--- a/pygments/styles/staroffice.py
+++ b/pygments/styles/staroffice.py
@@ -1,0 +1,15 @@
+from pygments.style import Style
+from pygments.token import Keyword, Name, Comment, String, Error, \
+     Number, Operator, Generic
+
+class StarofficeStyle(Style):
+    default_style = ""
+    styles = {
+        Comment:                '#808080',   # Gray
+        Error:                  '#800000',   # Red
+        Keyword:                '#000080',   # Blue
+        Name:                   '#008000',   # Green
+        Number:                 '#FF0000',   # Lightred
+        Operator:               '#000080',   # Blue
+        String:                 '#FF0000',   # Lightred
+    }


pygments/styles/staroffice.py
```diff
@@ -0,0 +1,28 @@
+"""
```
Jean-Abou-Samra, line 1: You need to add this style to the `STYLE_MAP` in `pygments/styles/__init__.py`. It won't be found without this. (Yes, it's inconsistent that we don't use an autogenerated mapping file for that like lexers and formatters.)

pygments/styles/staroffice.py
```diff
@@ -0,0 +1,28 @@
...
+
+    Style similar to StarOffice style, also OpenOffice and LibreOffice.
+
+    :copyright: Copyright 2006-2022 by the Pygments team, see AUTHORS.
+    :license: BSD, see LICENSE for details.
+"""
+
+from pygments.style import Style
+from pygments.token import Keyword, Name, Comment, String, Error, \
+     Number, Operator, Generic
```
Jean-Abou-Samra, line 13:
```suggestion
     Number, Operator

```

pygments/styles/staroffice.py
```diff
@@ -0,0 +1,28 @@
...
+    """
+    Style similar to StarOffice style, also OpenOffice and LibreOffice.
+    """
+    default_style = ""
+    styles = {
+        Comment:                '#808080',   # Gray
+        Error:                  '#800000',   # Red
+        Keyword:                '#000080',   # Blue
+        Name:                   '#008000',   # Green
+        Number:                 '#FF0000',   # Lightred
```
Jean-Abou-Samra, line 25: That's not "light" red, is it? The `#800000` marked "red" above is lighter ... Did you mix them up?

flywire: Colors are correct, they match [Color Types](https://github.com/LibreOffice/core/blob/e4a57dcdabc9ae7d381025e008b90635c1b7b10c/include/tools/color.hxx#L453-L466)

Jean-Abou-Samra:
> I can't successfully resolve these errors, I'll have to take advice:
> 
>     * staroffice.py has CRLF but I can't remove them

I didn't see them locally. If you Git is properly configured with core.autocrlf, they are in your source tree but automatically converted to LF when pushing.

>     * contrast test fail, perhaps Pygments uses different combinations

This was added some time ago to ensure new styles had good contrast (for visually impaired people). You need to change the colors slightly to make them more contrasted with the white background.

flywire:
> * contrast test fail, perhaps Pygments uses different combinations

Let's see if we can change LibreOffice colors: https://bugs.documentfoundation.org/show_bug.cgi?id=149934

birkenfeld:
haha, nice!

flywire:
This is _not_ right to go yet.

Expected default (blue) style to apply for all styles not named and their descendants, non-named and descendants showing as black.

```
    default_style =             '#000080'    # Blue
    styles = {
        Comment:                '#696969',   # DimGray
        Error:                  '#800000',   # Maroon
        Literal:                '#EE0000',   # Red
        Name:                   '#008000',   # Green
    }
```

Expected:

image

Getting (note first value black not default blue):

image

Code:
```bas
Sub OpenGreeting
    ' Sample code
    print "Hello World"
```

PR #2183, opened 2022-07-11
Register new style

### LINKED ISSUE #980: docs bug - registering new styles
(Original issue 1276 created by jcorneli on 2016-08-16T18:46:51.178115+00:00)

I've written a new file icky.py as a variant of default.py in the styles directory but when I install pygments (`sudo python setup.py install`), I don't see the new "icky" style in the list of available styles (`./pygmentize -L styles`).

The documentation says that all I need to do is "drop it in the *styles* subpackage".  But that's somewhat unclear.  What's really needde is to add the style to `__init__.py`:

```
#!python
    ....
    'arduino':  'arduino::ArduinoStyle',
    'icky':  'icky::IckyStyle'
```
Perhaps you can add this to http://pygments.org/docs/styles/

flywire:
Closes #980

### DIFF
--- a/doc/docs/styledevelopment.rst
+++ b/doc/docs/styledevelopment.rst
@@ -25,7 +25,7 @@ define some styles:
             String:                 'bg:#eee #111'
         }
 
-That's it. There are just a few rules. When you define a style for `Name`
+That's it, save it as ``your.py``. There are just a few rules. When you define a style for `Name`
 the style automatically also affects `Name.Function` and so on. If you
 defined ``'bold'`` and you don't want boldface for a subtoken use ``'nobold'``.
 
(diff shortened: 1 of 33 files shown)


Anteru:
Could you please add a test which exercises the new behavior?

Anteru:
Ok so wait, `default_style` was _never_ used previously? I thought this would be an observable behavior change if someone relied on `default_style` but guess I was mistaken?

birkenfeld:
Regarding this PR: I think the advice to copy stuff into the Pygments package should be removed wholesale, or at least clarified to only apply if you add a new style for submission as a PR.

flywire:
Paste in a sample plugin here and I'll rewrite https://pygments.org/docs/plugins/ then tweak https://pygments.org/docs/styledevelopment/ to make it consistent with the comments above.

flywire:
@birkenfeld [bd9eff0](https://github.com/pygments/pygments/pull/2183/commits/bd9eff03d182b5e4b20add2da22a866f019a07bf) only tries to represent the existing information more simply. The main need for it is the manual change to `__init__.py`. It would be easier if you just made any change you have in mind.

---

https://pygments.org/docs/  Hacking for Pygments is written around changing the Pygments package. I don't disagree it should be written around developing a Plugin but the section would need rewriting. It should start with https://pygments.org/docs/plugins/ containing a fully working `your` plugin consistent with the rest of the section. The emphasis of the other parts of this section should be reworked to be directed more towards the plugin than the package.

flywire:
It's not documented in the existing guide but implied. The guide has example style code followed by _"That's it."_. It truly is _it_ if the file is placed into the style folder and the user doesn't need the services from registering the style (ie bringing up on a list of styles).

It's useful functionality so I wouldn't remove it but it is probably not worth any more documentation than a comment in the code.

birkenfeld:
Hm, I don't like that either. It's also different from other components (lexers, formatters).

Anteru:
What's the next step here? It feels like we can merge this change and then continue with the rest elsewhere?

Anteru:
Merged, thanks. I'll add a warning note to the documentation to address @birkenfeld 's feedback. We still need a resolution for `get_style_by_name` though.

PR #2196, opened 2022-07-30
pytest: do not raise DeprecationWarnings as errors

birkenfeld:
This "fixes" test failures seen with Pillow >= 9.2.0, and any future such cases.


birkenfeld:
This is a RFC - do we want to keep raising these or only display them as warnings?

A third way would be to only make them errors in the CI, so that we see the deprecations early, but don't let them fail the test suite run by e.g. distributors/packagers.

PR #2198, opened 2022-07-30
formatters/img: use new API for bounding box for Pillow 9.2

birkenfeld:


birkenfeld:
TBH I just looked at https://pillow.readthedocs.io/en/stable/_modules/PIL/ImageFont.html#ImageFont.getsize which shows that all of the three methods (getsize, getbbox, getlength) use the same underlying API...

pygments/formatters/img.py
```diff
@@ -206,13 +206,17 @@ def get_char_size(self):
...
 
     def get_text_size(self, text):
         """
-        Get the text size(width, height).
+        Get the text size (width, height).
         """
-        return self.fonts['NORMAL'].getsize(text)
+        font = self.fonts['NORMAL']
+        if hasattr(font, 'getbbox'):  # Pillow >= 9.2.0
+            return font.getbbox(text)[2:4]
```
Anteru, line 217:
```suggestion
            return font.getlength(text)
```

Anteru (approved):
What was the reason to pick `getbox` over `getlength`? The underlying API is the same and this avoids unpacking the tuple.

pygments/formatters/img.py
birkenfeld: If I read the docs correctly, `getlength` only returns a single value, while `getsize` returns `(width, height)`.

PR #2203, opened 2022-08-05
languages doc page: mapping of languages with aliases

### LINKED ISSUE #2202: [Documentation] Improve languages list page
## Idea

It would be nice to enhance the page that list supported languages (https://pygments.org/languages/) to add the corresponding alias / keyword / CSS class.

## Resources

The template to modify:

https://github.com/pygments/pygments/blob/7620ef295ee9ac3ad2723847bc6722b5beec248e/doc/_templates/languages.html#L10-L17

The mapping between the name and the aliases is already done here: https://github.com/pygments/pygments/blob/master/pygments/lexers/_mapping.py

Example of other syntax highlighters that is already doing that on their documentation:

- List: https://prismjs.com/#supported-languages
- Table: https://highlightjs.readthedocs.io/en/latest/supported-languages.html

## Motivation

Some mapping are trivial but some others are non-intuitive or have many aliases. The page list the name of the languages supported but not the aliases which is the information users will need if they build a website using pygments (devs) or are just end users using a solution that use pygments. So having the mapping on the documentation and not only  on the source code will be very helpful.

noraj:
it's a naive and minimal implementation to fix #2202

it could be enhanced to display a table or to surround aliases in an inline code style with a mono font and a style added to it

### DIFF
--- a/doc/_templates/languages.html
+++ b/doc/_templates/languages.html
@@ -10,9 +10,9 @@
 {% for language in languages %}
 <li>
     {% if language.url %}
-    <a href="{{language.url}}">{{language.name}}</a>
+    <a href="{{language.url}}">{{language.name}}</a> - {{ ', '.join(language.aliases) }}
     {% else %}
-    {{language.name}}
+    {{language.name}} - {{ ', '.join(language.aliases) }}
     {% endif %}
 </li>
 {% endfor %}


noraj:
> Are you aware of https://pygments.org/docs/lexers/?

Yes but this is not concise and people will look for `pygments languages supports` on search engines or click on the `languages` button from the menu bar

image

noraj:
Keeping the two pages sounds good too. Only name and aliases which is easy and concise to read for end user that just want to know which keyword use above their markdown code blocks typically and the lexer page for devs that includes lexer options, class, file extensions, mime types, the version it where added, etc. which is nice to have for devs but make the page harder to read for non-technically-advanced people. Sounds like two pages for two use case for two kind of publics to me.

Anteru:
I'm working on an improved overview page, hopefully I'll have it finished this week. That'll replace the "languages" page list with a table, which will show the extensions accepted and a link to the relevant page on the lexer overview.

Anteru:
Implemented via #2260 .

noraj:
Nice work for https://pygments.org/languages/ !

PR #2204, opened 2022-08-07
Disable highlighting of Some Escape Codes for Python bytes Literals

LaurenceWarne:
Hi!  I saw escape codes of the form `"\N{name}"`, `"\uxxxx"` and `"\Uxxxxxxxx"` are currently highlighted within bytes literals, but according to https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals these escape codes are only recognized in string literals so this PR removes highlighting of these three forms from bytes literals.

I've added a couple of golden tests for escape code highlighting in string and bytes literals, apologies if this is elsewhere I couldn't find anything similar.

Let me know what you think, thanks :slightly_smiling_face:


LaurenceWarne:
Thanks :slightly_smiling_face:

PR #2208, opened 2022-08-14
CFamilyLexer: refuse quotes between parentheses for function definiti…

### LINKED ISSUE #2207: C++ Lexer ErrorToken: \ # '
I've attached some files that are failing with the C++ Lexer. Most have been broken since c1a0d82acbe6fa026a1f6843fee00cc1df626549. One has been broken since fc56ab883a80f6182339d7fbf95e531ff32e4809

[pygments_cpp_tests.tar.gz](https://github.com/pygments/pygments/files/9332799/pygments_cpp_tests.tar.gz)

jeanas:
…ons and declarations

Something like

id id2("){ ... }");

is no longer wrongly recognized as a "function"

id id2(") {
  ...
}
");

As the difference in the tests shows, this has the unfortunate side
effect that we no longer highlight something like

int f(param="default");

as a function declaration, but it is hard to imagine another way to
fix this (cf. “most vexing parse” problem).

Fixes #2207

### DIFF
--- a/pygments/lexers/c_cpp.py
+++ b/pygments/lexers/c_cpp.py
@@ -132,9 +132,9 @@ class CFamilyLexer(RegexLexer):
              r'(' + _possible_comments + r')'    # possible comments
              r'(' + _namespaced_ident + r')'             # method name
              r'(' + _possible_comments + r')'    # possible comments
-             r'(\([^;]*?\))'                          # signature
+             r'(\([^;"\']*?\))'                          # signature
              r'(' + _possible_comments + r')'    # possible comments
-             r'([^;{/]*)(\{)',
+             r'([^;{/"\']*)(\{)',
              bygroups(using(this), using(this, state='whitespace'), Name.Function, using(this, state='whitespace'),
                       using(this), using(this, state='whitespace'), using(this), Punctuation),
              'function'),
@@ -143,9 +143,9 @@ class CFamilyLexer(RegexLexer):
              r'(' + _possible_comments + r')'    # possible comments
              r'(' + _namespaced_ident + r')'             # method name
              r'(' + _possible_comments + r')'    # possible comments
-             r'(\([^;]*?\))'                          # signature
+             r'(\([^;"\']*?\))'                          # signature
              r'(' + _possible_comments + r')'    # possible comments
-             r'([^;/]*)(;)',
+             r'([^;/"\']*)(;)',
              bygroups(using(this), using(this, state='whitespace'), Name.Function, using(this, state='whitespace'),
                       using(this), using(this, state='whitespace'), using(this), Punctuation)),
             include('types'),
--- a/tests/examplefiles/cpp/example.cpp.output
+++ b/tests/examplefiles/cpp/example.cpp.output
@@ -944,7 +944,7 @@
 '    '        Text.Whitespace
 'string'      Name
 ' '           Text.Whitespace
-'getOpenTag'  Name.Function
+'getOpenTag'  Name
 '('           Punctuation
 'const'       Keyword
 ' '           Text.Whitespace
--- a/tests/snippets/c/test_string_resembling_decl_end.txt
+++ b/tests/snippets/c/test_string_resembling_decl_end.txt
@@ -1,0 +1,41 @@
+---input---
+// This should not be recognized as a function declaration followed by
+// garbage.
+string xyz(");");
+
+// This should not be recognized as a function definition.
+
+string xyz("){ }");
+
+---tokens---
+'// This should not be recognized as a function declaration followed by\n' Comment.Single
+
+'// garbage.\n' Comment.Single
+
+'string'      Name
+' '           Text.Whitespace
+'xyz'         Name
+'('           Punctuation
+'"'           Literal.String
+');'          Literal.String
+'"'           Literal.String
+')'           Punctuation
+';'           Punctuation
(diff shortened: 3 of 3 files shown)


birkenfeld:
Yeah we can't pretend to be a C++ parser.

birkenfeld (approved):

amitkummer:
Looks great!

martinburchell:
@Jean-Abou-Samra thanks for your prompt efforts here.

This is an improvement for our documentation. There is one regression with this change:
```

class QCP_LIB_DECL QCPLayer : public QObject
{
  Q_OBJECT
  /// \cond INCLUDE_QPROPERTIES
  Q_PROPERTY(LayerMode mode READ mode WRITE setMode)
  /// \endcond
public:

  /*!
    Defines the different rendering modes of a layer. Depending on the mode, certain layers can be
    replotted individually, without the need to replot (possibly complex) layerables on other
    layers.

    \see setMode
  */
  enum LayerMode { lmLogical   ///< Layer is used only for rendering order, and shares paint buffer with all other adjacent logical layers.
                   ,lmBuffered ///< Layer has its own paint buffer and may be replotted individually (see \ref replot).
                 };
};

```
now fails. The full file is at https://github.com/ucam-department-of-psychiatry/camcops/blob/master/tablet_qt/qcustomplot/qcustomplot.h

I realise this could ending up as a game of C++ lexer whac-a-mole and one failiure is better than seven. I appreciate it's never going to be perfect and I'm going to add the ability to skip problematic files in our docs build.

amitkummer:
**TLDR**: I don't have a fix, but I am sharing my findings on what happens wrong with the lexer when lexing @martinburchell's code. It is a little hard to spot.

In @martinburchell's snippet this entire part is somehow lexed as a function declaration:

```
Q_OBJECT
  /// \cond INCLUDE_QPROPERTIES
  Q_PROPERTY(LayerMode mode READ mode WRITE setMode)
  /// \endcond
public:

  /*!
    Defines the different rendering modes of a layer. Depending on the mode, certain layers can be
    replotted individually, without the need to replot (possibly complex) layerables on other
    layers.

    \see setMode
  */
  enum LayerMode { lmLogical   ///< Layer is used only for rendering order, and shares paint buffer with all other adjacent logical layers.
                   ,lmBuffered ///< Layer has its own paint buffer and may be replotted individually (see \ref replot).
                 };
```

More specifically:
- `Q_OBJECT` is matched as the return type (which is fine).
- `Q_PROPERTY` is matched as the function's name (also fine).
- **And this entire part is matched as the function's parameters (very much not fine)**:
 ```
(LayerMode mode READ mode WRITE setMode)
  /// \endcond
public:

  /*!
    Defines the different rendering modes of a layer. Depending on the mode, certain layers can be
    replotted individually, without the need to replot (possibly complex) layerables on other
    layers.

    \see setMode
  */
  enum LayerMode { lmLogical   ///< Layer is used only for rendering order, and shares paint buffer with all other adjacent logical layers.
                   ,lmBuffered ///< Layer has its own paint buffer and may be replotted individually (see \ref replot)
 ```

It appears that [this](https://github.com/pygments/pygments/blob/a060b201d5d661e82bb72d4d4f15cbc7abf4f516/pygments/lexers/c_cpp.py#L146) line in the 'functions' regex is what captures the above code block.
Not sure why this happens, as `r'(\([^;"\']*?\))'` has a `*?` and not just a `*` so it's supposed to be non-greedy, and only capture `(LayerMode mode READ mode WRITE setMode)` at the start -- not the entire above code block.

amitkummer:
Oh! Thanks for explaining that `*?`. Makes total sense now. I will make a small pull request with a fix.

And yes, this is not a regression.

martinburchell:
> By “failure”, do you mean you’re getting error tokens? (I’m not at a computer right now.)

Yes, but I believe this is fixed in the next release isn't it?

```
File "/home/martinb/.virtualenvs/camcops/lib/python3.8/site-packages/pygments/__init__.py", line 82, in highlight
return format(lex(code, lexer), formatter, outfile)
File "/home/martinb/.virtualenvs/camcops/lib/python3.8/site-packages/pygments/__init__.py", line 61, in format
formatter.format(tokens, realoutfile)
File "/home/martinb/.virtualenvs/camcops/lib/python3.8/site-packages/pygments/formatter.py", line 94, in format
return self.format_unencoded(tokensource, outfile)
File "/home/martinb/.virtualenvs/camcops/lib/python3.8/site-packages/pygments/formatters/html.py", line 988, in format_unencoded
for t, piece in source:
File "/home/martinb/.virtualenvs/camcops/lib/python3.8/site-packages/pygments/formatters/html.py", line 802, in _wrap_div
yield from inner
File "/home/martinb/.virtualenvs/camcops/lib/python3.8/site-packages/pygments/formatters/html.py", line 819, in _wrap_pre
yield from inner
File "/home/martinb/.virtualenvs/camcops/lib/python3.8/site-packages/pygments/formatters/html.py", line 843, in _format_lines
for ttype, value in tokensource:
File "/home/martinb/.virtualenvs/camcops/lib/python3.8/site-packages/pygments/filter.py", line 19, in _apply
yield from filter_.filter(lexer, stream)
File "/home/martinb/.virtualenvs/camcops/lib/python3.8/site-packages/pygments/filters/__init__.py", line 785, in filter
raise self.exception(value)
pygments.filters.ErrorToken: \
```

PR #2225, opened 2022-09-06
Objective-J: fix catastrophic backtracking in comment regex

jeanas:
This use of *? was both incorrect and catastrophically backtracking
when embedding this regex into a larger regex with a trailing pattern,
since it could match much more than intended and try exponentially
many positions in that process.


birkenfeld:
LGTM.

PR #2235, opened 2022-09-16
Makefile: handle "no pytest" case better

### LINKED ISSUE #2234: make test fails with can't find '__main__' module
I also tried `make all` with similar results.
```sh
$ git reset --hard && git clean -fd && make clean && make test
HEAD is now at b58e855e Fix syntax highlighting of INI language (#2217)
find . -name '__pycache__' -exec rm -rf {} +
rm -rf doc/_build build Pygments.egg-info
rm -f codetags.html
/usr/bin/python3: can't find '__main__' module in '/home/abentley/hacking/pygments/tests'
make: *** [Makefile:53: test] Error 1
```

birkenfeld:
Fixes #2234

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@
 #
 
 PYTHON ?= python3
+PYTEST ?= pytest
 
 export PYTHONPATH = $(shell echo "$$PYTHONPATH"):$(shell python -c 'import os; print ":".join(os.path.abspath(line.strip()) for line in file("PYTHONPATH"))' 2>/dev/null)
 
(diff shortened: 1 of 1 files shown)


birkenfeld:
Yeah, `-m pytest` should be much better.

I didn't suggest to get rid of the makefile, I like those little shortcuts :)

Anteru:
Now that we're doing `-m pytest`, is there any reason to keep this open? I think we should have a separate discussion about moving everything to `tox` (which I'd also prefer but I am using `make` quite a bit on Pygments.)

PR #2260, opened 2022-10-24
Improve the languages list.

Anteru:
* Move the generation into sphinext, similar to other overviews we have.
* Generate a table instead of the plain list. This table contains links
  to the lexer class now.

This addresses #2202, #2203.


Anteru:
This is how the languages page looks like with this change applied:
grafik

birkenfeld:
Very nice!

Does it need pagination? :)

Anteru:
Thanks! Future work?

I'll leave this open tonight and will merge it tomorrow then unless anyone has objections.

PR #2265, opened 2022-10-27
all: style fixes

birkenfeld:


birkenfeld:
No, that was mostly manual (with some editor macros)... thanks for taking a look!

PR #2266, opened 2022-10-27
Try switching to the `regex` module

birkenfeld:
This is the start of the experiment to see if we can switch to `regex` (the resulting version would probably be called 3.0), and then start using its additional features like matching time limits.

So far, I just mechanically replaced `re` by `regex`.

:heavy_plus_sign: only one test failure to fix; I suspect it's a Unicode version compatibility thing.

:heavy_minus_sign: the test suite now takes roughly 2x wallclock time. Just looking at `test/examplefiles` shows the same, and it doesn't seem like there is a heavy outlier. Haven't investigated further yet.


Moosems:
Is there a performance difference between the two

PR #2269, opened 2022-10-30
Add opaque pointer type to LLVM IR lexer

VladimirV99:
Reference: https://llvm.org/docs/OpaquePointers.html


birkenfeld:
Thanks!

PR #2270, opened 2022-11-01
Remove unneeded _format_unencoded_with_lineno function from IRCFormatter

Ayowel:
This PR removes `_format_unencoded_with_lineno ` from `IRCFormatter` to avoid code duplication and fix a bug where the newline and the linenumber would be inserted too early.

### DIFF
--- a/pygments/formatters/irc.py
+++ b/pygments/formatters/irc.py
@@ -128,38 +128,12 @@ class IRCFormatter(Formatter):
         self._lineno = 0
 
     def _write_lineno(self, outfile):
-        self._lineno += 1
-        outfile.write("\n%04d: " % self._lineno)
-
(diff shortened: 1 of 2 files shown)


Anteru:
Could you please add a dedicated test which highlights one of the bugs you'd have with that?

Ayowel:
@Anteru Done, see `test_linecount_output`. thank you for the comment

The old version would put the line-end `\` on a new line instead of at the end. I put two in the example, but one is enough to show the issue, the second was part of additional tests I ran that allowed me to simplify my changes as I had made some in anticipation of an issue that doesn't seem to arise in reality

Anteru:
Thanks for your contribution!

axljoker03 (approved):

PR #2271, opened 2022-11-04
Add support for Wren language

PureFox48:
This PR adds syntax highlighting support for the Wren programming language.

[Wren's github](https://github.com/wren-lang/wren)
[Wren's website](https://wren.io/)

### DIFF
--- a/pygments/lexers/wren.py
+++ b/pygments/lexers/wren.py
@@ -1,0 +1,97 @@
+"""
+    pygments.lexers.wren
+    ~~~~~~~~~~~~~~~~~~~~
+    Lexer for Wren.
+    :copyright: Copyright 2006-2022 by the Pygments team, see AUTHORS.
+    :license: BSD, see LICENSE for details.
+"""
+
+import re
+
+from pygments.lexer import RegexLexer
+from pygments.token import Whitespace, Punctuation, Keyword, Name, Comment, \
+    Operator, Number, String
+
+__all__ = ['WrenLexer']
+
+class WrenLexer(RegexLexer):
+    name = 'Wren'
+    aliases = ['wren']
+    filenames = ['*.wren']
+
+    flags = re.MULTILINE | re.DOTALL
+
+    tokens = {
+        'root': [
+            # Whitespace.
+            (r'\s+', Whitespace),
+            (r'[,\\\[\]{}]', Punctuation),
+
+            # Push a parenthesized state so that we know the corresponding ')'
+            # is for a parenthesized expression and not interpolation.
+            (r'\(', Punctuation, ('parenthesized', 'root')),
+
+            # In this state, we don't know whether a closing ')' is for a
+            # parenthesized expression or the end of an interpolation. So, do
+            # a non-consuming match and let the parent state (either
+            # 'parenthesized' or 'interpolation') decide.
+            (r'(?=\))', Whitespace, '#pop'),
+
+            # Keywords.
+            (r'(as|break|class|construct|continue|else|for|foreign|if|import|'
+             r'in|is|return|static|super|var|while)\b', Keyword),
+
+            (r'(true|false|null)\b', Keyword.Constant),
+
+            (r'this\b', Name.Builtin),
+
+            # Comments.
+            (r'/\*', Comment.Multiline, 'comment'), # Multiline, can nest.
+            (r'//.*?$', Comment.Single),            # Single line.
+            (r'#.*?(\(.*?\))?$', Comment.Special),  # Attribute or shebang.
+
+            # Names and operators.
+            (r'[!%&*+\-./:<=>?\\^|~]+', Operator),
+            (r'[a-z][a-zA-Z_0-9]*', Name),
+            (r'[A-Z][a-zA-Z_0-9]*', Name.Class),
+            (r'__[a-zA-Z_0-9]*', Name.Variable.Class),
+            (r'_[a-zA-Z_0-9]*', Name.Variable.Instance),
+
+            # Numbers.
+            (r'0x[0-9a-fA-F]+', Number.Hex),
+            (r'\d+(\.\d+)?([eE][-+]?\d+)?', Number.Float),
+
+            # Strings.
+            (r'"""', String, 'raw_string'),
+            (r'"', String, 'string'),
+        ],
+        'comment': [
+            (r'/\*', Comment.Multiline, '#push'),
+            (r'\*/', Comment.Multiline, '#pop'),
+            (r'.', Comment.Multiline), # All other characters.
+        ],
+        'raw_string': [
+            (r'"""', String, '#pop'),
+            (r'"', String),
+            (r'"', String, '#pop'),
+            (r'.', String), # All characters.
+        ],
+        'string': [
+            (r'"', String, '#pop'),
+            (r'\\[\\%"0abefnrtv]', String.Escape), # Escape.
+            (r'\\x[a-fA-F0-9]{2}', String.Escape), # Byte escape.
+            (r'\\u[a-fA-F0-9]{4}', String.Escape), # Unicode escape.
+            (r'\\U[a-fA-F0-9]{8}', String.Escape), # Long Unicode escape.
+
+            (r'%\(', String.Interpol, ('interpolation', 'root')),
+            (r'.', String), # All other characters.
+        ],
+        'parenthesized': [
+            # We only get to this state when we're at a ')'.
+            (r'\)', Punctuation, '#pop'),
+        ],
+        'interpolation': [
+            # We only get to this state when we're at a ')'.
+            (r'\)', String.Interpol, '#pop'),
+        ],
+    }
--- a/tests/examplefiles/wren/example.wren
+++ b/tests/examplefiles/wren/example.wren
@@ -1,0 +1,121 @@
+#!/bin/wren
+
+/* IMPORTS */
+import "random" for Random as Rand
+
+/* COMMENTS */
+
+//  single line comment
+
+/*
+    multiline comment
+    /*
+        nested multiline comment
+    */
+*/
+
+/* CLASSES & ATTRIBUTES */
+
+#!type = "parent"
+class Parent {
+    #method
+    static setField (field) {
+        __field = field
+    }
+
+    construct new(parent) {
+        _parent = parent
+        return
+    }
+
+    parent { _parent }
+
+    foreign method()
+}
+
+#!type = "child"
+#group(
+    multiple,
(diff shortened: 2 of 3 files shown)


pygments/lexers/wren.py
```diff
@@ -0,0 +1,97 @@
...
+from pygments.token import Whitespace, Punctuation, Keyword, Name, Comment, \
+    Operator, Number, String
+
+__all__ = ['WrenLexer']
+
+class WrenLexer(RegexLexer):
+    name = 'Wren'
+    aliases = ['wren']
+    filenames = ['*.wren']
+
```
Anteru, line 21: Please add an `url` as well which points to the language website, see: https://github.com/pygments/pygments/blob/master/pygments/lexer.py#L81

pygments/lexers/wren.py
```diff
@@ -0,0 +1,97 @@
...
+            # Push a parenthesized state so that we know the corresponding ')'
+            # is for a parenthesized expression and not interpolation.
+            (r'\(', Punctuation, ('parenthesized', 'root')),
+
+            # In this state, we don't know whether a closing ')' is for a
+            # parenthesized expression or the end of an interpolation. So, do
+            # a non-consuming match and let the parent state (either
+            # 'parenthesized' or 'interpolation') decide.
+            (r'(?=\))', Whitespace, '#pop'),
```
Anteru, line 38: Should this be `Punctuation`?

pygments/lexers/wren.py
```diff
@@ -0,0 +1,97 @@
...
+
+            # Keywords.
+            (r'(as|break|class|construct|continue|else|for|foreign|if|import|'
```
Anteru, line 41: Please use the `words()` function to create a match for all those names.

pygments/lexers/wren.py
```diff
@@ -0,0 +1,97 @@
...
+            (r'\d+(\.\d+)?([eE][-+]?\d+)?', Number.Float),
+
+            # Strings.
+            (r'"""', String, 'raw_string'),
+            (r'"', String, 'string'),
+        ],
+        'comment': [
+            (r'/\*', Comment.Multiline, '#push'),
+            (r'\*/', Comment.Multiline, '#pop'),
+            (r'.', Comment.Multiline), # All other characters.
```
Anteru, line 71: This should be:
```python
'comment': [
    (r'\*/', Comment.Multiline, '#pop'),
    (r'[^*]+', Comment.Multiline),
    (r'\*', Comment.Multiline),
]
```
See: https://github.com/pygments/pygments/blob/master/Contributing.md

pygments/lexers/wren.py
```diff
@@ -0,0 +1,97 @@
...
+        ],
+        'raw_string': [
+            (r'"""', String, '#pop'),
+            (r'"', String),
+            (r'"', String, '#pop'),
+            (r'.', String), # All characters.
```
Anteru, line 77: Same issue as above.

pygments/lexers/wren.py
```diff
@@ -0,0 +1,97 @@
...
+        ],
+        'string': [
+            (r'"', String, '#pop'),
+            (r'\\[\\%"0abefnrtv]', String.Escape), # Escape.
+            (r'\\x[a-fA-F0-9]{2}', String.Escape), # Byte escape.
+            (r'\\u[a-fA-F0-9]{4}', String.Escape), # Unicode escape.
+            (r'\\U[a-fA-F0-9]{8}', String.Escape), # Long Unicode escape.
+
+            (r'%\(', String.Interpol, ('interpolation', 'root')),
+            (r'.', String), # All other characters.
```
Anteru, line 87: And once more ...

Anteru (changes requested):
Thanks for the contribution! Please check the comments to (dramatically) reduce the number of generated tokens, at the moment, you're matching character-by-character in various places which produces enormous amounts of tokens.

pygments/lexers/wren.py
PureFox48: Done.

pygments/lexers/wren.py
PureFox48: Yes it should, thank you.

pygments/lexers/wren.py
PureFox48: Fixed. Note also that I've decided to reclassify 'in' and 'is' as 'Operator.Word' rather than 'Keyword'.

pygments/lexers/wren.py
PureFox48: Changed, thanks.

pygments/lexers/wren.py
PureFox48: I've removed the 'raw_string' section altogether as they are now fully described in 'root'.

pygments/lexers/wren.py
PureFox48: Improved in that last line can now pick up multiple characters. Added a line to 'example.wren' to check empty string being highlighted correctly.

PureFox48: Had imported 'include' unneccesarily. Removed now.

PureFox48: Just noticed that 'prefixes' weren't quite right. Fixed them now.

pygments/lexers/wren.py
PureFox48: Unlike many other C-family languages, block  comments in Wren can nest (see [here](https://wren.io/syntax.html#Comments)).

Although it's not a big problem, your standard pattern for block comments doesn't deal properly with that. So, if I try to highight this line:
```javascript
/* This is /* a nested */ comment. */
```
the word 'comment' is highlighted as a name rather than as part of the overall comment because the lexer thinks that the first ```*/``` is the end of the outer block.

Do you have a suggested pattern for that situation? If not, I'm happy to leave it as it is.

PureFox48: OK, thanks I've changed it to that. Nested block comments now working fine :)

pygments/lexers/wren.py
```diff
@@ -0,0 +1,102 @@
...
+
+            # Comments.
+            (r'/\*', Comment.Multiline, 'comment'), # Multiline, can nest.
+            (r'//.*?$', Comment.Single),            # Single line.
+            (r'#.*?(\(.*?\))?$', Comment.Special),  # Attribute or shebang.
+
+            # Names and operators.
+            (r'[!%&*+\-./:<=>?\\^|~]+', Operator),
+            (r'[a-z][a-zA-Z_0-9]*', Name),
+            (r'[A-Z][a-zA-Z_0-9]*', Name.Class),
```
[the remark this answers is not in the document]
PureFox48: Wren identifiers are ASCII only - no Unicode allowed for now.

pygments/lexers/wren.py
```diff
@@ -0,0 +1,102 @@
...
+            (r'__[a-zA-Z_0-9]*', Name.Variable.Class),
+            (r'_[a-zA-Z_0-9]*', Name.Variable.Instance),
```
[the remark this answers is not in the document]
PureFox48: See earlier comment.

pygments/lexers/wren.py
```diff
@@ -0,0 +1,102 @@
...
+
+import re
+
+from pygments.lexer import RegexLexer, words
+from pygments.token import Whitespace, Punctuation, Keyword, Name, Comment, \
+    Operator, Number, String
+
+__all__ = ['WrenLexer']
+
+class WrenLexer(RegexLexer):
```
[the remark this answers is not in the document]
PureFox48: I've added one.

pygments/lexers/wren.py
```diff
@@ -0,0 +1,102 @@
...
+        ],
+        'string': [
+            (r'"', String, '#pop'),
+            (r'\\[\\%"0abefnrtv]', String.Escape), # Escape.
+            (r'\\x[a-fA-F0-9]{2}', String.Escape), # Byte escape.
+            (r'\\u[a-fA-F0-9]{4}', String.Escape), # Unicode escape.
+            (r'\\U[a-fA-F0-9]{8}', String.Escape), # Long Unicode escape.
+
+            (r'%\(', String.Interpol, ('interpolation', 'root')),
+            (r'[^\\"%]*', String), # All remaining characters.
```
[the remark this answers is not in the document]
PureFox48: I've changed it as per your suggestion. Empty strings still highlighting correctly.

Anteru:
Can you please take a look at the failing test? `--randomly-seed=2087311863`? I presume it generated some input which led to an infinite number of tokens being generated.

PureFox48:
Unfortunately, I don't have a working Windows machine - I only use Linux these days - so I've no idea why the first Windows test is failing. Possibly, it doesn't like the prefixes I've used - ```r'(?<!\.)'``` in the keywords section. I could change those to ```r'\b'``` without difficulty because, syntactically, keywords can't be used as method names in Wren.

Do you have any clue what the problem could be from past experience?

Incidentally, on my own machine (Ubuntu 22.04), 'make test' runs fine - 4131 passed, 4 skipped.

As regards 'check (pull_request)` failing, this is my first docstring:
```
"""
    pygments.lexers.wren
    ~~~~~~~~~~~~~~~~~~~~
    Lexer for Wren.
    :copyright: Copyright 2006-2022 by the Pygments team, see AUTHORS.
"""
```
Looking at other lexers, I suspect I need to insert a blank line before and after 'Lexer for Wren'.

My second docstring looks like this but I can't see anything wrong with it at all:
```
    """
    For Wren source code, version 0.4.0.

    .. versionadded:: 2.14.0
    """
```

birkenfeld:
You're correct regarding the module docstring. It also should begin in line 1, not line 2.

As for the Windows test, the tests have a random component (testing well-behavedness with a random sequence of input characters), so it should be reproducible regardless of OS with the pytest argument that @Anteru provided.

PureFox48:
OK, thanks.

I'll change the module docstring and see if I can get to the bottom of the other problem by running pytest myself with that random seed though I only have Python 3.10.6 installed.

PureFox48:
I've corrected the module docstring but other issue not yet resolved.

PureFox48:
Hopefully, I've fixed the failing test by adding another regex at the end of 'root' to catch an arbitrary series of tokens.

I've also taken this opportunity to reclassify 'this' as an ordinary keyword (analogous to 'super') rather than a named built-in as it wasn't really highlighting properly before.

PureFox48:
Ah, I see now where the problem was.

Yes, it looks OK to me and is working fine.

Thanks for your help in finally resolving this issue :)

PR #2275, opened 2022-11-13
Update GitHub actions.

Anteru:
The https://github.com/actions/setup-python action is at version 4, so use that, and update checkout to v3 while we're at it. Turn on caching in the hope it'll make things faster one day, and stop trying to pull the latest `pip` version every time.


birkenfeld:
Sure!

PR #2277, opened 2022-11-14
Accept colon(s) in YAML key names

gdubicki:
as it's ok according to the YAML specs and is widely used in the real world, f.e. in Puppet's Hiera (https://puppet.com/docs/puppet/7/hiera_quick.html#values_common_data)

### DIFF
--- a/pygments/lexers/data.py
+++ b/pygments/lexers/data.py
@@ -231,7 +231,7 @@ class YamlLexer(ExtendedRegexLexer):
             # whitespaces separating tokens
             (r'[ ]+', Whitespace),
             # key with colon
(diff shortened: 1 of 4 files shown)


birkenfeld:
This also accepts question marks, and it allows leading colons; is that intended?

gdubicki:
Thanks for a super-fast response @birkenfeld! 😊

The former was not intended, I fixed it now.

The latter was intended as it's the same as having a colon at any other place in the key name. I added a test for that.

birkenfeld:
Looks good, thanks!

axljoker03 (approved):

PR #2280, opened 2022-11-19
Add the NASM file format supports

AshrafAliS:
tianocore EDK2 have the NASM file format alternative to ASM based On which compiler user is using.
https://github.com/tianocore/edk2

NASM are same as ASM file format

Signed-off-by: Ashraf Ali S <ashraf.ali.s@intel.com>


Anteru:
Thanks for your contribution!

PR #2282, opened 2022-11-24
Added support for F# script files, aka .fsx files

voronoipotato:
I noticed this was missing and was causing issues for me as I use fsx files and even though the syntax is more or less the same it's still helpful to get the highlight without having to rename the file.


voronoipotato:
@Anteru  awaiting approval :) thank you!

voronoipotato:
No problem, always happy to help!

PR #2287, opened 2022-11-29
New attempt to add Python 3.11 to CI now that it's released

jeanas:


Anteru:
Yeah, that's on me -- we only use `lxml` for unit tests and I'm thinking to have a fall back if `lxml` isn't present. The problem is that there's no nice equivalent of `lxml.html`. It may be possible to do something with `html.parser` ... I'll give it a try.

Anteru:
Yes, agreed. The problem is I did tinker with `html.parser` when I wrote the structural diff originally and I don't remember now why I didn't end up using it any more, so it may take some time to get a workaround in place.

Anteru:
Take this! #2301

PR #2292, opened 2022-12-03
fix solidity lexer not picking up many of the operators (e.g. boolean and)

f0rki:
I noticed that the lexer for Solidity would not pick up boolean operators. Odd that nobody noticed this. To fix it I switched to the same regex for operators in the solidity lexer as in the `CFamilyLexer`. I updated the testcase with a function that covers all the operators I am aware of.

### DIFF
--- a/pygments/lexers/solidity.py
+++ b/pygments/lexers/solidity.py
@@ -60,7 +60,7 @@ class SolidityLexer(RegexLexer):
             include('constants'),
             (r'[a-zA-Z_]\w*', Text),
             (r'[~!%^&*+=|?:<>/-]', Operator),
-            (r'[.;:{}(),\[\]]', Punctuation)
+            (r'[.;{}(),\[\]]', Punctuation)
         ],
         'comments': [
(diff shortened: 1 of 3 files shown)


pygments/lexers/solidity.py
```diff
@@ -59,7 +59,7 @@ class SolidityLexer(RegexLexer):
             (datatype, Keyword.Type),
             include('constants'),
             (r'[a-zA-Z_]\w*', Text),
-            (r'[!<=>+*/-]', Operator),
+            (r'[~!%^&*+=|?:<>/-]', Operator),
```
birkenfeld, line 62: `:` is now recognized as both `Operator` and `Punctuation` - which makes more sense?

f0rki: good catch. it is somewhat ambiguous. `:` can be used in the shorthand if operator `a ? b : c` or also in structs like `{x : y}`. I would stick with Operator.

birkenfeld:
Thanks! Tests are happy too, so let's go.

PR #2298, opened 2022-12-04
Force whitespace tokens in bygroups.

Anteru:
This solves tricky cases like Praat for example using `'([\w.]+)(:|\s*\()'` as a regex with bygroups, and there's no easy way to find out which path is taken there. By handling it in bygroups, we can "post-process" those matches. This has a performance cost of course, and I'd prefer a better way to handle this so I'm open for ideas.

Another downside is that this overwrites token types like `Literal.String`, because I don't know at replacement time if the original action was `Text` (which seems most common) or something else like `Punctuation`. As a result, I can't recommend merging this, but I'd like to start the discussion how to approach this :)

### DIFF
--- a/pygments/lexer.py
+++ b/pygments/lexer.py
@@ -19,11 +19,13 @@ from pygments.util import get_bool_opt, get_int_opt, get_list_opt, \
     make_analysator, Future, guess_decode
 from pygments.regexopt import regex_opt
 
+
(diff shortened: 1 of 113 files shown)


Anteru:
Thanks, maybe I'm just too dumb then and/or shouldn't try it late :) I tried to refactor it but always ended up with the choice not working, and there were a few more cases where I figured it's just easier to force it later. I'll revisit them and see if I can restructure them as well.

Issue #2299, opened 2022-12-06
Add lexer for Portugol language


LorhanSohaky:
Portugol is a simplified programming language based on Pascal language. This language is used for teach programming for people that speak Portuguese but don't understand English.

https://www.apoioinformatica.inf.br/produtos/visualg/linguagem

denilsonsa:
And this is the list of the keywords from that language: https://www.apoioinformatica.inf.br/produtos/visualg/item/29-referencias-da-linguagem-de-programacao-do-visualg

PR #2300, opened 2022-12-06
Add support for PortugolLexer

LorhanSohaky:
Portugol don't have a extensive documentation, so I based on Pascal lexer and the [manual of Visualg](https://www.apoioinformatica.inf.br/produtos/visualg/linguagem) that is a program that's implements the Portugol.

issue #2299


LorhanSohaky:
> This looks like lots of copy/paste from `DelphiLexer`, can you merge it into it by introducing a new boolean option to the lexer similar to `turbopascal` and `freepascal`?

It's because Portugol doesn't understand Delphi's or Pascal's reserved words

LorhanSohaky:
Some differences:
- `hex` and `char` types are not supported
- there are no classes
- only single comments are supported
- the operators `;` `@` and `^` are not supported
- portugol has the operator `<-`
- asm is not supported
- `label` and `goto` are not supported
- function modifiers are not supported
- strings only with double quotes

Still do you think it's better to put portugol in DelphiLexer? A bit afraid is to appear several `if portugol` in the code or stylize tokens that shouldn't.

LorhanSohaky:
@Jean-Abou-Samra , ok. How can I pass this parameter to the Lexer

image

LorhanSohaky:
> `lexer = DelphiLexer(portugol=True)`

It's because when running the tests I don't know how to pass this parameter.

```python -m pygments -x -l pygments/lexers/pascal.py:DelphiLexer tests/examplefiles/delphi/example.alg```

And every time the user uses it, will he have to pass this parameter? For example, in mkdocs material

\`\`\`portugol(portugol=true)

// My code
\`\`\`

https://squidfunk.github.io/mkdocs-material/reference/code-blocks/

LorhanSohaky:
Perfect! Do you recommend me putting it in the same file or creating a file just for Portugol Lexer as a Delphi inheritor?

LorhanSohaky:
@Jean-Abou-Samra , thanks!

LorhanSohaky:
> Thanks for the update, looks much better now. I'll do some touch-ups.

I thank you. Can you tell me when the next release will be? I'm very excited to use

PR #2301, opened 2022-12-07
Remove the lxml dependency.

Anteru:
Replace lxml with html.parser for the structural-diff.


Anteru:
I'm not entirely sure why I used to `strip()` the text content previously, this seems to work fine in my quick testing, and if it breaks because of whitespace adding a `strip()` back is easy so I'll merge it now.

PR #2311, opened 2022-12-30
styles gallery: Make docstring in example follow PEP 257

not-my-profile:
The example is displayed multiple times at https://pygments.org/styles/. Code displayed on the Pygments website should follow best practices, such as https://peps.python.org/pep-0257/.


Anteru:
Merged, nice catch :)

PR #2325, opened 2023-01-25
Native: fix contrast ratio to meet WCAG AA guidelines

hugovk:
# Before

The red `#define _Py_SHARED_SHIFT 2` shown [here](https://peps.python.org/pep-0703/#biased-reference-counting), with `class="cp"` is `#cd2828`:

image

Has a contrast ratio of 3.04 which doesn't meet the WCAG AA minimum of 4.5:

image

image

And is flagged by Lighthouse:

image

# After

Using Google's suggestion of `#ff3a3a` increases it to 4.59:

image

image

(For comparison, the AAA suggestion of `#ff8888` is a bit pale, so let's go for AA.)

<details>
<summary>AAA</summary>

image

</details>

### DIFF
--- a/pygments/styles/native.py
+++ b/pygments/styles/native.py
@@ -27,7 +27,7 @@ class NativeStyle(Style):
         Whitespace:         '#666666',
 
         Comment:            'italic #ababab',
-        Comment.Preproc:    'noitalic bold #cd2828',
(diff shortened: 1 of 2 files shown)


birkenfeld:
Works for me. Thanks!

PR #2326, opened 2023-01-30
Fix lots of missing encoding="utf-8" parameters

Jean-Abou-Samra:
Some of these are probably unnecessary (ASCII-only content), but it's easier not to think about it.


birkenfeld (approved):
Yeah, makes sense.

PR #2328, opened 2023-01-30
Match lexers and formatters by extension separatly

dirkmueller:
d0487e3b3ec7 "Remove filename pattern caches" (#2153) introduced a huge performance regession. While it is true that fnmatch already uses functools.lru_cache, that cache is limited to 127 entries and we have over 1000 matching patterns, which means the cache is evicted entirely on every iteration.

We can be more clever without reverting that patch by avoiding fnmatch calls alltogether, which provides an even better speedup. The bulk (> 99%) of all filename matches are on the filename extension, so in gen_mapfiles.py we split these into a separate tuple list that we match without calling fnmatch().

This restores previous performance speedup without the overhead of an extra regexp cache.


dirkmueller:
I haven't done a full microbenchmark on this other than simple timeit executions:

```python
print(timeit.timeit('fnmatch.fnmatch("foo.py", "*.py")', setup='import fnmatch'))
print(timeit.timeit('"foo.py".endswith(".py")'))
print(timeit.timeit('"py" in ("lala", "loop", "py")'))
```

which prints here:

```
0.5052752160008822
0.08419276200220338
0.028339595002762508
```

so  a "if str in tuple of strings" is ~ factor 20 faster than fnmatch(), even with a cached regexp, so beyond the pathological case where the regexp cache is overflowing like in my original case. Also, many of the regular expressions actually have no wildcard at all. > 90% have wildcards that match only on extension, and another small percentage is matching on the full filename (so no regular expression at all). a meaningful regexp is the execption.

Thats what led me into writing the full patch, which I applied in my (much larger) program and my profiled bottleneck went completely away.  There's certainly a bit of cleanup that can be done on my patch to make it less convoluted. an API break would make this simpler (if the plugin would specify matches on the extension separately for example). 

my original program that let me to profile this essentially runs lexers.guess_lexer_for_filename() several 10 thousands of times.

dirkmueller:
I have pushed a plain revert (with some import cleanups to pass tests) instead. let me know how you want to proceed.

dirkmueller:
@Jean-Abou-Samra I don't have such a slowdown with python 3.11. with the simple revert patch applied, your benchmark.py 
is

```
python3.11:
0.7443678290001117

python3.10:
0.7769828950004012
```

which is in line what is expected of the general speedup of python 3.11. so looks good to me. thank you for the quick merge!

PR #2334, opened 2023-02-02
Add missing built-in functions for Python

otaithleigh:
Adds two missing builtins for the Python lexer, [`callable`](https://docs.python.org/3/library/functions.html#callable) and [`aiter`](https://docs.python.org/3/library/functions.html#aiter).

`callable` was removed in Python 3.0, but added back in 3.2, while `aiter` was added in 3.10.

Interestingly, `help` is also missing from the list of builtins, but I'm assuming that's more of a conscious exclusion (it's also missing from the Python2 and Cython lexers), so I haven't added it in this PR.


BhinbahadurUK:
Just thinking if merged string auto write of python with voice or audio as normal input formula for sequence for entry

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

birkenfeld:
Seems to be just strange spam. User should be reported if it happens more .

PR #2342, opened 2023-02-14
Move metadata to pyproject.toml

jeanas:
Split off from #2331.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,6 @@
 [build-system]
-requires = ["setuptools"]
+# setuptools added pyproject.toml support in v61.0.0
+requires = ["setuptools >= 61"]
 build-backend = "setuptools.build_meta"
 
 [project]


pyproject.toml
```diff
@@ -1,6 +1,59 @@
 [build-system]
-requires = [
-  "setuptools >= 40.9.0",
-  "wheel",
+requires = ["setuptools"]
```
Anteru, line 2: Out of curiosity, is `wheel` now automatically included? Your build has a `.whl` but I always had to add it here for `python -m build` to pick it up.

Anteru (approved):

Anteru:
Thanks! Glad to see this updated so we remain good citizens of the Python ecosystem.

setup.cfg
```diff
@@ -1,54 +0,0 @@
...
-    Topic :: Utilities
-project_urls =
-    Documentation = https://pygments.org/docs/
-    Source = https://github.com/pygments/pygments
-    Bug Tracker = https://github.com/pygments/pygments/issues
-    Changelog = https://github.com/pygments/pygments/blob/master/CHANGES
-
-[options]
-packages = find:
-zip_safe = false
```
birkenfeld, line 39: This and `include_package_data` doesn't seem to show up in the new one - is it the default now?

PR #2344, opened 2023-02-15
Move more tests to tests/snippets instead of .py files

jeanas:
These are easier to update in tests/snippets/ (just run with --update-goldens).  Also, a couple of them were checking too lenient conditions, and this uncovers that they actually look broken (JavaScript).

This is partial work; there is more to do.


birkenfeld:
I don't think the `test_` prefix is needed?

birkenfeld:
Right, I confused it with examplefiles. In the snippets most seem to have a `test_` prefix so let's keep it.

Do we have an issue for the JavaScript problems?

Issue #2348, opened 2023-02-21
Current implementation of Rebol code coloriser is not suitable for code snippets


Oldes:
Pygments currently expects Rebol header in the code and so is useful only for full scripts. But Pygments is used also in Zulip chat and there one is usually pasting just code snippets.

Here is how it looks now [without the header](https://pygments.org/demo/?lexer=rebol&style=default&formatter=html&code=foreach+%3Atype-record+boot-types+%5B%0A++++if+find+%5Bfunction+operator%5D+evalclass+%5B%0A++++++++emit-line%2Fvar+%22Do_%22+type+none%0A++++%5D%0A%5D), and this is [with the header](https://pygments.org/demo/?lexer=rebol&style=default&formatter=html&code=Rebol+%5B%5D%0Aforeach+%3Atype-record+boot-types+%5B%0A++++if+find+%5Bfunction+operator%5D+evalclass+%5B%0A++++++++emit-line%2Fvar+%22Do_%22+type+none%0A++++%5D%0A%5D).

PR #2349, opened 2023-02-21
Don't require script headers in Rebol and Red codes.

Oldes:
Resolves: https://github.com/pygments/pygments/issues/2348

### DIFF
--- a/pygments/lexers/rebol.py
+++ b/pygments/lexers/rebol.py
@@ -4,7 +4,7 @@
 
     Lexers for the REBOL and related languages.
 
-    :copyright: Copyright 2006-2023 by the Pygments team, see AUTHORS.
+    :copyright: Copyright 2006-2022 by the Pygments team, see AUTHORS.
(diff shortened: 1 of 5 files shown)


birkenfeld:
The header should still be highlighted in some way if present I suppose?

Oldes:
@birkenfeld unfortunately as there is no way to find out if the code is a complete script or just a snipped of it, the header may not be highlighted how it was. Because one could have such an input:
```rebol
switch language [
    Rebol [ print "bla" ]
    Perl  [ print "foo" ]
]
```
But as the header is just an ordinary Rebol data, it will be highlighted too... the only difference is, that what is before header in a complete script will not be handled like a comment.

birkenfeld:
OK, makes sense. Just one nit: please remove the unneeded code instead of commenting.

Oldes:
@birkenfeld the code is removed now.

birkenfeld:
Thanks!

PR #2350, opened 2023-02-22
Fix GLSL and HLSL preprocessor line continuation

bitsawer:
Both shading languages allow their preprocessor to use line continuation with backlashes `\`, if they are the last character on a line before newline or carriage return. This PR adds support for that:

``` glsl
// Multiline replacement using backslashes for continuation:
#define SAMPLE(param1, param2, param3, param4) long_function_call( \
        param1, \
        param2, \
        param3, \
        param4 \
)
```

Noticed this problem when generating Godot's documentation for its shader preprocessor (which is very similar to GLSL and HLSL preprocessor). Notice how the first shader code block in https://docs.godotengine.org/en/latest/tutorials/shaders/shader_reference/shader_preprocessor.html does not get syntax highlighting at all as the current lexer produces an error because of `\` characters. This PR fixes the issue and the code block is highlighted correctly.


Anteru:
Thanks! Could you please add/extend an example file with this so we can capture future regressions? You'll need to run `pytest --update-goldens` afterwards.

bitsawer:
Thanks for the feedback! Removed "\r?" and added some tests, let me know if I missed something.

PR #2357, opened 2023-02-26
Fix GLSL and HLSL preprocessor directive start

bitsawer:
Currently, both shader language lexers require that when parsing preprocessor directives `#` is the first character on a line. This PR loosens the restriction as these should all be valid directives:

``` glsl
#define NORMAL 1
#    define VALUE 10
    #define MORE 20
    #    define ANOTHER 30
```

This also allows putting line numbers, non-breaking spaces etc. in certain code snippets:

``` glsl
10. #ifdef VALUE
11. ...
12. #endif
```

### DIFF
--- a/pygments/lexers/graphics.py
+++ b/pygments/lexers/graphics.py
@@ -30,7 +30,7 @@ class GLShaderLexer(RegexLexer):
 
     tokens = {
         'root': [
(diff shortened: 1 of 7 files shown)


Anteru:

PR #2376, opened 2023-03-15
add support for lexing attributes in Chapel

arezaii:
This PR adds support for language changes in Chapel. 

Chapel 1.30 includes support for attributes. The lexer was not able to process them without some adjustments.

### DIFF
--- a/pygments/lexers/chapel.py
+++ b/pygments/lexers/chapel.py
@@ -135,6 +135,6 @@ class ChapelLexer(RegexLexer):
             (r'[^()]*', Name.Other, '#pop'),
         ],
         'attributename': [
-            (r'([a-zA-Z_][.\w$]*)', Name.Decorator, '#pop'),
+            (r'[a-zA-Z_][.\w$]*', Name.Decorator, '#pop'),
         ],
     }


pygments/lexers/chapel.py
```diff
@@ -133,4 +134,7 @@ class ChapelLexer(RegexLexer):
             (words(known_types, suffix=r'\b'), Keyword.Type),
             (r'[^()]*', Name.Other, '#pop'),
         ],
+        'attributename': [
+            (r'([a-zA-Z_][.\w$]*)', Name.Decorator, '#pop'),
```
amitkummer, line 138: Secondly, is `Name.Attribute` more appropriate here? Additionally, the parenthesis here can also be omitted. 

```suggestion
            (r'[a-zA-Z_][.\w$]*', Name.Attribute, '#pop'),
```

arezaii: I was not certain on the best choice for token type here. Our "attributes" don't really align with any of the existing types as I read them in https://pygments.org/docs/tokens/. They look and function something like Java annotations, which is why I thought maybe `Decorator` was appropriate, but I would defer to someone with more knowledge of the token system in `pygments`.

amitkummer: `Decorator` sounds good then, thanks!

Issue #2391, opened 2023-03-29
copyright check looks for 2022, not 2023


dneto0:
This is the cause of the copyright check failure in #3047

Anteru:
Looks for 2023 as of #2393

PR #2392, opened 2023-03-29
Allow 2022 or 2023 as the copyright year

dneto0:
Fixes: #2391


dneto0:
This is probably not what you have in mind.
Updating strictly to 2023 requires updating many source files.

Anteru:
Let's try #2393.

Issue #2399, opened 2023-03-31
pygments fails to detect scdoc lexer


stacyharper:
Hey there,

I'm investigating why scdoc files are rendered that badly on srht, and it looks like this is because pygments fails to guess the lexer correctly.

Example of rendered pages:

https://git.sr.ht/~stacyharper/sxmobar/tree/main/item/docs/sxmobar.1.scd
https://git.sr.ht/~sircmpwn/himitsu/tree/master/item/docs/hiq.1.scd

Trying pygments locally:

```
$ scripts/debug_lexer.py -g sxmobar.1.scd
Using lexer: Transact-SQL (pygments.lexers.sql.TransactSqlLexer)
Error parsing sxmobar.1.scd on line 3
Previous tokens:
   'sxmobar'                     Token.Name 
   '('                           Token.Punctuation 
   '1'                           Token.Literal.Number.Integer 
   ')'                           Token.Punctuation 
   '\n\n'                        Token.Text.Whitespace 
Error token:
   '#' 
```

How can I help?

PR #2402, opened 2023-04-05
Improve scdoc analyze_text

stacyharper:
Fixes: https://github.com/pygments/pygments/issues/2399


stacyharper:
> @stacyharper This PR is marked “draft”, does that still reflect its status?

I'm not very familiar with pygments. This is still a RFC to me. If you guys thinks it is good, or could be improved, let me knows

PR #2404, opened 2023-04-05
Improve Java properties lexer

### LINKED ISSUE #2356: Catastrophic backtracking on Java's properties Lexer
The following pattern is inefficient:
`r'([ \f\t]*)([=:]*)([ \f\t]*)(.*(?<!\\)(?:\\{2})*)(\\)(?!\\)$`
[https://github.com/pygments/pygments/blob/master/pygments/lexers/configs.py#L141](https://github.com/pygments/pygments/blob/master/pygments/lexers/configs.py#L141)

jeanas:
Use special lexer rules for escapes; fixes catastrophic backtracking, and highlights them too. (Also change Operator to Punctuation for the key/value separator.)

Fixes #2356

### DIFF
--- a/pygments/lexers/configs.py
+++ b/pygments/lexers/configs.py
@@ -129,26 +129,42 @@ class PropertiesLexer(RegexLexer):
 
     tokens = {
         'root': [
-            (r'\s+', Whitespace),
+            # comments
             (r'[!#].*|/{2}.*', Comment.Single),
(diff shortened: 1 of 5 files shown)


Anteru:
The change looks good, but could you please elaborate why operator -> punctuation? `=` feels like an operator to me.

PR #2406, opened 2023-04-08
Fix token for None after fromimport in PythonLexer

rdbende:
Pygments yields `Token.Name.Builtin.Pseudo` for `None` after a `from` keyword. Seems like this behavior remained from Python2Lexer, where it was indeed `Name.Builtin.Pseudo`, but the correct token in Python3Lexer would be `Keyword.Constant`.


Anteru:
Good catch, thanks!

PR #2410, opened 2023-04-12
Python console: do not require output that looks like a traceback to be valid

### LINKED ISSUE #2407: Traceback parsing changes in 2.15.0 error on valid pycon code blocks
(Sorry, accidentally hit the submit shortcut before I was done writing it)

Unfortunately, the traceback parsing changes to fix #2226 and #2329 implemented in fe42aac (I can't seem to find the associated pull request) appears to result in Pygments treating certain valid code blocks as invalid, issuing a warning and failing Sphinx builds (with `-W`). In particular, this occurred on the following code block in the CPython's `multiprocessing` module docs:

```rst
.. note::

   Functionality within this package requires that the ``__main__`` module be
   importable by the children. This is covered in :ref:`multiprocessing-programming`
   however it is worth pointing out here. This means that some examples, such
   as the :class:`multiprocessing.pool.Pool` examples will not work in the
   interactive interpreter. For example::

      >>> from multiprocessing import Pool
      >>> p = Pool(5)
      >>> def f(x):
      ...     return x*x
      ...
      >>> with p:
      ...     p.map(f, [1,2,3])
      Process PoolWorker-1:
      Process PoolWorker-2:
      Process PoolWorker-3:
      AttributeError: Can't get attribute 'f' on <module '__main__' (<class '_frozen_importlib.BuiltinImporter'>)>
      AttributeError: Can't get attribute 'f' on <module '__main__' (<class '_frozen_importlib.BuiltinImporter'>)>
      AttributeError: Can't get attribute 'f' on <module '__main__' (<class '_frozen_importlib.BuiltinImporter'>)>
```

 that worked in 2.14.9, which when 2.15.0 was released suddenly broke the CPython Docs CI job with the error:

```
WARNING: Could not lex literal_block as "pycon". Highlighting skipped.
```

Here's a more minimal example that also doesn't parse:

```rst
Here's a code block::

   >>> print("\n".join(["Traceback (most recent call last):"] * 2))

```

This also occurs with an explicit `pycon` lexer specified:

```rst
Here's another code block:

.. code-block:: pycon

   >>> print("\n".join(["Traceback (most recent call last):"] * 2))
```

and also with arbitrary text after the traceback line:

```rst
Here's a code block::

   >>> print("\n".join(["Traceback (most recent call last):", "blah blah"])
   blah blah
```

While this example parses:

```rst
Here's a code block::

   >>> print("\n".join(["Traceback (most recent call last):"] * 1))
```

[Here's an example of a failing run](https://github.com/python/cpython/actions/runs/4658331019/jobs/8243879847).

birkenfeld:
which causes Error to be flagged with output that just happens to look like a traceback

Fixes #2407

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -3,9 +3,12 @@ Pygments changelog
 
 Pull request numbers before 2.4.2 are not linked as they refer to the now defunct Bitbucket project.
 
-Version 2.16.0
+Version 2.15.1
 --------------
 (unreleased)
+
+- Fix Python console traceback lexing being too strict (#2407)
+
 
 Version 2.15.0
 --------------
@@ -67,6 +70,7 @@ Version 2.15.0
 - The top-level ``Makefile`` has been removed. Instead, all shortcuts
   for developing are now defined and run through tox.  The ``doc`` folder
   still contains a ``Makefile`` as an alternative to ``tox -e doc``.
+
 
 Version 2.14.0
 --------------
--- a/pygments/lexers/python.py
+++ b/pygments/lexers/python.py
@@ -763,7 +763,8 @@ class PythonTracebackLexer(RegexLexer):
             (r'^([^:]+)(: )(.+)(\n)',
              bygroups(Generic.Error, Text, Name, Whitespace), '#pop'),
             (r'^([a-zA-Z_][\w.]*)(:?\n)',
-             bygroups(Generic.Error, Whitespace), '#pop')
+             bygroups(Generic.Error, Whitespace), '#pop'),
+            default('#pop'),
         ],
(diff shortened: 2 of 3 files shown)


birkenfeld:
Note: alternate fix would be to just allow stray lines in `PythonTracebackLexer`.

Anteru:
The fix is fine, but it produces "one character tokens". Can we track if we're in "error mode" and while we're in error mode, accumulate tokens, and if we're in the first non-error token, yield the whole text accumulated so far?

birkenfeld:
Yeah, true, I like this better then.

PR #2412, opened 2023-04-12
Draft: Refactor PythonConsoleLexer as a DelegatingLexer

### LINKED ISSUE #2411: pycon lexer still reorders input
Unfortunately, even after https://github.com/pygments/pygments/commit/fe42aaca7b000200ea9915c04c01f2fb3f4dbaf6 and https://github.com/pygments/pygments/pull/2410, PythonConsoleLexer still reorders some inputs. For example, this:

```
>>> unterminated_traceback()
>>> 
```

is rendered as

```
>>> unterminated_traceback()
>>> 
```

jeanas:
This is simpler and more reliable than hand-coding the state machine.

Fixes #2411

I need to split this into more commits and check a few things before it can be 'undrafted'.

Also, while this should IMHO be the way going forward, I'm not sure about including a full rewrite of a lexer in a bugfix release. On the other hand, #2411 is not a regression compared to 2.14 as far as I can see.

### DIFF
--- a/pygments/lexers/python.py
+++ b/pygments/lexers/python.py
@@ -11,8 +11,8 @@
 import re
 import keyword
 
-from pygments.lexer import Lexer, RegexLexer, include, bygroups, using, \
-    default, words, combined, do_insertions, this, line_re
+from pygments.lexer import DelegatingLexer, Lexer, RegexLexer, include, \
+    bygroups, using, default, words, combined, do_insertions, this, line_re
 from pygments.util import get_bool_opt, shebang_matches
 from pygments.token import Text, Comment, Operator, Keyword, Name, String, \
     Number, Punctuation, Generic, Other, Error, Whitespace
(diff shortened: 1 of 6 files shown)


Anteru:
I like that solution better than the current workarounds. I just checked, the new lexer doesn't appear in the documentation, so that's good. I don't mind shipping a lexer rewrite given it passes all existing tests, and the current test coverage is _better_ than what we shipped the last major update with (hence the regressions!) Unless @birkenfeld has any reservations, my vote is on merging this and getting a 2.15.1 release out of the door as soon as we can.

birkenfeld:
I'm in holiday, my only reservations are in hotels. Go ahead :)

Anteru (approved):

Anteru:
@jeanas Good to go or do you want to do anything else before merging? It's currently in draft so I cannot trivially merge this.

Anteru:
Thanks! Yes -- that's enough for a point release. I'll get #2404 included as well. I should get to it tomorrow. Good work on both!

PR #2424, opened 2023-05-04
Update uxntal lexer to reflect current runes

tbsp:
Since the lexer was originally created the ' and : runes have been deprecated, and the -_=?! runes have been added. 

Uxntal reference: https://wiki.xxiivv.com/site/uxntal.html

The original example piano.tal file has been updated with a newer reference file (used with permission), as the old one included a macro which conflicted with the new - rune, and used fairly old uxntal coding style.


non:
Looks good to me. Thanks! :+1:

neauoire:
Thank you for the new runes support.

PR #2440, opened 2023-05-24
Add Eclipse-Light style

Sam2567:
Reference :  [Syntax_Highlighting](https://www.eclipse.org/pdt/help/html/syntax_highlighting.htm). 
The test result:
Test

Also, I suggest that we should sort the **STYLE_MAP** dictionary in alphabet order. I could quickly do it if everyone thinks it's a good idea.

Thanks xD

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -27,6 +27,7 @@ Version 2.16.0
     positive; add bold italic markup (#2447)
 
 - Add ``Generic.EmphStrong`` token for bold italic markup (#2444)
+- Improve contrast in Monokai style (#2448)
 
 Version 2.15.1
 --------------


pygments/styles/eclipse.py
```diff
@@ -0,0 +1,40 @@
...
+
+class EclipseStyle(Style):
+    background_color = "#ffffff"
+    PURPLE = "#7F0055"
+    BLUE = "#0000FF"
+    RED = "#FF0000"
+    BLUE_COMMENT = "#3F5FBF"
+    GREEN_COMMENT = "#3F7F5F"
+    BLACK = "#000000"
+
```
Anteru, line 14: Can you also please set:

```
    line_number_color
    line_number_background_color
    line_number_special_color
    line_number_special_background_color
    highlight_color
```

In addition to `background_color`, so you're not getting defaults that may or may not work for this? I suspect there's a particular highlight color you want to use for Eclipse?

Anteru (changes requested):
I agree, that dictionary should be sorted, but I'll do it afterwards to not pollute this PR with unrelated changes.

pygments/styles/eclipse.py
Anteru: Thanks for adding the other colors!

It's your style -- all I'm saying is the default colors might change one day, and hence I'd prefer if new styles define all colors we support to ensure they continue to work if we ever decide that the default should be bright pink. We haven't been paying much attention to this previously and it's yet another item on my todo list to automatically check this :(

Sam2567: I get it now. Thanks for your reply!
I think it's a good idea to keep my style not having too much adherent with the default settings. 

I have already added the missing copyright header to the file. Now I am working on the contrast issue that @jeanas mentioned 
Seems like my style failed a test due to not having enough contrast. Does that mean I need more "black&white" stuff to make my style more readable? Seems like whatever I try to edit the style and run the `test_contrasts.py`, I would only get the following error.

"AssertionError: contrast degradation for style 'eclipse'
The following rules have a contrast lower than the required 4.5:
* 3.80 Token.Error"

Sam2567: > It's no problem to have failing tests on a PR — to the contrary, we can help you figure out how to fix them and/or push commits ourselves to fix them.
Thanks for your help && reply
Hi, I changed the error to even darker colour like matte black. Seems like that number increased from 3.8 to 4.0! I probably need to tweak a few more colours to make my style a bit more readable.

Issue #1422, opened 2020-04-06
Option or lexer modification to ignore existing ANSI sequences within the target text 


r4dr3fr4d:
I would like to have existing ANSI sequences already within some target text that I want to highlight.  I'm doing this for a custom text editor which will need background coloring in arbitrary locations for visual highlighting.  To make handling the offsets where this highlighting starts and ends as easy as possible, I'd like to add those sequences in prior to adding in the ANSI sequences resulting from the syntax highlighting.  Is there a way to do this or a better way to meet the same end?

Issue #1148, opened 2019-08-31
Can't format console/ shell-session output that includes ANSI colors


Anteru:
(Original issue 1444 created by purplefishies on 2018-06-06T16:50:49.003688+00:00)

I'm trying to take log data from Tmux ( with capture-pane -e ) that includes ANSI color codes that colorize some of the output. 

When I run pygmentize -f latex  sample.shell-session it generates Latex with the ANSI codes still in the latex file.

This causes a latex compile failure.

kdeldycke:
I just managed to hack my way into [rendering shell session output's with ANSI colors](https://kdeldycke.github.io/click-extra/tutorial.html#codecell3).

New ANSI-capable styles are available in my upcoming [`click-extra`](https://github.com/kdeldycke/click-extra) v3.0.0 release. You should be able to do:
```shell-session
$ pip install click-extra >= 3.0.0
```

Then use my new `ansi-`prefixed lexers.

Here is an example in Sphinx:

```rst
.. code-block:: ansi-shell-session
    $ hello --help
    [94m[1m[94m[1mUsage[0m: [0m[97mhello[0m [OPTIONS]
    
      Simple program that greets NAME for a total of COUNT times.[0m
    
    [94m[1m[94m[1mOptions[0m:[0m
      [36m--dummy-flag / --no-flag[0m  [default: no-flag][0m
      [36m--my-list TEXT[0m
      [36m--count INTEGER[0m           Number of greetings.  [default: [35m1[0m][0m
      [36m--name TEXT[0m               The person to greet.[0m
      [36m--time / --no-time[0m        Measure and print elapsed execution time.  [default:[0m
                                no-time][0m
      [36m--color, [36m--ansi[0m / --no-color, --no-ansi[0m
                                Strip out all colors and all ANSI codes from output.[0m
                                [default: color][0m
      [36m-C, [36m--config[0m [90mCONFIG_PATH[0m[0m  Location of the configuration file. Supports both[0m
                                local path and remote URL.  [default:[0m
                                ~/.hello/config.{toml,yaml,yml,json,ini,xml}][0m
      [36m-v, [36m--verbosity[0m [90mLEVEL[0m[0m     Either [35mCRITICAL[0m, [35mERROR[0m, [35mWARNING[0m, [35mINFO[0m, [35mDEBUG[0m.[0m
                                [default: [35mINFO[0m][0m
      [36m--version[0m                 Show the version and exit.[0m
[...]

Issue #477, opened 2019-08-31
Support ANSI (ECMA-48) color-coded text input


Anteru:
(Original issue 770 created by mgorny on 2012-05-14T06:18:43.247173+00:00)

I feel like many pastebin users would appreciate such a feature in pygments: reading (already colored) data stream with ECMA-48 color codes embedded (like console program output) and outputting it in any format pygments support.

I guess such a feature would require adding a new token types to represent console colors.

If there's interest in this, I can assemble a patch for it.

Anteru:
(Original comment by tshatch on 2012-11-12T23:04:42.629400+00:00)

I'd be interested in at least a 16-color version of this.  Some of the other codes (like bold, reverse, etc) might overcomplicate it however.  Can you get a basic version ready to review in the 1.7 time frame (guessing 6 months out)?

Anteru:
(Original comment by mgorny on 2012-11-13T05:15:04.206501+00:00)

Yes, I think I can. Unless I forget ;).

Anteru:
(Original comment by tshatch on 2013-05-10T01:35:36.912696+00:00)

How's it coming?

Anteru:
(Original comment by mgorny on 2013-05-10T07:00:57.332637+00:00)

I'm sorry but I'm overburdened with work all the time, and this one doesn't come as a priority to me. I will be happy to work on this but I don't know when I'll have a free hand.

Anteru:
(Original comment by dejanm on 2014-11-20T20:25:36.473941+00:00)

I was looking into this too. Seems to be somewhat botched. Michał, do you still intend to work on such a feature?

Anteru:
(Original comment by mgorny on 2014-11-20T20:44:11.394289+00:00)

I'm sorry but I can't afford to work on this in any foreseeable future.

Anteru:
(Original comment by dejanm on 2014-11-26T14:55:02.287765+00:00)

I started working on this, but I'm not entirely sure how to proceed. My first idea was to replace ANSI color codes with RGB color codes, but I'm not sure how to do that, i.e. how to insert those RGB color codes into the output. Can a lexer do that kind of preprocessing?

Or is it necessary to have this kind of token rules:

(r'%s(3[0-7])m(.*?)%s%s' % (esc, esc, reset_re), ansi2rgb)

where ansi2rgb would lookup and yield appropriate color tokens. If so, where would we define these tokens? All tokens are defined in various styles, but that seems like an overkill.

[...]

PR #2442, opened 2023-05-25
Document all language REPLs and ANSI support

kdeldycke:
This PR:
- lists all missing interactive sessions from langage REPL
- groups sessions in OS shells or langage REPLs
- documents the generic output lexer 
- adds session examples
- documents the lack of ANSI support in output
- adds pointers to third-party projects for ANSI support

Relates to:
- #1148
- #1422
- #477


doc/docs/terminal-sessions.rst
```diff
@@ -21,26 +32,288 @@ prompt.
...
 
-  .. code-block:: ps1con
+Interactive consoles
+--------------------
 
-     PS[any]> Get-ChildItem
-     PS> Get-ChildItem
-     >> Get-ChildItem
+Similarly to systems shells, Pygments recognize a variety of interactive
+language sessions:
```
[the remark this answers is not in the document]
kdeldycke: OK, I just moved all the examples to the docstrings.

kdeldycke:
This PR is ready to be reviewed and merged, even if the CI job on 3.12-dev is failing. See: https://github.com/pygments/pygments/pull/2445

Anteru:
Thanks! I'm a bit worried that the examples in the doc string will actually work with the lexer .. I'd prefer them to be also committed as test snippets. At that point however, we probably shouldn't duplicate them from the test snippets into the documentation. This could be something to solve via a private class field for the documentation, maybe something like `_doc_example = '/path/to/example'` which is then parsed by the lexer page generator. This would have to go here: https://github.com/pygments/pygments/blob/master/pygments/sphinxext.py#L148

This may sound like overkill, but I think that's the best long term solution to make sure those things don't go out of sync, and it's probably (long term) a good idea to have some example file linked to every lexer.

kdeldycke:
Thanks @Anteru for the detailed feedback!

I have these kind of reflexion too with my [Click Extra](https://github.com/kdeldycke/click-extra) project, in which I have to maintain some code to keep the document and code in sync.

I have several strategy for that:
- [dedicated Sphinx extension](https://kdeldycke.github.io/click-extra/sphinx.html) to produce Click code and invokation example
- stringent [unittests to check the metadata and content of third-party dependencies](https://github.com/kdeldycke/click-extra/blob/main/click_extra/tests/test_pygments.py#L58-L145) (including Pygments itself) 
- regular runs via GitHub workflows to [automatically produce dynamic documentation](https://github.com/kdeldycke/click-extra/blob/b5c456254378056160c6c2d541e38dba35e53246/click_extra/docs_update.py#L70-L97) and push a PR

It's a little bit all over the place by does the job of reducing the cognitive work of maintenance. I'm just pointing to my stuff as inspiration: it might not be adapted to the nature of Pygments. But I guess I know enough about Pygments and Sphinx to attempt an implementation of the long term solution you described.

But should this be done in this PR? I don't think so because of the bigger scope of that feature: it would introduce new code while this PR is only introducing rST content. So what's missing for this PR to be considered for a merge upstream?

Anteru:
Honestly, I'm not sure if the added cost of having to convert this later is worth to merge this now (I'm really afraid of code rot quickly.) I'll see if I can spare some cycles to convert this, maybe it's not as complicated as I'm afraid it is. I'm not a huge fan of having this extra field on the lexer class to point to a document in the test tree, so the other alternative is to have tests somehow marked up as examples that should show up in the doc for that particular lexer. Let me think a bit more about this, and at the very least try to prototype the `_example` thing ... I did some work on the Sphinx extension and it feels like this shouldn't be that hard other than having to somehow pipe it through Pygments while building the docs and _then_ insert it back into the doctree.

kdeldycke:
Since your last comment @Anteru, I realized you went ahead with the `_example` attribute and implemented it upstream in #2451.

So I updated this PR to use it to document all REPL sessions. All tests are passing, this PR is ready to be merged upstream! :)

doc/docs/terminal-sessions.rst
```diff
@@ -1,46 +1,55 @@
...
-with a specially formatted prompt.
+Pygments support the parsing and highlighting of terminal sessions, like
+command-line shells, interactive consoles and language `REPL
+<https://en.wikipedia.org/wiki/Read–eval–print_loop>`_.
 
-Supported shells with examples are shown below. In each example, prompt parts in
-brackets ``[any]`` represent optional parts of the prompt, and prompt parts
-without brackets or in parenthesis ``(any)`` represent required parts of the
-prompt.
+They are typiccaly command lines or code, mixed with generic output.
```
Anteru, line 9:
```suggestion
They are typically command lines or code, mixed with generic output.
```

doc/docs/terminal-sessions.rst
```diff
@@ -1,46 +1,55 @@
...
-     > ls -lh
+Operating system shells
+-----------------------
 
-* **MSDOS Session** (doscon):
+These lexers are expecting a prompt to identify user input. So to highlight a
+shell session, prefix your code snippet with a specially formatted prompt.
 
-  .. code-block:: doscon
+They are typiccaly named ``<shell> Session``.
```
Anteru, line 20:
```suggestion
They are typically named ``<shell> Session``.
```

doc/docs/terminal-sessions.rst
```diff
@@ -1,46 +1,55 @@
...
+Interactive consoles
+--------------------
 
-  .. code-block:: tcshcon
+Similarly to systems shells, Pygments recognize a variety of interactive
+language sessions.
 
-     (any)> ls -lh
-     ? ls -lh
+Their IDs are typiccaly follow the ``<language>-console`` or
```
Anteru, line 29:
```suggestion
Their IDs typically follow the ``<language>-console`` or
```

doc/docs/terminal-sessions.rst
```diff
@@ -1,46 +1,55 @@
...
+To display standalone terminal output and keep styling consistent, you can use
+the generic ``output`` lexer.
 
+
+ANSI rendering
+--------------
+
+In all the lexers above, the command results are parsed as generic output.
+Which means they are rendered as-is, without any tentative of rendering
+the styling carried on by ANSI codes.
```
Anteru, line 45:
```suggestion
Which means they are rendered as-is, without any styling applied, for example by ANSI codes.
```

Anteru (changes requested):

kdeldycke:
@Anteru I just updated the PR with your requested changes, and squashed the commits. All tests are passing, this PR is ready to be merged upstream.

Anteru:
Thanks!

PR #2444, opened 2023-05-28
Add Generic.EmphStrong token

### LINKED ISSUE #2307: Add a Generic.EmphStrong
For markup languages like Markdown, syntaxes like `***bold italic***` create a ***bold italic*** styled text. Unfortunately I didn't find a way to achieve this given that we only have `Generic.Emph` and `Generic.Strong` token type. Given that such syntaxes exist in many markup languages, should we add a token type like `Generic.EmphStrong`?

diskdance:
Note that some theme do not use bold or italic styling for Generic.Emph or Generic.Strong token. For these themes, I didn't touch them.

Tested to work on default theme.

Closes #2307.

### DIFF
--- a/doc/docs/tokens.rst
+++ b/doc/docs/tokens.rst
@@ -369,6 +369,9 @@ highlight a programming language but a patch file.
 `Generic.Strong`
     Marks the token value as bold (e.g. for rst lexer).
 
+`Generic.EmphStrong`
+    Marks the token value as bold and emphasized.
+
 `Generic.Subheading`
     Marks the token value as subheadline.
 
--- a/pygments/styles/autumn.py
+++ b/pygments/styles/autumn.py
@@ -54,6 +54,7 @@ class AutumnStyle(Style):
         Generic.Error:              '#aa0000',
         Generic.Emph:               'italic',
         Generic.Strong:             'bold',
+        Generic.EmphStrong:         'bold italic',
         Generic.Prompt:             '#555555',
         Generic.Output:             '#888888',
         Generic.Traceback:          '#aa0000',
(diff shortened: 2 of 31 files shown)


Anteru:
@jeanas Is there an easy way to allow failures on -dev Python variants to fail gracefully? This seems completely unrelated/a Python 3.12 specific issue which we can't solve from our end?

hugovk:
The `-W error` is converting a deprecation warning into an error.

This change will allow the other jobs to continue, even if 3.12 is failing:

```diff
diff --git a/.github/workflows/build.yaml b/.github/workflows/build.yaml
index 3f5c26ad..bb02f37b 100644
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -12,6 +12,7 @@ jobs:
   build:
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
         os: [windows-latest, ubuntu-latest]
         python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12-dev"]
```

hugovk:
Please see https://github.com/pygments/pygments/pull/2445 for that.

The deprecation warning is:

```
DeprecationWarning: ast.Str is deprecated and will be removed in Python 3.14; use ast.Constant instead
```

Out of scope for this PR, but let's find where that is coming from and either fix it or temporarily silence it. To unblock this PR we could temporarily remove `-W error` for 3.12, or even comment out 3.12 from the CI. 

The warning could be coming from pytest, I know they're fixing some of those and should have a fix after the next 3.12 beta: https://github.com/pytest-dev/pytest/pull/10894#issuecomment-1565091020

And the next beta is scheduled for Tuesday: https://peps.python.org/pep-0693/

PR #2445, opened 2023-05-28
CI: If one job fails, allow others to continue

hugovk:
Re: https://github.com/pygments/pygments/pull/2444#issuecomment-1566271577

Allow other jobs to continue, even if 3.12 falls.


hugovk:
I still find it useful. 

For example, I may open a PR that only fails for Python 3.7 because it uses a new API from 3.8.

Or it only fails on Windows due to path or encoding issues.

This change means we can clearly see it fail on a subset and pass for the rest, instead of failing for one and the others are all cancelled.

PR #2451, opened 2023-06-08
Add support for lexer examples.

Anteru:
This adds support for a new attribute `_example` which, if present, will
load the file under `tests/examplefiles` pointed to by `_example` and
show it as part of the lexer documentation. This allows us to have
unit-tested files show up as small code snippets in the documentation,
without having to worry that the documentation output contains possibly
incorrectly highlighted code.

This would allow #2442 to use this machinery instead of embedding the snippets as "plain code" into the doc string.


birkenfeld:
:+1:

PR #2470, opened 2023-07-05
SystemdLexer and DesktopLexer

randomstuff:
This implements more precise rules for systemd unit and .desktop files. For example, this fixes entries such as:

~~~
Categories=System;TerminalEmulator;
MimeType=image/*;application/x-sh;application/x-shellscript;inode/directory;text/*;x-scheme-handler/kitty;
~~~

which were considered to contain a trailing comment.


pygments/lexers/configs.py
```diff
@@ -70,6 +66,71 @@ def analyse_text(text):
         return text[0] == '[' and text[npos-1] == ']'
 
 
+
+class DesktopLexer(RegexLexer):
+    """
+    Lexer for .desktop files.
+    """
+
+    name = 'Desktop'
```
birkenfeld, line 75: `Desktop file` please

pygments/lexers/configs.py
```diff
@@ -70,6 +66,71 @@ def analyse_text(text):
         return text[0] == '[' and text[npos-1] == ']'
```
birkenfeld, line 66: This will still catch the other files if this analyzer happens to run before them, so it will need to return 0.99 or something smaller than 1 now.

pygments/lexers/configs.py
```diff
@@ -70,6 +66,71 @@ def analyse_text(text):
...
+    url = "https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html"
+    aliases = ['desktop']
+    filenames = ['*.desktop']
+
+    tokens = {
+        'root': [
+            (r'^[ \t]*\n', Whitespace),
+            (r'^(#.*)(\n)', bygroups(Comment.Single, Whitespace)),
+            (r'(\[[^\]\n]+\])(\n)', bygroups(Keyword, Whitespace)),
+            (r'([-A-Za-z0-9]+)([ \t]*)(=)([ \t]*)([^\n]*)([ \t\n]*\n)',
```
birkenfeld, line 85: I'd also like to see special support for localized entries of the form
```desktop
Name[de]=Hallo
```

pygments/lexers/configs.py
```diff
@@ -70,6 +66,71 @@ def analyse_text(text):
...
+            (r'^([;#].*)(\n)', bygroups(Comment.Single, Whitespace)),
+            (r'([ \t]*)([^\n]*)(\\)(\n)',
+             bygroups(Whitespace, String, Text, Whitespace)),
+            (r'([ \t]*)([^\n]*)(\n)',
+             bygroups(Whitespace, String, Whitespace), "#pop"),
+        ],
+    }
+
+    def analyse_text(text):
+        return text.startswith("[Unit]")
```
birkenfeld, line 131: AFAICS, many unit files, e.g. those shipped by systemd itself, are prefixed by comments. So I think this should instead search some reasonable prefix for this, like
```
"[Unit]\n" in text[:1000]

birkenfeld (changes requested):
Generally a good idea, and can be extended later e.g. recognizing systemd's quoting and escaping mechanisms. Just a few nits...

pygments/lexers/configs.py
randomstuff: Yes, good idea. I'm not sure which token type would be suitable for this.

randomstuff: Abusing the `Name.Namespace` for this for now.

randomstuff:
Added support for .desktop language tags. Currently I'm (ab)using `Name.Namespace`.

I see that TurtleLexer uses `Generic.Emph` for a similar usage (`"..."@en`) which is not really better.

birkenfeld:
LGTM. Thanks!

birkenfeld (approved):

PR #2482, opened 2023-08-03
Update Igor Pro lexer for Igor Pro 9

MichaelHuth:
The functions and operations list in the Igor Pro lexer is updated for the recent Igor Pro 9 version.


Anteru:
Thanks for the contribution!

Issue #2499, opened 2023-08-28
HTML formatter doesn't include trailing newlines in spans of class "go", Generic.Output


vkoukis:
Lexer "console" assigns class `go: Generic.Output` to lines that correspond to the output of commands.
For example
```console
$ grep Generic.Output tests/snippets/console/prompt_in_output.txt
'line1\n'     Generic.Output
'> file content, not prompt!\n' Generic.Output
```

However, the HTML formatter first ends a span of class `go`, then adds the line separator.
```html
<div class="highlight"><pre><span></span><span class="gp gp-VirtualEnv">(venv) </span><span class="gp">user:~$ </span>ssh-keygen
<span class="go">Generating public/private rsa key pair.</span>
<span class="go">[...]</span>
</pre></div>
```

This is problematic, because even though one can assign CSS properties to the `go` class, they will not have an effect on the actual line changes. For example, even when I assign `user-select: none`  to class `go`, the user can still select and copy stray newlines between consecutive output lines, i.e., between consecutive spans.

Given that the lexer originally assigns `Generic.Output` to the whole line, including the newline character at the end, we can fix this by extending the line separator into the span itself, so the output becomes:

```html
<div class="highlight"><pre><span></span><span class="gp gp-VirtualEnv">(venv) </span><span class="gp">user:~$ </span>ssh-keygen
<span class="go">Generating public/private rsa key pair.
</span><span class="go">[...]
</span></pre></div>
```

This does allow the attributes for class `go` to also extend to the newline characters, and makes things more predictable.

Looking forward to your feedback, I can follow up with a PR that implements this behavior.

PR #2500, opened 2023-08-28
Implement various improvements to the "console" lexer, closes issues #2496-#2498

### LINKED ISSUE #2496: Lexer "console" treats part of the prompt as whitespace when in virtualenv
virtualenv adds a prefix as part of the prompt, and this prefix contains a single space after the name of the virtualenv in parentheses by default. Commit 3f403687 which resolved #1253 assigns the "Text" class to any whitespace between the virtualenv name and the rest of the prompt, but this is problematic, because it doesn't include it as part of the prompt, and makes it selectable when using the "html" formatter and the mouse, for example.

I think it would make sense to just include the whitespace as part of the virtualenv prefix, and assign it `Generic.Prompt.VirtualEnv`, since it actually is part of the prefix that the virtual environment adds to the prompt.

Here is a before an after:

**Before:**
```console
(pygments) user@host:~$ pygmentize -l console -f html  <<<"(pygments) user@host:~/some/dir$ echo one two"
<div class="highlight"><pre><span></span><span class="gp gp-VirtualEnv">(pygments)</span> <span class="gp">user@host:~/some/dir$ </span><span class="nb">echo</span><span class="w"> </span>one<span class="w"> </span>two
</pre></div>
```

**After:**
```console
(pygments) user@host:~$  pygmentize -l console -f html  <<<"(pygments) user@host:~/some/dir$ echo one two"
<div class="highlight"><pre><span></span><span class="gp gp-VirtualEnv">(pygments) </span><span class="gp">user@host:~/some/dir$ </span><span class="nb">echo</span><span class="w"> </span>one<span class="w"> </span>two
</pre></div>
```

Looking forward to your feedback. I can submit a PR to make this change, and update the tests accordingly.

### LINKED ISSUE #2497: Lexer "console" treats parts of the output as prompts, if they look like prompts
By default, `ShellSessionBaseLexer` attempts to detect multiple prompts within the input. This works nicely when input contains multiple commands, but has problems when the output of the commands contain things that looks like prompts, e.g., commented-out lines that start with `# `.

This is problematic when showing output files, because it marks parts of them as prompts, gives them a different color, and -- depending on the theme -- may even prevent the user from selecting this text.

There already exists a test for this [`prompt_in_output.txt`], but it only contains the secondary prompt in the output, not the primary prompt. Including the primary prompt, fails:

```console
$ cat <<EOF|pygmentize -l console -f html
> $ cat \
> > test.txt
> line1
> # line2, file content, not prompt!
> EOF
<div class="highlight"><pre><span></span><span class="gp">$ </span>cat<span class="w"> </span>&gt;<span class="w"> </span>test.txt
<span class="go">line1</span>
<span class="gp"># </span>line2,<span class="w"> </span>file<span class="w"> </span>content,<span class="w"> </span>not<span class="w"> </span>prompt!
</pre></div>
```

Since there is no guaranteed way to actually know which lines are part of the output and which lines actually correspond to prompts, would you be open to adding a lexer option to support only matching the prompt once, and stop parsing for prompts when we've already matched the prompt once? This allows us to optionally standardize on a single command per console block, and remove the ambiguity.

I can submit a PR to include an option, and also extend the test snippets so we can specify custom lexer options per snippet.

### LINKED ISSUE #2498: Lexer "console" doesn't seem to support heredoc syntax
When highlighting a command which uses a heredoc, pygments seems to highlight the heredoc as part of output [it assigns it the `go` class] instead of input. Here is an example:

```console
$ cat <<EOF
> line1
> line2
> EOF
line1
line2
$ cat <<EOF|pygmentize -l console -f html
> $ cat <<EOF
> > line1
> > line2
> > EOF
> line1
> line2
> EOF
<div class="highlight"><pre><span></span><span class="gp">$ </span>cat<span class="w"> </span>&lt;&lt;EOF
<span class="go">&gt; line1</span>
<span class="go">&gt; line2</span>
<span class="go">&gt; EOF</span>
<span class="go">line1</span>
<span class="go">line2</span>
</pre></div>
```

It would be appropriate to only assign the output class to the last two lines:
```console
$ cat <<EOF|pygmentize -l console -f html
$ cat <<EOF
> line1
> line2
> EOF
line1
line2
EOF
<div class="highlight"><pre><span></span><span class="gp">$ </span>cat<span class="w"> </span><span class="s">&lt;&lt;EOF</span>
<span class="gp">&gt; </span><span class="s">line1</span>
<span class="gp">&gt; </span><span class="s">line2</span>
<span class="gp">&gt; </span><span class="s">EOF</span>
<span class="go">line1</span>
<span class="go">line2</span>
</pre></div>
```

Shall I submit a PR?
This issue seems relevant to #326, #1068.

vkoukis:
ShellSessionBaseLexer: Support heredoc input
ShellSessionBaseLexer: Only match prompt once
tests: Support arbitrary Lexer options in snippets
ShellSessionBaseLexer: Support whitespace in virtualenv prompt

Closes #2496 
Closes #2497 
Closes #2498

### DIFF
--- a/pygments/lexers/shell.py
+++ b/pygments/lexers/shell.py
@@ -174,6 +174,7 @@ class ShellSessionBaseLexer(Lexer):
         insertions = []
         prompt_found = False
         backslash_continuation = False
+        heredoc_continuation = False
 
         for match in line_re.finditer(text):
             line = match.group()
@@ -202,14 +203,19 @@ class ShellSessionBaseLexer(Lexer):
                                    [(0, Generic.Prompt, m.group(1))]))
                 curcode += m.group(2)
                 backslash_continuation = curcode.endswith('\\\n')
-            elif backslash_continuation:
+                heredoc_continuation = curcode.endswith('<<EOF\n')
+            elif backslash_continuation or heredoc_continuation:
                 if line.startswith(self._ps2):
                     insertions.append((len(curcode),
                                        [(0, Generic.Prompt,
                                          line[:len(self._ps2)])]))
-                    curcode += line[len(self._ps2):]
+                    rem = line[len(self._ps2):]
+                    curcode += rem
+                    if heredoc_continuation:
+                        heredoc_continuation = not (rem == 'EOF\n')
                 else:
                     curcode += line
+                    heredoc_continuation = False
                 backslash_continuation = curcode.endswith('\\\n')
             elif self._bare_continuation and line.startswith(self._ps2):
                 insertions.append((len(curcode),
--- a/tests/snippets/console/test_heredoc.txt
+++ b/tests/snippets/console/test_heredoc.txt
@@ -1,0 +1,29 @@
+# Test that an input command can include a heredoc as input
+
+---input---
+$ cat <<EOF
+> line1
+> line2
+> EOF
+line1
+line2
+
+---tokens---
+'$ '          Generic.Prompt
+'cat'         Text
+' '           Text.Whitespace
+'<<EOF\n'     Literal.String
+
+'> '          Generic.Prompt
+'line1\n'     Literal.String
+
+'> '          Generic.Prompt
+'line2\n'     Literal.String
+
+'> '          Generic.Prompt
+'EOF'         Literal.String
+'\n'          Text.Whitespace
+
+'line1\n'     Generic.Output
+
+'line2\n'     Generic.Output


vkoukis:
This is a PR to close issues #2496, #2497, #2498, #2499.
Looking forward to any feedback you may have, happy to also split this into smaller changes, if you think these modifications need more discussion.

vkoukis:
@jeanas 
> Yes, please at least separate the HTML formatter change from console lexer changes.

ACK, I have forced push into this branch so we focus this PR on the lexer changes.
I will submit the HTML formatter change separately.
I have edited my comments above to reflect this.

PR #2501, opened 2023-08-28
formatters/html: Fold newlines into spans

### LINKED ISSUE #2499: HTML formatter doesn't include trailing newlines in spans of class "go", Generic.Output

vkoukis:
This PR updates the HTML formatter so it folds line separators into spans instead of leaving them outside spans.
This does allow the attributes for classes to also extend to the newline characters.
See #2499 for a detailed description and a before-and-after example.

Closes #2499

### DIFF
--- a/pygments/formatters/html.py
+++ b/pygments/formatters/html.py
@@ -887,7 +887,7 @@ class HtmlFormatter(Formatter):
                     yield 1, ''.join(line)
                     line = []
                 elif part:
-                    yield 1, ''.join((cspan, part, (cspan and '</span>'), lsep))
+                    yield 1, ''.join((cspan, part, lsep, (cspan and '</span>')))
                 else:
                     yield 1, lsep
             # for the last line


Anteru:
I'm now a bit unsure if this is an artifact of not recognizing the `\n` as whitespace properly. I've been doing some cleanup of whitespace handling, but I presume the end-of-line character is _special_ in that regard as the processing is line-based. I don't see this causing lots of user impact though so I'm fine with this, but it needs to be mentioned in the changelog (I'll take care of that, no worries.) I'd also prefer to have a specific test case for this behavior, as apparently our thousands of HTML tests didn't catch the change, meaning this is untested behavior to start with.

vkoukis:
@Anteru 
>I'm now a bit unsure if this is an artifact of not recognizing the \n as whitespace properly. I've been doing some cleanup of whitespace handling, but I presume the end-of-line character is special in that regard as the processing is line-based.

Looking at lexer tests under `tests/snippets/console`, I understand the lexer does include newline characters in the `Generic.Output` class, but yes, they are special, and the HTML formatter has at least one option, `lineseparator`, that changes how it terminates lines.

> I don't see this causing lots of user impact though so I'm fine with this, but it needs to be mentioned in the changelog (I'll take care of that, no worries.)

Thank you for taking care of this.

> I'd also prefer to have a specific test case for this behavior, as apparently our thousands of HTML tests didn't catch the change, meaning this is untested behavior to start with.

Happy to extend the tests, I've done it for `console` lexer tests in a related PR, #2500, but I'm not sure where to start with HTML tests. Do we have any infrastructure describing input and expected HTML output pairs I could extend, for example? Thank you for any pointers you may have.

Anteru:
There's some machinery (`structural_diff`) you can use, check out the https://github.com/pygments/pygments/blob/master/tests/test_html_formatter_linenos_elements.py tests. I'd put something similar looking into the `test_html_formatter` for a few select cases (and possibly some fairly stable lexers, so it's easy to regenerate if needed.)

That said, your proposed change will break when `linespans` is set, won't it? Because now the line doesn't terminate any more with a tag, but there's a trailing tag lagging into the next line. I'm not sure there's a way to solve this, and thinking more about it, the table output will be also affected as it has one row per line? This definitely needs some tests to look at the output and figure out what makes the most sense.

vkoukis:
@Anteru 

> There's some machinery (structural_diff) you can use, check out the https://github.com/pygments/pygments/blob/master/tests/test_html_formatter_linenos_elements.py tests. I'd put something similar looking into the test_html_formatter for a few select cases (and possibly some fairly stable lexers, so it's easy to regenerate if needed.)

ACK, I'll look into this test file and how it uses the `structural_diff` mechanism.

> That said, your proposed change will break when linespans is set, won't it? Because now the line doesn't terminate any more with a tag, but there's a trailing tag lagging into the next line. I'm not sure there's a way to solve this, and thinking more about it, the table output will be also affected as it has one row per line? This definitely needs some tests to look at the output and figure out what makes the most sense.

To make sure I am in sync, you're suggesting I test this change with `linespans = True` and with `linenos = "table"` in the options, right? Let me check, I'll send an update here.

Since we agree that the proposed change makes sense and has practical importance [you can actually match the whole output via the `go` class, instead of excluding the newlines], I can look into only enabling it when it actually works, so we don't break existing use cases. But let me look into a few test cases, and we can discuss more.

Anteru:
@vkoukis Did you have any chance to look into this? And yes, my suggestion is to try with `linespans` enabled.

PR #2505, opened 2023-09-11
Add package extra for colorama dependency

jeanas:
(Related to #2492)


Anteru:
Looks good to me, I'll add a note to this effect to the changelog as well.

PR #2509, opened 2023-09-17
Bump checkout action to v4

marcauberer:
Upgrade checkout action to v4 to make use of Node 20.
Node 16, which v3 is using, reached end of life on 2023-09-11.


Anteru:
Merged, thanks!

PR #2513, opened 2023-09-18
autopygmentize: support more file types, use xdg-mime

rrthomas:
Use xdg-mime instead of file.

Includes new languages; in particular, allow application/foo or text/foo for
scripting languages.


rrthomas:
Good point, I'll revert to file.

rrthomas:
I've made the update; thanks for your feedback!

rrthomas:
I have also updated the title of the PR and the commit message to reflect the fact that this PR no longer switches to `xdg-mime` but merely updates MIME types.

PR #2522, opened 2023-09-23
Update CHANGES

vanillajonathan:


Anteru:
I'm going to take care of the Changelog separately -- no need for this PR. Thanks for thinking about it though :)

PR #2528, opened 2023-09-29
added "elseif" as a Fortran keyword

ghammond86:
"elseif" is an supported alternate for "else if" in Fortran.


birkenfeld:
Thanks!

PR #2530, opened 2023-10-02
add vyper lexer

mo-anon:
adding a lexer for vyper: https://docs.vyperlang.org/en/stable/toctree.html


Anteru:
Thanks for the contribution, but you're missing several things (which is why I'm not even running CI as it'll fail right away):

* One example file
* The file must adhere to the coding guidelines/format requirements
* The lexer is missing an `url` attribute

You can check this locally using `tox`, `tox -e check`, to avoid multiple roundtrips through CI.

PR #2531, opened 2023-10-03
adding a lexer for vyper

mo-anon:
adding a new lexer for vyper (smart contract language): https://docs.vyperlang.org/en/stable/

A thing I require some help with: when running `tox -e check` the following error occurs:
_pygments/lexers/vyper.py:4: wrong module name underline, should be ~~~...~_

i am using tilde, so im not really sure whats the issue there.

thanks in advance.

### DIFF
--- a/pygments/lexers/vyper.py
+++ b/pygments/lexers/vyper.py
@@ -38,7 +38,7 @@ class VyperLexer(RegexLexer):
             (r'(def)\s+([a-zA-Z_][a-zA-Z0-9_]*)', bygroups(Keyword, Name.Function)),
 
             # Event and Struct
-            (r'(event|struct|interface)\s+([a-zA-Z_][a-zA-Z0-9_]*)', bygroups(Keyword, Name.Class)),
+            (r'(event|struct|interface|log)(\s+)([a-zA-Z_][a-zA-Z0-9_]*)', bygroups(Keyword, Text.Whitespace, Name.Class)),
 
             # Imports
             (r'(from)\s+(vyper\.\w+)\s+(import)\s+(\w+)', bygroups(Keyword, Name.Namespace, Keyword, Name.Class)),
--- a/tests/snippets/vyper/test.txt
+++ b/tests/snippets/vyper/test.txt
@@ -1,79 +1,24 @@
 ---input---
-interface Factory:
-    def fee_receiver() -> address: view
-    def admin() -> address: view
-    def views_implementation() -> address: view
+log Transfer
 
-interface ERC1271:
-    def isValidSignature(_hash: bytes32, _signature: Bytes[65]) -> bytes32: view
+event Transfer:
+
+struct Tessst:
 
 ---tokens---
-'interface'   Keyword
-'Factory'     Name.Class
-':'           Punctuation
-'\n    '      Text.Whitespace
-'def'         Keyword
-'fee_receiver' Name.Function
-'('           Punctuation
-')'           Punctuation
+'log'         Keyword
 ' '           Text.Whitespace
-'-'           Operator
-'>'           Operator
-' '           Text.Whitespace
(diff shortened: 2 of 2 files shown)


Anteru:
That problem is fairly easy: You have:

```
    pygments.lexers.vyper
    ~~~~~~~~~~~~~~~~~~
```

The length of the underline must match the text above, so you're missing a few more `~`. I'll see if I can improve the script error for that.

pygments/lexers/vyper.py
```diff
@@ -0,0 +1,129 @@
...
+            (r"'([^'\\]*(?:\\.[^'\\]*)*)'", String.Single),
+            (r'"([^"\\]*(?:\\.[^"\\]*)*)"', String.Double),
+
+            # Numeric Literals
+            (r'\b0x[0-9a-fA-F]+\b', Number.Hex),
+            (r'\b\d{1,3}(?:_\d{3})*\b', Number.Integer),
+            (r'\b\d+\.\d*\b', Number.Float),
+
+            # Keywords
+            (r'\b(def|event|pass|return|for|while|if|elif|else|assert|raise|import|in|struct|implements|interface|from)\b', Keyword),
```
Anteru, line 41: Use the `words` function for this which creates an optimized regex (and the other cases below as well.)

pygments/lexers/vyper.py
```diff
@@ -0,0 +1,129 @@
...
+            (r'\b[a-zA-Z_]\w*\b:', Name.Variable),
+            (r'\b[a-zA-Z_]\w*\b', Name),
+
+            # Error for unmatched sequences
+            (r'.', Text)
+        ],
+
+        'multiline-comment': [
+            (r'\"\"\"', Comment.Multiline, '#pop'),
+            (r'.|\n', Comment.Multiline)
```
Anteru, line 92: This should eat more than one character at a time. You probably want something like `[^\"]+` or similar, otherwise, you're going to generate one token per character. See: https://pygments.org/docs/lexerdevelopment/, last comment

pygments/lexers/vyper.py
```diff
@@ -0,0 +1,129 @@
...
+            (r'#.*$', Comment.Single)
+        ],
+
+        'funcparams': [
+            (r'\)', Punctuation, '#pop'),
+            (r'\w+', Name.Variable),
+            (r':', Punctuation),
+            (r'\w+', Keyword.Type),
+            (r',', Punctuation),
+            (r'\s+', Text.Whitespace),
```
Anteru, line 125: This is the first time I'm seeing someone rely on the fact that the lexer will remember which state it was last in and continue in order.

@birkenfeld , @jeanas : Should we document this better on https://pygments.org/docs/lexerdevelopment/ ? 

It currently says:

> States are groups of regular expressions that are matched against the input string at the current position. If one of these expressions matches, a corresponding action is performed (such as yielding a token with a specific type, or changing state), the current position is set to where the last match ended and the matching process continues with the first regex of the current state.

Which I feel should be: "The process continues using the next rule (in the order of the rule definitions in the lexer source.)" or something, to make this clear. But maybe it's just me given the contributor here figured this out on their own.

Anteru (changes requested):

pygments/lexers/vyper.py
birkenfeld: That's not how it works though? I haven't looked at the source right now but it is intended to work exactly like in the docs, restart at first rule of the state.

If you have a sequence of rules to be matched in order, you should use one rule and bygroups.

Anteru: Ok, then I'm not loosing my mind here. I'll open a PR to clarify this.

Anteru: Also, this needs to be fixed, @mo-anon

PR #2535, opened 2023-10-06
Change `gdscript_example.gd` indentation to tabs

Repiteo:
GDScript by default will expect tabs instead of spaces, and the `snippets` files respect this. `examplefiles` was an exception to this, using spaces instead of tabs. This PR rectifies that & updates the `.output` file accordingly


Anteru:
Thank you, good catch!

PR #2536, opened 2023-10-08
New lexer for the Ursa programming language

rrthomas:
See: https://ursalang.github.io

This is a new programming language with a syntax similar to JavaScript, on which I based the lexer patterns. I already offer the lexer as a custom lexer for Pygments; it would be nice to have it available e.g. on Rosetta Code, which uses Pygments for syntax highlighting.


Anteru:
Also covered here now in the goals/non-goals section: https://pygments.org/docs/contributing/

rrthomas:
Thanks, I'll go the plugin route for now. I would say that I'm leaning hard on the developer experience; I already offer an Emacs mode, and patterns for GEdit and Pico, as well as tree-sitter support that is available via syntastica. I quite understand that a mature project like Pygments may want to wait, and I'll be back later, I hope.

PR #2545, opened 2023-10-22
Put the lean lexer in its own file

eric-wieser:
In the near future it is likely that more lean lexers will be added for:
* Lean 4
* Lake

As preparation for this, it makes sense to have these in their own file. To avoid any breakage, `LeanLexer` is still re-exported by `lexers.theorem`.

I have not made any modifications to the lexer.


Anteru:
Merged, thanks!

PR #2546, opened 2023-10-22
Make `lean3` an alias of Lean

eric-wieser:
At some point there might be a push to make `lean` refer to Lean 4. This is consistent with how `python2` is used for Python 2 code.


Anteru:
This is broken now as I merged the other lean change. Would you mind rebasing this?

eric-wieser:
Done

eric-wieser:
@jeanas: does the commit I just pushed look ok?

eric-wieser:
Done; this results in the mapping file being updated too, but I guess that's ok?

eric-wieser:
Adding such an alias to `__all__` breaks the test. I could switch to `class LeanLexer(Lean3Lexer): pass` if you prefer?

eric-wieser:
I think I found the fix

eric-wieser:
Thanks!

PR #2561, opened 2023-11-01
Clea(r|n)er file for the Dracula style (no style changes)

dbitouze:
Here is a clea(r|n)er file for the Dracula style, with no changes to the style by itself. 

BTW, this style doesn't follow the [official `Dracula` style specifications](https://spec.draculatheme.com/#). For instance:

- [`Error`](https://github.com/pygments/pygments/blob/master/pygments/styles/dracula.py#L55)  should be [`red`](https://spec.draculatheme.com/#Error) and not `#f8f8f2` (`foreground`),
- [`Generic.Deleted`](https://github.com/pygments/pygments/blob/master/pygments/styles/dracula.py#L43), corresponding to `Dracula`'s `DiffDeleted` should be [`red`](https://spec.draculatheme.com/#DiffDeleted) and not `#8b080b`.

Is it worth to provide an “official” `Dracula` style?

### DIFF
--- a/pygments/styles/dracula.py
+++ b/pygments/styles/dracula.py
@@ -22,86 +22,67 @@ __all__ = ['DraculaStyle']
 class DraculaStyle(Style):
     name = 'dracula'
 
-    background_color = "#282a36"
-    highlight_color = "#44475a"
-    line_number_color = "#f1fa8c"
-    line_number_background_color = "#44475a"
-    line_number_special_color = "#50fa7b"
-    line_number_special_background_color = "#6272a4"
+    background = "#282A36"
+    foreground = "#F8F8F2"
+    selection  = "#44475A"
+    comment    = "#6272A4"
+    cyan       = "#8BE9FD"
+    green      = "#50FA7B"
+    orange     = "#FFB86C"
+    pink       = "#FF79C6"
+    purple     = "#BD93F9"
+    red        = "#FF5555"
+    yellow     = "#F1FA8C"
+
+    deletion   = "#f1fa8c"
 
     styles = {
-        Whitespace: "#f8f8f2",
+        Whitespace             : foreground         ,
 
-        Comment: "#6272a4",
-        Comment.Hashbang: "#6272a4",
-        Comment.Multiline: "#6272a4",
-        Comment.Preproc: "#ff79c6",
-        Comment.Single: "#6272a4",
-        Comment.Special: "#6272a4",
+        Comment                : comment            ,
+        Comment.Preproc        : pink               ,
 
-        Generic: "#f8f8f2",
-        Generic.Deleted: "#8b080b",
-        Generic.Emph: "#f8f8f2 underline",
-        Generic.Error: "#f8f8f2",
-        Generic.Heading: "#f8f8f2 bold",
-        Generic.Inserted: "#f8f8f2 bold",
-        Generic.Output: "#44475a",
-        Generic.Prompt: "#f8f8f2",
-        Generic.Strong: "#f8f8f2",
-        Generic.EmphStrong: "#f8f8f2 underline",
-        Generic.Subheading: "#f8f8f2 bold",
-        Generic.Traceback: "#f8f8f2",
+        Generic                : foreground         ,
+        Generic.Deleted        : deletion           ,
+        Generic.Emph           : "underline"        ,
+        Generic.Heading        : "bold"             ,
+        Generic.Inserted       : "bold"             ,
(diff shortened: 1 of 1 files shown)


dbitouze:
> Thanks, but this is actually a much less standard Python style. PEP 8 [explicitly recommends against it](https://peps.python.org/pep-0008/#whitespace-in-expressions-and-statements).

[Is it](https://github.com/pygments/pygments/compare/master...dbitouze:pygments:master) better now?

dbitouze:
> You removed some class attributes, why?

Because of (see https://pygments.org/docs/styledevelopment/):

> When you define a style for Name the style automatically also affects Name.Function and so on.

For instance, in [`Comment` token and subtokens styles](https://github.com/pygments/pygments/blob/0fba7b8a1dfc905e71c70ae7c29b78ce5baed3e8/pygments/styles/dracula.py#L35-L40), only `Comment.Preproc` subtoken style differs from the `Comment` (parent) token style so other subtokens styles unnecessarily clutter the file (IMHO).

dbitouze:
> @dbitouze This isn't what "class attributes" is about: it refers to `background_color`, `line_number_color` and such. These are used by the `Style` class.

Ah, sorry. I'll reintroduced them.

dbitouze:
> Thanks, but this is actually a much less standard Python style. PEP 8 [explicitly recommends against it](https://peps.python.org/pep-0008/#whitespace-in-expressions-and-statements).

BTW, many existing styles, including the `default` one, don't follow the less standard Python style.

birkenfeld:
> BTW, many existing styles, including the `default` one, don't follow the standard Python style.

There's a difference between
```
Name     : "bold"    ,
String   : "italic"  ,
```
and
```
Name:     "bold",
String:   "italic",
```
in that the latter doesn't introduce spaces where none belong, only adds additional ones for alignment. This is ok for us.

dbitouze:
> In this style, e.g. `line_number_special_color` is defined as `"#50fa7b"` but `"#50fa7b"` (instead of `line_number_special_color`) is used at several places. Why?

So, since `line_number_special_color` is first defined as `"#50fa7b"`, is it good practice to use `line_number_special_color` instead of `"#50fa7b"` in the rest of the file?

birkenfeld:
> So, since `line_number_special_color` is first defined as `"#50fa7b"`, is it good practice to use `line_number_special_color` instead of `"#50fa7b"` in the rest of the file?

The name `line_number_special_color` is not chosen by the style but Pygments requires a class attribute with this name for a special purpose (coloring special line numbers).

So unless you want to put emphasis on the fact that e.g. class names are using the same color as special line numbers, using that name there is more confusing than helpful.

If you want to name the colors (which is a good idea), do it like you did and then assign
```
line_number_special_color = green
```
below.

For even more clarity, move your color constants out of the class and make them module level constants.

dbitouze:
So, if I understand correctly, something like that:
```python
[...]
__all__ = ['DraculaStyle']

background = "#282a36"
foreground = "#f8f8f2"
selection = "#44475a"
comment = "#6272a4"
cyan = "#8be9fd"
green = "#50fa7b"
orange = "#ffb86c"
pink = "#ff79c6"
purple = "#bd93f9"
red = "#ff5555"
yellow = "#f1fa8c"

deletion = "#8B080B"

class DraculaStyle(Style):
    name = 'dracula'

    background_color = background
    highlight_color = selection
    line_number_color = yellow
    line_number_background_color = selection
    line_number_special_color = green
    line_number_special_background_color = comment

    styles = {
        [...]
```
is okay, isn't it?

dbitouze:
I hope it's okay now :crossed_fingers:

What about [an “official” `Dracula` style](https://github.com/pygments/pygments/pull/2561#issue-1972485728)?

dbitouze:
Sorry! I checked on my local repo and it wrongly looked OK. Should be fixed now.

dbitouze:
> I believe that `italic` has been lost on `Token.Name.Builtin`.

Sorry for that! I carefully checked and it should be okay now. 

I also removed the style specifications of some subtokens. For instance:
```python
Name.Variable.Class
Name.Variable.Global
Name.Variable.Instance
```
the style of which being identical to the style of their parent `Name.Variable`. I hope it's OK.

dbitouze:
Scripts are better than humans, it's a shame.... :wink: Thanks! About the `master` branch being protected (sorry about that: I usually don't encounter this kind of problem), I created another branch (with the `Name.Builtin.Pseudo` fix) : I hope it's OK now.

PR #2563, opened 2023-11-03
Fix get_formatter_for_filename on plugin formatters

jeanas:
find_plugin_formatters returns (name, cls) tuples.


Anteru:
Is there a way to unit test this? I'm surprised this has been around for so long. Also, does this mean no plugin formatters worked previously?

PR #2564, opened 2023-11-05
Make wcag-contrast-ratio module optional.

mcepl:


Anteru:
Without any explanation as to why, we're going to close this issue soon.

mcepl:
On Sun Nov 5, 2023 at 7:35 PM CET, Matthäus G. Chajdas wrote:
> Without any explanation as to why, we're going to close this issue soon.

It will probably remain just a distro-specific patch, just that
for our enterprise distribution we want to eliminate as many
unnecessary packages as possible. We want to run at least basic
test suite for all packages (it is kind of frown upon when we
deliver in a package non-functional junk), but it doesn’t have
to be full one (when access to the Internet or some specialized
packages are required). We don’t need wcag-contrast-ratio for
anything else than this one test, so I feel right to eliminate
it.

The patch as it stands right now is completely harmess to anybody
who has the wcag-contrast-ratio package installed, it just makes
it optional.

-- 
http://matej.ceplovi.cz/blog/, @***@***.***
GPG Finger: 3C76 A027 CA45 AD70 98B5  BC1D 7920 5802 880B C9D8
 
Books aren’t written - they’re rewritten. Including your own. It
is one of the hardest things to accept, especially after the
seventh rewrite hasn’t quite done it.
  -- Michael Crichton, alluding to Steele MacKaye (1889) article
     where he said this about theater plays.

Anteru:
Sure, but that adds a complexity for us, as we may want to use other packages in the future. Our recommended way to test is using `tox` which can install any package for testing. I don't think we want to commit to making all packages optional during testing and we wouldn't even know where to draw the line. As it stands right now, while harmless, the patch is confusing at best (no explanation in the source code as to why this package is special) and it create a maintenance promise/burden for an unspecified requirement. I'm not a fan of that to say the least.

Also, if a style fails the WCAG contrast test, your local copy of Pygments may be actually non-functional junk, as you wouldn't know someone changed all colors to solid white for example. So it's one of the more important tests in my book.

mcepl:
Cool, then I just step out.

PR #2567, opened 2023-11-06
Update layout.html

kinlearn:
Copyright year updated to 2023 from 2022


Anteru:
Thanks, good catch. And just before we have to change it again :)

kinlearn:
Yeah! 😅

PR #2573, opened 2023-11-07
Use hatchling as build backend

jeanas:
This reduces tox's setup time, as measured on an environment with no commands, from ~1.7s to ~0.8s.

It also fixes requirements.txt and tox.ini not being included in sdists in spite of being part of the source tree.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-*.egg
 *.pyc
 *.pyo
 .*.sw[op]
@@ -8,7 +7,6 @@
 /.tags
 /.tox/
 /.cache/
-/Pygments.egg-info/*
 /TAGS
 /build/*
 /dist/*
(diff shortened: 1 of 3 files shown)


birkenfeld:
Feels definitely snappier. Is hatchling commonly enough used so that this doesn't create problems for downstream packagers?

birkenfeld:
Good to go then!

Anteru:
Ok, I just tried this and `python -m build` uses `hatchling` now (apparently without having to install anything manually), so that's good. Does this impact the build itself in any way? For me at the very least it changed the capitalization of the output files -- i.e. it produced lower-case `pygments` now. If you look at the release checklist, that breaks the command line there. I can obviously rename it manually, but is there a way to teach `hatchling` about this?

Anteru:
Lower-case is totally fine for me, it's just something that needs to be mentioned in the changelog because there must be someone out there using a script which expects it upper-case.

I don't know what's up with WASM. @birkenfeld ?

Anteru:
Indeed it does :)

PR #2576, opened 2023-11-08
Revise TOML lexer

### LINKED ISSUE #2488: TOML: Multi-line string lexing issue when using same quote character
Using the TOML syntax highlighter breaks in multi-line strings when re-using the same quote character in the text body:

```
negative_reasoning = """The customer uses the 😤 emoji and seems aggravated,
so the sentiment is negative."""
negative_output = """
{
  "negative": [
    {
      "date": "2023-07-24",
      "conversation": [
        "A: What can I help you with?",
        "C: I CAN'T CONNECT TO MY 😤 ACCOUNT",
        "A: Are you sure it's not your caps lock?",
        "C: 😤! You're right!"
      ]
    }
  ]
}
"""
```

Renders like so:

image

jeanas:
The new lexer matches the TOML spec much more closely.

User-visible differences should be these:

* Add MIME type
* Highlight string escapes
* Recognize \uXXXX and \UXXXX escapes
* Also recognize booleans if they are followed by a comment
* Fix single quotes inside multiline literal strings (closes #2488)
* Prevent multiline literal strings from eating comments
* Add multiline basic strings (""")
* Improve datetime recognition: recognize times without dates, dates without times and datetimes without time zone; allow sub-millisecond precision
* Recognize floats with exponents (they used not to be recognized when having a decimal point)
* Recognize binary, octal and hex literals
* Recognize strings inside table headers
* Recognize table headers followed by comments
* Don't parse sequences of digits as integers when they are actually keys

Includes several new tests, most of which were not working before.

### DIFF
--- a/pygments/lexers/configs.py
+++ b/pygments/lexers/configs.py
@@ -1114,54 +1114,168 @@ class AugeasLexer(RegexLexer):
 
 class TOMLLexer(RegexLexer):
     """
-    Lexer for TOML, a simple language
-    for config files.
+    Lexer for TOML, a simple language for config files.
 
     .. versionadded:: 2.4
     """
 
     name = 'TOML'
-    url = 'https://github.com/toml-lang/toml'
     aliases = ['toml']
     filenames = ['*.toml', 'Pipfile', 'poetry.lock']
+    mimetypes = ['application/toml']
+    url = 'https://toml.io'
+
+    # Based on the TOML spec: https://toml.io/en/v1.0.0
+
+    # The following is adapted from CPython's tomllib:
+    _time = r"\d\d:\d\d:\d\d(\.\d+)?"
+    _datetime = rf"""(?x)
+                  \d\d\d\d-\d\d-\d\d # date, e.g., 1988-10-27
+                (
+                  [Tt ] {_time} # optional time
+                  (
+                    [Zz]|[+-]\d\d:\d\d # optional time offset
+                  )?
+                )?
+              """
 
     tokens = {
         'root': [
-            # Table
-            (r'^(\s*)(\[.*?\])$', bygroups(Whitespace, Keyword)),
+            # Note that we make an effort in order to distinguish
+            # moments at which we're parsing a key and moments at
+            # which we're parsing a value. In the TOML code
+            #
+            #   1234 = 1234
+            #
+            # the first "1234" should be Name, the second Integer.
 
-            # Basics, comments, strings
-            (r'[ \t]+', Whitespace),
-            (r'\n', Whitespace),
-            (r'#.*?$', Comment.Single),
-            # Basic string
-            (r'"(\\\\|\\[^\\]|[^"\\])*"', String),
-            # Literal string
-            (r'\'\'\'(.*)\'\'\'', String),
-            (r'\'[^\']*\'', String),
-            (r'(true|false)$', Keyword.Constant),
-            (r'[a-zA-Z_][\w\-]*', Name),
+            # Whitespace
+            (r'\s+', Whitespace),
 
-            # Datetime
-            # TODO this needs to be expanded, as TOML is rather flexible:
-            # https://github.com/toml-lang/toml#offset-date-time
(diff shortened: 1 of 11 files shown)


pygments/lexers/configs.py
```diff
@@ -1114,54 +1114,168 @@ class AugeasLexer(RegexLexer):
...
+        'table-key': [
+            # This is like 'key', but highlights the name components
+            # and separating dots as Keyword because it looks better
+            # when the whole table header is Keyword. We do highlight
+            # strings as strings though.
+            # Start of bare key (only ASCII is allowed here).
+            (r'[A-Za-z0-9_-]+', Keyword),
+            (r'"', String.Double, 'basic-string'),
+            (r"'", String.Single, 'literal-string'),
+            (r'\.', Keyword),
```
birkenfeld, line 1185: Any reason not to put the dot in the above char class?

birkenfeld (approved):
Looks good!

Anteru:
Looks good. Thanks! I'll try to wrap up a new release this or next weekend.

pygments/lexers/configs.py
```diff
@@ -1114,54 +1114,171 @@ class AugeasLexer(RegexLexer):
...
+            # End of inline table
+            (r'\}', Punctuation, '#pop'),
+        ],
+        'basic-string': [
+            (r'"', String.Double, '#pop'),
+            include('escapes'),
+            (r'[^"\\]+', String.Double),
+        ],
+        'literal-string': [
+            (r".*'", String.Single, '#pop'),
```
ilevkivskyi, line 1264: I think this is a bug. This will capture too much if there is a literal string followed by a comment containing `'`. This broke mypy docs build on this line:
```
              'two\.pyi$',  # but TOML's single-quoted strings do not
```
see https://github.com/python/mypy/actions/runs/6916132945/job/18815845352

ilevkivskyi: Thanks, NP!

Anteru: Oh dear. I'll try to get it done today, somehow. I'm always afraid this happens, and our release process is still fairly manual :( Goals for 2024 I guess.

birkenfeld: Let me know if you want me to step in. Also, what would you like to automate?

Anteru: Things I'd like to automate:

* From tag to PyPi -- ideally, to test-pypi on every tagged commit (https://github.com/marketplace/actions/pypi-publish) -- ideally, the actual publication would be a special action I just click on. It's not that it takes a lot of time, but I'm always nervous I mess something up with the command line, forget to delete a file, git clean, etc. -- I very diligently work through the `release-checklist` to avoid that. Literally signing things off.
* Auto-formatting -- I tend to clean up the formatting of the lexers every time close to release, at least the worst offenders. I use `autopep8` at the moment, would rather apply `flake8` or `black` on the entire codebase.
* Auto-check that the new arguments like URL etc. are present on new lexers
* Auto-check `.. versionadded:: ` is there -- costs me _a lot of time_ to open up every Lexer close to release and make sure it's present and in the right format (i.e. `2.17.0` vs. `2.17`)
* Actually get all checks working/passing (i.e. the additional checkers I wrote and possibly PyLint). `check_whitespace_tokens` and `check_repeated_tokens` need an expected-fail list so we can whitelist currently existing lexers until we fix those, but new lexers should always pass those tests.
* Verify all PR numbers closed/merged since last release are mentioned in the `CHANGES` file. I'm pretty good at assigning tasks to milestones now, but I still miss things in the `CHANGES` file, and it's super time consuming to open 100 tabs, go through each item one-by-item, check the PR number/issue number is present, etc. If there was a way to auto-generate the changelog that would be even better, but my experience is that those look pretty ugly and some manual checkup is fine.

I'll get to the release in a moment, thanks for the offer though!

ilevkivskyi: Thank you for the quick fix! :+1:

PR #2580, opened 2023-11-11
Update .dockerignore

jeanas:
Ignoring .rst files made the CI fails because unlike setuptools, hatchling gives an error if the description file is not found.

Also remove files that were generated by setuptools.


Anteru:
Looks good, thanks!

birkenfeld:
Thanks!

PR #2586, opened 2023-11-19
Yet another formatting fix in CHANGES

muzimuzhi:
Three backticks were used, while only two were needed.

Introduced in commit 5874c34 (Fix code formatting in the CHANGES file., 2023-11-18).


Anteru:
Thanks!

Issue #2585, opened 2023-11-19
Release automation


jeanas:
Things I'd like to automate:

* From tag to PyPi -- ideally, to test-pypi on every tagged commit (https://github.com/marketplace/actions/pypi-publish) -- and the actual release would be a special action I just click on. It's not that it takes a lot of time, but I'm always nervous I mess something up with the command line, forget to delete a file, git clean, etc. -- I very diligently work through the `release-checklist` to avoid that. Literally signing things off.
* Auto-formatting -- I tend to clean up the formatting of the lexers every time close to release, at least the worst offenders. I use `autopep8` at the moment, would rather apply `flake8` or `black` on the entire codebase.
* Auto-check that the new arguments like URL etc. are present on new lexers
* Auto-check `.. versionadded:: ` is there -- costs me _a lot of time_ to open up every Lexer close to release and make sure it's present and in the right format (i.e. `2.17.0` vs. `2.17`)
* Actually get all checks working/passing (i.e. the additional checkers I wrote and possibly PyLint). `check_whitespace_tokens` and `check_repeated_tokens` need an expected-fail list so we can whitelist currently existing lexers until we fix those, but new lexers should always pass those tests.
* Verify all PR numbers closed/merged since last release are mentioned in the `CHANGES` file. I'm pretty good at assigning tasks to milestones now, but I still miss things in the `CHANGES` file, and it's super time consuming to open 100 tabs, go through each item one-by-item, check the PR number/issue number is present, etc. If there was a way to auto-generate the changelog that would be even better, but my experience is that those look pretty ugly and some manual checkup is fine.

I'll get to the release in a moment, thanks for the offer though!

_Originally posted by @Anteru in https://github.com/pygments/pygments/pull/2576#discussion_r1398348679_

Anteru:
Yes. The problem is that we haven't been doing this consistently though, so I'd like _one_ file which contains all exclusions. I.e. run this one over all lexers, find which checkers work/don't work, then write it out with all exclusions, so we can run it for new ones while reducing the exclusion list. That requires however unifying all the checkers somehow or port them to the test framework, and I haven't had enough time to figure out how to integrate all of this with `pytest`. I'm sure it's doable.

birkenfeld:
Yes, Ruff is really good but its linting only completely replaces pyflakes for now. pylint catches way more things (also way more false positives, sigh). A clean Ruff run could be part of CI.

[...]

PR #2588, opened 2023-11-19
Add URLs to lexers and require the `url` attribute

jeanas:
Partially addresses https://github.com/pygments/pygments/issues/2585


Anteru:
Woah! I didn't expect that kind of solution. This looks good, thanks for digging through all of them.

birkenfeld:
Indeed!

PR #2589, opened 2023-11-19
Add and require `version_added` lexer attribute

jeanas:
Next step for #2585 :)

- MXML: Remove bogus mimetypes line
- Savi: Fix indentation
- Move versionadded data to a lexer attribute 
  
  That way, we can set it to "" for old lexers, and check that it's
  present on new lexers. (In the future, we might also use it for better
  presentation in the documentation.)

- Make documentation use the new version_added lexer attribute
- Require the version_added lexer attribute
- Harmonize version_added format

### DIFF
--- a/pygments/lexers/arturo.py
+++ b/pygments/lexers/arturo.py
@@ -30,7 +30,7 @@ class ArturoLexer(RegexLexer):
     aliases = ['arturo', 'art']
     filenames = ['*.art']
     url = 'https://arturo-lang.io/'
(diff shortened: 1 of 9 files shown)


birkenfeld:
Very nice!

Anteru:
Looks good to me, thanks! I assume finding the ones which are set to `""` at the moment is a question of Git history analysis? After all, they must have been added at some point (not that I think it's worth fixing for now.)

Anteru (approved):

PR #2592, opened 2023-11-20
Replace pyflakes with Ruff

jeanas:
- Fix/ignore Ruff lints
- Replace pyflakes with Ruff in `tox -e check`

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -2,13 +2,16 @@
 # Pygments documentation build configuration file
 #
 
-import re, sys, os, itertools
+import os
+import re
+import sys
 
 # If extensions (or modules to document with autodoc) are in another directory,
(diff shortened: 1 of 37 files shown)


birkenfeld:
You're on a roll! Looks good.

Anteru:
I was going to say, taking a week off or something :) Thanks, looks good. Makes me look really bad here though :(

Anteru:
This is weird. When I run `ruff check .`, I get 1301 errors, unless I use `ruff check --ignore E501 .`. Why does `tox` not complain? It supposedly runs `ruff check .` unless I'm missing something?

birkenfeld:
> I was going to say, taking a week off or something :) Thanks, looks good. Makes me look really bad here though :(

How does it make it you look bad? This is not a competition :)

PR #2593, opened 2023-11-21
fix building from source on macos

branchvincent:
Since #2573 switched to `hatchling`, macOS users cannot build from source:

```console
$ python3 -m venv .venv
$ .venv/bin/pip install --no-binary=pygments pygments
$ .venv/bin/python3 -c 'import pygments'
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'pygments'
```

This is because the wheel gets packaged as `Pygments` (prob due to macOS's case-insensitive filesystem). This explicitly tells `hatchling` that the package is `pygments`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,6 +54,9 @@ Changelog = "https://github.com/pygments/pygments/blob/master/CHANGES"
 [project.scripts]
 pygmentize = "pygments.cmdline:main"
 
+[tool.hatch.build]
+packages = ["pygments"]
+
 [tool.hatch.version]
 path = "pygments/__init__.py"
 


branchvincent:
It's not broken for everyone, just macOS users building from source (i.e. using `--no-binary`, which is very few people but does include Homebrew. See https://github.com/Homebrew/homebrew-core/pull/154759, and our failing CI [here](https://github.com/Homebrew/homebrew-core/actions/runs/6939336503/job/18876565416)).

Anteru:
Time for a macOS CI builder? Also, this is weird, because the file generated is `pygments` now, no longer `Pygments` (at least on Linux) and I noticed the case change. Why would macOS `hatchling` suddenly produce an upper-case `Pygments`?

branchvincent:
> Why would macOS hatchling suddenly produce an upper-case Pygments?

So the sdist is correctly cased as `pygments`, but when `hatchling` builds a wheel from that sdist it infers the package name by first checking if `Pygments/__init__.py` is a file and then later `pygments/__init__.py` [here](https://github.com/pypa/hatch/blob/a0f2bd6d8d87d22d9166db49e3eee7467da7acba/backend/src/hatchling/builders/wheel.py#L180). That works on Linux but on a case-insensitive filesystem like macOS, the former is "found" and thus hatchling uses it

birkenfeld:
Sounds like a bug in hatchling...

birkenfeld:
> > Time for a macOS CI builder?
> 
> Yes, probably.

Are they still unproportionally slow? If yes, I'd vote to only use it selectively before release.

Anteru:
We could configure one to only run on "tagged" builds. I think GitHub allows for that.

Anteru:
@jeanas Do I get it right we need a 2.17.2 to fix the packaing issue on macOS as currently there's nothing working on macOS? Can't they use the package from PyPI? I don't have a mac so I don't understand the extent of the issue. Not being able to build Pygments on macOS doesn't seem to warrant a hotfix if they can still use it.

birkenfeld:
IMO this kind of bug is an annoyance that should be addressed by a release, especially since it's a regression caused by us switching the build system.

birkenfeld:
That said, I have no objections to just picking the single commit onto `2.17.1` for the release, instead of releasing from current `main`.

Anteru:
I do think it should be fixed in the next release, I'm arguing if this warrants 2.17.2 on its own if PyPI users can use Pygments using the recommended installation method. If it's "much more rare" -- is it rare enough to meet whatever arbitrary bar we have for a point release, or is it too rare to warrant a fix now? I understand it's a regression, but it seems a fairly minor one.

If we do agree that 2.17.2 it is, then yes, I'll branch it off 2.17.1, and we'll cherry-pick that commit over. I would argue though that we should have a CI test for this to make sure we're actually not breaking anything while doing so. I have no access to a mac machine to try myself, and I feel somewhat uncomfortable about this. Sure, I can merge this PR and believe everyone it's working, but having a way to actually verify this would be much better.

Can I get a quick vote on 2.17.2 yay/nay for this one fix only? My vote is no, fold into 2.18, but we strategically decided  to have three maintainers so you can overrule me.

Anteru:
Ok. Getting a 2.17.2 out now.

Anteru:
Something is broken. I'm building the branch `release/2.17.x`, the wheel size looks fine, but the `tar.gz` is now 983 KiB instead of the expected 4.8 MiB :(

Anteru:
Thanks. I added some more information. If I had to guess, the packages statement makes it only include the `pygments` folder.

Anteru:
@birkenfeld Not a chore per se, more like a wild ride with no safety belt and on a rusty looping designed by someone playing Rollercoaster Tycoon.

Anteru:
Ok. I'll cherry-pick that as well.

branchvincent:
can confirm it works for us, thanks everyone!

ofek:
Sorry for the trouble everyone! Here's an update: https://github.com/pypa/hatch/issues/1054#issuecomment-1837559305

PR #2598, opened 2023-11-28
Use the token type `Comment`, not `Comment.Preproc`, for RST comments

kurtmckee:
While documenting a project that interacts with headers across different file formats, I noticed that RestructuredText comments rendered differently from the others:

image

For clarity: the difference between RestructuredText comments and other lexers' comments is that all other comments are styled as blue text and italicized, but RestructuredText uses green text and is un-italicized.

The root cause is that RestructuredText comments are given the token type `Comment.Preproc`, which is styled uniquely by most styles that I reviewed in Pygments.

This PR modifies the token type for RST comments to `Comment`. This results in rendering that matches my expectations:

image

Please let me know if anything is amiss here! As always, thanks for your work on Pygments!


birkenfeld:
Yeah, there's no good reason for this. Thanks!

Anteru:
It's not a preprocessor anyways, so yes, this is the right change.

Anteru:
Too slow this morning. More coffee needed apparently to beat @birkenfeld on the job :(

PR #2601, opened 2023-12-01
Drop support for EOL Python 3.7

hugovk:
Dropping 3.7 means the https://pypi.org/project/importlib-metadata and `pkg_resources` fallbacks can be removed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@ build-backend = "hatchling.build"
 [project]
 name = "Pygments"
 dynamic = ["version"]
-requires-python = ">=3.7"
+requires-python = ">=3.8"
 license = {text = "BSD-2-Clause"}
 authors = [
   {name = "Georg Brandl", email = "georg@python.org"}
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -41,7 +40,6 @@ classifiers = [
 ]
 
 [project.optional-dependencies]
-plugins = ["importlib-metadata;python_version<'3.8'"]
```
hugovk, line 44: I wonder if this should be kept as a dummy, empty `plugins` extra, so people can still do `pip install pygments[plugins]`?

And then remove it in the next major version?

Anteru:
Thanks, high time to revisit this (and also update the link in our docs ...)

@birkenfeld I just checked: RHEL9 ships with 3.9, Ubuntu 22.04 uses 3.10, and Debian Bookworm is Python 3.11. So we could technically drop 3.7 as per our own guidelines (https://pygments.org/docs/contributing/) -- any concerns?

birkenfeld:
👍

pyproject.toml
hugovk: Re-added as a no-op.

PR #2602, opened 2023-12-01
Add val and var as xtend keywords

ebousse:
The current [Xtend](https://eclipse.dev/Xtext/xtend/) support does not highlights two important keywords of the language, namely `val` and `var`, which allow declaring variables.

Example:

```xtend
class Hello {
   def sayHello() {
      val message = "Hello world"
      println(message)
   }
}
```
When processed with Pygments, we get:

image

This PR fix this problem by adding these two keywords to the appropriate list.


Anteru:
Thank you, good find!

Issue #1111, opened 2019-08-31
Line breaks instead of paragraph


Anteru:
(Original issue 1407 created by gargoylesoft on 2017-12-20T06:49:57.823421+00:00)

I'm copying a section of code from Xcode and then running this command in the terminal:

pbpaste | pygmentize -l swift -f rtf -O style=xcode | pbcopy -Prefer rtf

When I then paste into Pages for Mac all of the newlines are being set as Paragraph inserts instead of line inserts, so I have to manually go and change them all.  Is there a way to fix this from pygments?

akraus53:
I also have this issue, currently looking into why this happens

akraus53:
So it seems (!!!!) this is due to the `\sa` keyword not being used. This stands for "space after" (in twips) and defaults to 200, which is 10pt. if we set `\sa0` in [line 116 of the formatter](https://github.com/pygments/pygments/blob/0f3ddb3a6e3ed99957fe20aab695446f85835387/pygments/formatters/rtf.py#L116) this issue doesn't occur anymore. If anyone can double-check this again, I'd be thankful tho.

akraus53:
Update: As it seems, most editors follow the default "\sa0" which is referenced in this beautiful [documentation](https://www.biblioscape.com/rtf15_spec.htm) but Powerpoint doesn't. Setting this default explicitly will not have any negative impact on the functionality. I'll create a pull request really quick.

Anteru:
Fixed in 544e58c5f356fdb7b58b504ae088442f6c81bdcb

PR #2607, opened 2023-12-07
[Waiting for #2654] add the `\sa0` keyword to the rtf formatter

akraus53:
This sets a default explicitly, which is set differently by some rtf displayers, e.g. MS PowerPoint. Here, \sa200 is assumed which adds a 10pt spacing after every newline, thus making the code very spaced out. For further discussion, see #1111. This issue is solved by this PR.

Edit: I referenced the wrong issue...

### DIFF
--- a/pygments/formatters/rtf.py
+++ b/pygments/formatters/rtf.py
@@ -113,7 +113,7 @@ class RtfFormatter(Formatter):
                         int(color[4:6], 16)
                     ))
                     offset += 1
-        outfile.write('}\\f0 ')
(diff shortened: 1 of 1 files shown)


akraus53:
Here you can see a direct comparison of the changes:

| before | after |
| -------- | ------- |
| image| image  |

Because of the enormous spacing after every line i had to shrink down the code a lot.

Anteru:
Applied manually in 544e58c5f356fdb7b58b504ae088442f6c81bdcb, thanks.

Issue #2606, opened 2023-12-06
LESS lexer marks `@import` within selector as invalid


stevepiercy:
With the following code sample, where an `@import` statement is placed within a selector:

```less
.loadThemeFonts() {
  @import "./typography.css";
}
```

...the [Pygments LESS lexer](https://pygments.org/demo/?lexer=less&style=bw&formatter=html&code=.loadThemeFonts%28%29+%7B%0A++%40import+%22.%2Ftypography.less%22%3B%0A%7D%0A) marks it as invalid. Whereas in the [Less Sandbox](https://www.lesstester.com/) there is no error.

An `@import` statement on its own is OK in both parsers, although the Less Sandbox warns about the lack of a file manager:

```less
@import "./typography.css";
```

PR #2611, opened 2023-12-12
fix: added @import support to Lesscss

tahamukhtar20:
Fixes: #2606
Fixes: [ plone/volto#5499 ](https://github.com/plone/volto/issues/5499)


jeanas (changes requested):
As it is, there is no way that this rule can get used since it comes just after `(r'@\w+', Name.Variable)` which is also going to match `@import`.

Please fix this, and also add a test (see https://pygments.org/docs/lexerdevelopment/#how-to-test-your-lexer).

Issue #2591, opened 2023-11-20
Shell prompt > not recognized


VanNostrand:
I just noticed that prompts like vagrant@seminar:~> cause the result to be grey like console output when used with LaTeX package minted.
It changes as soon as I replace the > character with #, $ or %.
Could you add the character to the possible prompt tokens please? It is the C-Shell style prompt and a default in SuSE.

PR #2612, opened 2023-12-13
add C-Shell style prompt to BashSession lexer

VanNostrand:
This is related to issue #2591: SuSE uses a C-Shell style prompt like user@host:path>
Previously only prompts ending in $#% symbols were supported.


jeanas:
Thanks, but you need to update the golden test output with `tox -- --update-goldens`. The change looks reasonable to me if this is the prompt in SUSE, but there are tests that explicitly check this isn't done (specifically `tests/snippets/console/fake_ps2_prompt.txt`) so that needs to be updated.

PR #2613, opened 2023-12-19
Add MIME-Type to DesktopLexer

JakobDev:


birkenfeld:
Thank you! Please also run `tox -e mapfiles` to regenerate the mapping created from these attributes, that should take care of the build failure.

jeanas (approved):
Thanks!

PR #2614, opened 2023-12-23
lean: fix the name regex

eric-wieser:
This corrects it to include the full range of unicode characters that are legal.

This also takes the opportunity to add support for highlighting ``` `name ``` and ``` ``checked_name``` literals.

### DIFF
--- a/tests/examplefiles/lean/test.lean
+++ b/tests/examplefiles/lean/test.lean
@@ -211,3 +211,5 @@ end zorn
 -- other bits of tricky syntax
 @[to_additive "See note [foo]"]
 lemma mul_one : sorry := sorry
+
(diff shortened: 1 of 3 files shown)


tests/examplefiles/lean/test.lean
```diff
@@ -211,3 +211,8 @@ end zorn
 -- other bits of tricky syntax
 @[to_additive "See note [foo]"]
 lemma mul_one : sorry := sorry
+
+variables {ι A B : Type*} (𝒜 : ι → A) (ℬ : ι → B)
+
+#check `𝒜.a
```
eric-wieser, line 217: This is mis-highlighted by github, the `.a` is part of the symbol too!

Anteru:
Merged, thanks. My Unicode skills aren't good enough to understand whether the ranges map but I presume you know what you're doing there :)

PR #2618, opened 2023-12-30
Add a lexer for Lean 4

eric-wieser:
This is taken from https://github.com/leanprover/lean4/blob/d92948bc20b12f53542814c79469711ceff19fbf/doc/latex/lean4.py, with subsequent commits showing the changes made on top of this file.

For that reason, I'd prefer if the history could be preserved; though I'd be happy to rewrite it if you want fewer commits / different message conventions.

The test file I'd added is the lean4 translation of the test file that was being used for Lean 3.

cc @Kha


pygments/lexers/lean.py
```diff
@@ -122,5 +123,113 @@ class Lean3Lexer(RegexLexer):
...
+        "(?:(?![λΠΣ])[_a-zA-Zα-ωΑ-Ωϊ-ϻἀ-῾℀-⅏𝒜-𝖟0-9'ⁿ-₉ₐ-ₜᵢ-ᵪ!?])*")
+    _name = _name_segment + r"(\." + _name_segment + r")*"
+
+    keywords1 = (
+        'import', 'abbreviation', 'opaque_hint', 'tactic_hint', 'definition',
+        'renaming', 'inline', 'hiding', 'parameter', 'lemma', 'variable',
+        'theorem', 'axiom', 'inductive', 'structure', 'universe', 'alias',
+        'help', 'options', 'precedence', 'postfix', 'prefix',
+        'infix', 'infixl', 'infixr', 'notation', '#eval',
+        '#check', '#reduce', '#exit', 'coercion', 'end', 'private', 'using', 'namespace',
```
Julian, line 157: I know this list comes from the Lean 4 repo -- but just to note a few that I don't see used anywhere neither in core nor Mathlib:

* `coercion`
* `parameter`
* `help`
* `options`
* `including`
* `context`
* `expose`
* `print` / `reduce` / `synth` -- probably these are meant to be `#print`, `#reduce` / `#synth`?
* `opaque_hint` (looks like it was removed quite awhile ago in https://github.com/leanprover/lean4/commit/e3e2370a3806f30239643c4211ef51341113ba0f)
* `tactic_hint` (a quick search didn't find a specific commit removing it but it doesn't seem to exist anymore, all the commits referencing it are also from 2014)
* `reducible` / `irreducible` -- these only appear as attributes, not keywords -- I didn't double check whether they used to be (keywords) but if they're on this list nonetheless perhaps you should add `semireducible`


and a few that I encountered while double checking the above which *don't* appear here:

* `unif_hint` (appears in https://github.com/leanprover/lean4/blob/master/src/Init/Hints.lean#L11)
* `#check`
* `#check_failure`
* `#reduce`
* `#eval`
* `#synth`
* `#exit`
* `#print`
* `#print axioms`

pygments/lexers/lean.py
```diff
@@ -122,5 +123,113 @@ class Lean3Lexer(RegexLexer):
...
+        # Sorts
+        'Type', 'Prop', 'Sort',
+    )
+
+    operators = (
+        '!=', '#', '&', '&&', '*', '+', '-', '/', '@', '!',
+        '-.', '->', '.', '..', '...', '::', ':>', ';', ';;', '<',
+        '<-', '=', '==', '>', '_', '|', '||', '~', '=>', '<=', '>=',
+        '/\\', '\\/', '∀', 'Π', 'λ', '↔', '∧', '∨', '≠', '≤', '≥',
+        '¬', '⁻¹', '⬝', '▸', '→', '∃', 'ℕ', 'ℤ', '≈', '×', '⌞',
```
Julian, line 181: ℕ and ℤ are here, how about \Q and \R?

pygments/lexers/lean.py
```diff
@@ -122,5 +123,113 @@ class Lean3Lexer(RegexLexer):
...
+            (r'\s+', Whitespace),
+            (r'/--', String.Doc, 'docstring'),
+            (r'/-', Comment, 'comment'),
+            (r'--.*?$', Comment.Single),
+            (words(keywords3, prefix=r'\b', suffix=r'\b'), Keyword.Type),
+            (words(operators), Name.Builtin.Pseudo),
+            (words(punctuation), Operator),
+            (_name_segment, Name),
+            (r'``?' + _name, String.Symbol),
+            (r'\d+', Number.Integer),
```
Julian, line 199:
```suggestion
            (r'\d+', Number.Integer),
            (r'(\d+\.\d*|\d*\.\d+)([eE][+-]?[0-9]+)?j?', Number.Float),
```

Julian (commented):
Nice! This looks pretty good. I'm sure I've missed a thing or two, but I tried to double check the keyword list and give a cursory look over the token list which pretty much looks good!

pygments/lexers/lean.py
```diff
@@ -122,5 +123,113 @@ class Lean3Lexer(RegexLexer):
...
 LeanLexer = Lean3Lexer
+
+
+class Lean4Lexer(RegexLexer):
+    """
+    For the Lean 4 theorem prover.
+    """
+    name = 'Lean4'
+    url = 'https://github.com/leanprover/lean4'
+    aliases = ['lean4']
```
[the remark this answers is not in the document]
eric-wieser: I figured I'd leave that to a follow-up PR. Changing the alias is risky because it breaks all lean3 highlighting in the wild.

cmarqu:
Maybe also add a line to the `CHANGES` file.

pygments/lexers/lean.py
eric-wieser: That looks shameless copied from the Python lexer ;)

```suggestion
            (r'(\d+\.\d*|\d*)([eE][+-]?[0-9]+)?', Number.Float),
```

pygments/lexers/lean.py
eric-wieser: Done, I think. I also added `scoped` and `local`

pygments/lexers/lean.py
Julian: Correct :D -- though I double checked Lean was happy with the scientific notation and it seemed it was. Is the syntax still different?

pygments/lexers/lean.py
eric-wieser: I'd be inclined to just remove `ℕ`, `ℤ`, rather than try to capture all notation in mathlib.

eric-wieser:
```suggestion
        '¬', '⁻¹', '⬝', '▸', '→', '∃', '≈', '×', '⌞',
```

Julian (approved):
Nice. This LGTM at this point I think (with or without the operator tweaking)!

pygments/lexers/lean.py
```diff
@@ -7,12 +7,13 @@
     :copyright: Copyright 2006-2023 by the Pygments team, see AUTHORS.
     :license: BSD, see LICENSE for details.
 """
+import re
```
jeanas, line 10:
```suggestion

import re
```

pygments/lexers/lean.py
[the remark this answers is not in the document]
jeanas: Agreed.

pygments/lexers/lean.py
```diff
@@ -122,5 +124,116 @@ class Lean3Lexer(RegexLexer):
...
+            return 0.1
+
 
 LeanLexer = Lean3Lexer
+
+
+class Lean4Lexer(RegexLexer):
+    """
+    For the Lean 4 theorem prover.
+    """
```
jeanas, line 138:
```suggestion
    """
    
```

pygments/lexers/lean.py
```diff
@@ -122,5 +124,117 @@ class Lean3Lexer(RegexLexer):
...
+
+    punctuation = ('(', ')', ':', '{', '}', '[', ']', '⦃', '⦄',
+                   ':=', ',')
+
+    tokens = {
+        'expression': [
+            (r'\s+', Whitespace),
+            (r'/--', String.Doc, 'docstring'),
+            (r'/-', Comment, 'comment'),
+            (r'--.*?$', Comment.Single),
```
jeanas, line 196:
```suggestion
            (r'--.*$', Comment.Single),
```

pygments/lexers/lean.py
```diff
@@ -122,5 +124,117 @@ class Lean3Lexer(RegexLexer):
...
+            (r'\s+', Whitespace),
+            (r'/--', String.Doc, 'docstring'),
+            (r'/-', Comment, 'comment'),
+            (r'--.*$', Comment.Single),
+            (words(keywords3, prefix=r'\b', suffix=r'\b'), Keyword.Type),
+            (words(operators), Name.Builtin.Pseudo),
+            (words(punctuation), Operator),
+            (_name_segment, Name),
+            (r'``?' + _name, String.Symbol),
+            (r'(?<=\.)\d', Number),
```
jeanas, line 202: Not sure I understand this rule, what is it for? Do you really intend to match just one digit?

pygments/lexers/lean.py
```diff
@@ -122,5 +124,117 @@ class Lean3Lexer(RegexLexer):
...
+            (r'(\d+\.\d*)([eE][+-]?[0-9]+)?', Number.Float),
+            (r'\d+', Number.Integer),
+            (r'"', String.Double, 'string'),
+            (r'[~?][a-z][\w\']*:', Name.Variable),
+            (r'\S', Name.Builtin.Pseudo),
```
jeanas, line 207: What is this rule for?

tests/test_theorem.py
```diff
@@ -0,0 +1,26 @@
+"""
+Tests of the theorem lexers
+"""
```
jeanas, line 3:
```suggestion
"""

```

jeanas:
This looks generally good, only a couple minor questions above.

pygments/lexers/lean.py
eric-wieser: Nice catch, this should match multiple digits. It's matching `foo.123`, which means "take the 123rd attribute of foo`. I would imagine that in the wild this never exceeds one digit, but we may as well be correct here

eric-wieser:
```suggestion
            (r'(?<=\.)\d+', Number),
```

pygments/lexers/lean.py
eric-wieser: Arbitrary notation defined by the user; without it, all such notation would be marked as invalid

jeanas (approved):
Good. Thank you!

eric-wieser:
Thanks! I've made a small follow-up in #2626; it would be great if they could both land in [the coming release](https://github.com/pygments/pygments/milestone/23).

jeanas:
Sure, this will all be in 2.18.

PR #2619, opened 2023-12-31
Logtalk: highlight the uninstantiation_error/1 method and escape sequences in double-quoted terms

pmoura:


pygments/lexers/prolog.py
```diff
@@ -205,9 +205,9 @@ class LogtalkLexer(RegexLexer):
             (r'(>>|<<|/\\|\\\\|\\)', Operator),
             # Predicate aliases
             (r'\bas\b', Operator),
-            # Arithmetic evaluation
+            # Arithemtic evaluation
```
jeanas, line 208: This doesn't look right, unless I'm missing something.

pmoura: Oops! Odd typo. Fixed. Thanks for noticing.

jeanas (approved):

PR #2623, opened 2024-01-11
Change deprecated .textsize into .textlength

mironleon:
See https://pillow.readthedocs.io/en/stable/releasenotes/10.0.0.html#font-size-and-offset-methods

Changed in pillow 10, `textsize` method no longer exists

### DIFF
--- a/pygments/formatters/img.py
+++ b/pygments/formatters/img.py
@@ -633,7 +633,7 @@ class ImageFormatter(Formatter):
                                fill=self.hl_color)
         for pos, value, font, text_fg, text_bg in self.drawables:
             if text_bg:
-                text_size = draw.textsize(text=value, font=font)
+                text_size = draw.textlength(text=value, font=font)
                 draw.rectangle([pos[0], pos[1], pos[0] + text_size[0], pos[1] + text_size[1]], fill=text_bg)
             draw.text(pos, value, font=font, fill=text_fg)
         im.save(outfile, self.image_format.upper())
--- a/pygments/formatters/img.py
+++ b/pygments/formatters/img.py
@@ -7,6 +7,7 @@
     :copyright: Copyright 2006-2024 by the Pygments team, see AUTHORS.
     :license: BSD, see LICENSE for details.
 """
+from packaging import version
 import os
 import sys
 
@@ -19,6 +20,7 @@ import subprocess
 # Import this carefully
 try:
     from PIL import Image, ImageDraw, ImageFont
+    import PIL
     pil_available = True
 except ImportError:
     pil_available = False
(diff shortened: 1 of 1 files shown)


jeanas:
Thanks for the patch! Do you know if there is any functional difference between .textsize() and .textlength(), or is this just a renaming?

jeanas:
Also, [.textlength()](2b8c76254180c8ddc4f76f78761df9e67c881905) seems to have been added in version 9.2.0 which is from July 2022 according to https://pypi.org/project/pillow/#history. Maybe we still want to call .textsize() for older Pillow versions...

mironleon:
> Thanks for the patch! Do you know if there is any functional difference between .textsize() and .textlength(), or is this just a renaming?

Should be exactly the same:
image

mironleon:
> Also, [.textlength()](2b8c76254180c8ddc4f76f78761df9e67c881905) seems to have been added in version 9.2.0 which is from July 2022 according to https://pypi.org/project/pillow/#history. Maybe we still want to call .textsize() for older Pillow versions...

Good catch, I added a version check in fac8b0e9f33099954ebc2ef68ee18f9358205a9f

mironleon:
Let me know if you need anything else from my side!

sagittarian:
I don't think this will work.  `.textsize` returned a pair of (`length`, `height`), while `textlength` returns only the length.  I was able to fix it for myself using `font.getbbox(value)`.

pygments/formatters/img.py
```diff
@@ -633,7 +635,11 @@ def format(self, tokensource, outfile):
                                fill=self.hl_color)
         for pos, value, font, text_fg, text_bg in self.drawables:
             if text_bg:
-                text_size = draw.textsize(text=value, font=font)
+                # see deprecations https://pillow.readthedocs.io/en/stable/releasenotes/9.2.0.html#font-size-and-offset-methods
+                if version.parse(PIL.__version__) < version.Version('9.2.0'):
+                    text_size = draw.textsize(text=value, font=font)
+                else:
+                    text_size = draw.textlength(text=value, font=font)
```
sagittarian, line 642:
```suggestion
                    text_size = font.getbbox(value)[2:]
```

pygments/formatters/img.py
```diff
@@ -7,6 +7,7 @@
     :copyright: Copyright 2006-2024 by the Pygments team, see AUTHORS.
     :license: BSD, see LICENSE for details.
 """
+from packaging import version
```
jeanas, line 10: `packaging` isn't an stdlib module. It's often installed because pip needs it, but I think that, e.g., a vanilla Debian system doesn't have it. I'd just do a `hasattr` check here.

mironleon: f53e3b77cb1bfe931c1aec6cc55eb2888a3dcb78

jeanas:
This looks good now. I've tested it with Pillow 10 and Pillow 9.5.0 (the latter was painful to build...). Thanks!

PR #2624, opened 2024-01-11
Update ThingsDB Lexer with new functions and syntax

joente:
- Added new functions to the lexer.
- Added name constants to the lexer.
- Added a simplified version of template strings to the lexer. _(this can be improved by supporting full template strings, see https://docs.thingsdb.net/v1/overview/template-strings/)_

### DIFF
--- a/pygments/lexers/thingsdb.py
+++ b/pygments/lexers/thingsdb.py
@@ -18,12 +18,12 @@ __all__ = ['ThingsDBLexer']
 class ThingsDBLexer(RegexLexer):
     """
     Lexer for the ThingsDB programming language.
-
-    .. versionadded:: 2.9
     """
     name = 'ThingsDB'
     aliases = ['ti', 'thingsdb']
     filenames = ['*.ti']
+    url = 'https://www.thingsdb.net'
+    version_added = '2.9'
 
     tokens = {
         'root': [
@@ -140,5 +140,5 @@ class ThingsDBLexer(RegexLexer):
             (',', Punctuation),
             (r'\(', Punctuation, '#push'),
             (r'\)', Punctuation, '#pop'),
-        ]
+        ],
     }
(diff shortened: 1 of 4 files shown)


pygments/lexers/thingsdb.py
```diff
@@ -39,21 +39,24 @@ class ThingsDBLexer(RegexLexer):
...
             (r"(?:'(?:[^']*)')+", String.Single),
+            (r"(?:`(?:[^`]*)`)+", String.Backtick),
 
             # literals
             (r'(true|false|nil)\b', Keyword.Constant),
 
+            # name constants
+            (r'(FULL|USER|GRANT|CHANGE|JOIN|RUN|QUERY|'
+             r'DEBUG|INFO|WARNING|ERROR|CRITICAL|'
+             r'NO_IDS|INT_MIN|INT_MAX)\b', Name.Constant),
```
jeanas, line 55:
```suggestion
            (words(('FULL', 'USER', 'GRANT', 'CHANGE', 'JOIN', 'RUN',
                    'QUERY', 'DEBUG', 'INFO', 'WARNING', 'ERROR', 'CRITICAL',
                    'NO_IDS', 'INT_MIN', 'INT_MAX'), suffix=r'\b'),
             Name.Constant),
```

jeanas: Oh well. I was going to change such regexes to `words()` below to, but testing it locally, it appears that `words()` introduces capturing groups (they should probably be non-capturing) which makes it not usable with `bygroups()` later. I'll keep these as big regexes for now, will take a look later at whether we can make `words()`'s introduced groups non-capturing.

pygments/lexers/thingsdb.py
```diff
@@ -39,21 +39,24 @@ class ThingsDBLexer(RegexLexer):
...
+
             # regular expressions
             (r'(/[^/\\]*(?:\\.[^/\\]*)*/i?)', String.Regex),
 
-            # thing id's
-            (r'#[0-9]+', Comment.Preproc),
```
jeanas, line 55: Could you explain why this is removed?

joente: Retrieving things using the hash # syntax was only available in old 0.x versions of ThingsDB and was later removed as part of the language. A simple function call which is more generec, thing(ID), must be used instead.

jeanas:
Thanks!

PR #2625, opened 2024-01-13
feat: Add formatter for wx.RichTextCtrl

TEParsons:
Control to format is supplied as `outFile`, and styles are applied according to the pygments style given. The formatter stores the contents of each control the last time it was formatted to avoid formatting unchanged text twice, making it fast enough to call on each wx.EVT_TEXT event & style the control as text is entered.


Anteru:
I'm not convinced we want a `wx` dependency in Pygments instead of having this in a plugin. It complicates testing (if we have a `wx` extra for example), and I don't like Pygments changing behavior based on the presence of other libraries. Can you explain the rationale while this is useful vs. a plugin? Is there a place where a plugin won't work and it has to be in Pygments upstream?

TEParsons:
My thinking was that importing wx inside methods avoids it having to be a dependency - it's only needed if you're using the wx formatter, which you're not going to do unless you're formatting a wx element (so would have it installed). I see what you mean about complicating testing and making pygments' behaviour variable though, and I don't see any reason it can't be a plugin so I'll do that instead :)

TEParsons:
Okay, moved it over to a plugin: https://github.com/TEParsons/pygments-gui

@Anteru is there any kind of directory I can add this to to help people find it?

Anteru:
Uh, sorry for missing this. Not that I'm aware of really ... maybe we could start adding this to the docs somewhere? `integrate.rst` sounds like a decent place?

PR #2627, opened 2024-01-14
WIP: try making release 'push button' via GitHub actions

jeanas:
(Right now this uses a test project on TestPyPI, we'll see if it works.)

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,0 +1,45 @@
+# When triggered, this workflow publishes a release to PyPI.
+# https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/
+
+name: Publish release to PyPI
+
+on: workflow_dispatch # only runs when manually triggered
+
+jobs:
+  build:
+    name: Build distributions
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: Set up Python
+      - uses: actions/setup-python@v4
+        with:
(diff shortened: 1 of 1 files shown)


jeanas:
The latest iteration seems to have worked. It published https://test.pypi.org/project/Pygments-publishing-test/

@anteru @birkenfeld These are the next steps:

- Decide if we agree that this is the way to go. I assume so, since Matthäus said he feared to mess up something with the release procedure. This eliminates the “`python -m build` + `twine upload`” step in favor of “push button in GitHub actions to upload to PyPI”. (If you want, this can also be triggered upon pushing a Git tag instead of manually.)
- Give me access to the Pygments project on **Test**PyPI (https://test.pypi.org/project/pygments) — this is not strictly required but it would simplify testing.
- I don't have access to Pygments on the actual PyPI (and I'm not asking to change this, plus, we might add other committers in the future). To keep it that way, either of you needs to create an environment 'pypi' with both of you as “required reviewers”. This will ask for approval from at least one of you before publishing to PyPI runs. https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment#creating-an-environment

eric-wieser:
Is this PR indicative that [2.18.0](https://github.com/pygments/pygments/milestone/23) is almost ready to release?

eric-wieser:
Only that I'm looking forward to being able to remove all the hacks we have documented to use Lean 4 in LaTeX; after the next release, the answer will be "it works out of the box with `minted`". Another benefit is that the Zulip chat the Lean 4 community uses will learn the new highlighter.

Of course, the fact I'm excited for it is not a good reason to change your release cadence :).

PR #2631, opened 2024-01-27
Improve the Rust lexer

alanhdu:
This changes the Rust lexer to handle `::`, `:`, and `->` better. Previously, all of these tokens would labeled as text, which leads to poor rendering on certain styles. For instance, using a `monokai` HTML style w/ inline CSS, I get

image

where the tokens are rendered using the default (black) text which is not really visible against the background.

This changes the lexer so that:
* `:` and `->` are considered punctuation
* `::` is considered an operator -- I am not entirely sure that's the correct thing to do, but it seems like a resaonable choice
* I went through and used `Text.Whitespace` instead of plain `Text` a little more thoroughly.

### DIFF
--- a/pygments/lexers/rust.py
+++ b/pygments/lexers/rust.py
@@ -105,7 +105,7 @@ class RustLexer(RegexLexer):
             builtin_funcs_types,
             builtin_macros,
             # Path separators, so types don't catch them.
-            (r'::\b', Operator),
+            (r'::\b', Punctuation),
             # Types in positions.
             (r'(?::|->)', Punctuation, 'typename'),
             # Labels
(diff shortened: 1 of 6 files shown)


pygments/lexers/rust.py
```diff
@@ -98,16 +98,16 @@ class RustLexer(RegexLexer):
...
-            (r'(default)(\s+)(type|fn)\b', bygroups(Keyword, Text, Keyword)),
+            (r'(default)(\s+)(type|fn)\b', bygroups(Keyword, Whitespace, Keyword)),
             keyword_types,
             (r'[sS]elf\b', Name.Builtin.Pseudo),
             # Prelude (taken from Rust's src/libstd/prelude.rs)
             builtin_funcs_types,
             builtin_macros,
             # Path separators, so types don't catch them.
-            (r'::\b', Text),
+            (r'::\b', Operator),
```
birkenfeld, line 108: I'd prefer Punctuation here too. Operators are typically more emphasized than is good for a name separator.

alanhdu: Done!

jeanas (approved):

jeanas:
Thanks!

PR #2638, opened 2024-02-03
Fix TexLexer (#2637)

### LINKED ISSUE #2637: [Bug] TexLexer uses inaccurate definition for command tokens
The `TexLexer` class from [`pygments/lexers/markup.py`](https://github.com/pygments/pygments/blob/master/pygments/lexers/markup.py) posits that (La)TeX commands are of the format `\\([a-zA-Z]+|.)`:

https://github.com/pygments/pygments/blob/e2f206a35bee1f8642f3c4948f1a436b60d5abd2/pygments/lexers/markup.py#L300

This, however, is inaccurate. For example, in [`expl3`](https://ctan.org/pkg/expl3) packages, one will often find lines like this:

```latex
\new_dim:n \l_@@_example_dim
```

For commands such as these, the colon, underscore, and @ characters should all be considered valid parts of the command names.

However, inside math environments, these characters should not be considered valid parts of command names, since, for example, `\alpha_i` is not a single command, but rather "`\alpha` with subscript `i`" (in LaTeX, equivalent to `\alpha\sb{i}`).

Of course, these claims disregard some exceptional cases (for example, an `expl3`-style command being used inside math mode), but I believe accurately describes the expected syntax highlighting for the vast majority of situations.

I'll make a PR to change the definition in `TexLexer`, but I'm not familiar with the inner workings of the Pygments library, so I'm not sure if the changes will have the intended effects (in particular, I don't know if the changes in the definition of the command tokens will affect the syntax highlighting inside math mode as well). For that, and any other part of this issue, your input would be much appreciated. Thank you!

nvlang:
Closes #2637.


jeanas:
Test showed that this was causing the Wikitext lexer to lex `<math>\log_k(x)</math>` as a command `\log_k`. I've pushed a fix that makes it use the TeX lexer in math mode instead.

Looks good, could you add a test though? (I can try but I'm not TeX wizard.)

nvlang:
> Test showed that this was causing the Wikitext lexer to lex `<math>\log_k(x)</math>` as a command `\log_k`. I've pushed a fix that makes it use the TeX lexer in math mode instead.
> 
> Looks good, could you add a test though? (I can try but I'm no TeX wizard.)

Thank you! Regarding adding a test: I'd love to, but I'm afraid I'm not familiar with how to write tests for the Pygments library (or `pytest` in general). I can describe the expected behavior, though, and give some examples, if that'd help (see also #2637, the issue this PR aims to fix):

**TL;DR**: A string is a LaTeX command if and only if it matches the regex [`(\\[A-Za-z_@:]+|\\\S?)`](https://regex101.com/r/2AkP6x/1).

In LaTeX, `@` (at sign), `:` (colon), and `_` (underscore) can be used in the names of commands under some circumstances (usually outside of the document body, i.e., the content inside `\begin{document}...\end{document}`, since commands containing these special characters are often advanced, lower-level commands.

Strictly speaking, to make `@` a valid part of a command name, the command definition has to be wrapped by `\makeatletter...\makeatother` (for `:` and `_`, the same is true but with `\ExplSyntaxOn...\ExplSyntaxOff` instead) — however, this should not be relevant to the syntax highlighting in my opinion, for the reasons mentioned in #2637.

Regardless, this basically just means that, for example, each of the following should be a valid command name as far as the lexer is concerned (and, as a matter of fact, these command names are taken from official documentation):

```latex
\toks@
\sixt@@n
\::N
\:::
\clist_set:Nn
\sort_return_same:
\@@_case:NnTF
```

jeanas:
> Thank you! Regarding adding a test: I'd love to, but I'm afraid I'm not familiar with how to write tests for the Pygments library (or pytest in general).

See https://pygments.org/docs/lexerdevelopment/#how-to-test-your-lexer

In short, add a file with test TeX code, either `tests/snippets/tex/testname.txt` or `tests/examplefiles/tex/testname.tex`, and run `tox -- --update-goldens`.

nvlang:
@jeanas Thank you for the hint, I missed that :) I added some basic tests now (I was a bit surprised that, as far as I could tell, there weren't already tests for TexLexer).

Also, I made another small change in the TexLexer (see commit [d40643e](https://github.com/pygments/pygments/pull/2638/commits/d40643e27732354f0699e2460e26e760516054a5)). I replaced `.` with `\S?`, for the following reason: previously, the strings `\ `, `\\t`, and `\\n` were each parsed as a single command, instead of `\` plus the whitespace. In LaTeX, the `\` command, on its own, just inserts a space; it just so happens that this command can only be invoked if it is followed by whitespace, since otherwise we're bound to invoke a different command, but the whitespace itself is not part of the command, at least as far as I know.

jeanas:
Sorry for the delay. Thank you for the tests, looks good!

PR #2642, opened 2024-02-09
Add OneDarkVivd style

tristan957:
This is inspired by the VSCode One Dark Pro vivid variant, which was originally inspired by the One Dark theme for Atom.

### DIFF
--- a/pygments/styles/onedark_vivid.py
+++ b/pygments/styles/onedark_vivid.py
@@ -1,6 +1,6 @@
 """
     pygments.styles.onedark_vivid
-    ~~~~~~~~~~~~~~~~~~~~~~~
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
     One Dark Vivid Theme for Pygments by Tristan Partin (https://github.com/tristan957).
 
--- a/pygments/styles/onedark_vivid.py
+++ b/pygments/styles/onedark_vivid.py
@@ -1,0 +1,63 @@
+"""
+    pygments.styles.onedark_vivid
+    ~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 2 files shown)


tristan957:
Thanks for maintaining pygments!

pygments/styles/onedark_vivid.py
```diff
@@ -0,0 +1,63 @@
+"""
+    pygments.styles.onedark_vivid
+    ~~~~~~~~~~~~~~~~~~~~~~~
```
jeanas, line 3:
```suggestion
    pygments.styles.onedark_vivid
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

tristan957:
Would you like me to run that tox command to fix the CI or will you? Don't want to step on your toes.

tristan957:
Thank you!

tristan957:
Hmm. What do you think I should do about the test failure? I'm also happy to just not upstream this if it doesn't meet the requirements you want upstream.

jeanas:
For styles which are clones of existing editor themes, we cannot always improve accessibility while changing the look subtly enough that the "brand" is not modified, so this is not a blocker (our page https://pygments.org/styles/ clearly demarcates styles with a lower contrast so you can make an informed choice). On the other hand, what I'm wondering about is whether we want a style this close to another existing style. I opened the demo and I was initially confused as to whether it was really different from `one-dark`, although I can see a slight difference when zooming.

image

image

tristan957:
I'll leave it up to you

Anteru:
To be quite honest, this is extremely close, and even on a calibrated screen the differences a minor. I'd say if `vivid` hits the accessibility contrast rules, that would make it a good candidate (and clearly it's a derivative already so we're free to diverge a bit more), otherwise I don't see the value add.

PR #2648, opened 2024-02-17
add: Material for MkDocs

zeegin:
https://pygments.org/faq/
Who uses Pygments?

https://squidfunk.github.io/mkdocs-material/setup/extensions/python-markdown-extensions/#highlight


Anteru:
Added, thanks!

PR #2651, opened 2024-02-20
[elpi] matching parens of same color + nested quotations

FissoreD:
Original (matching parens of different color)
image

Fixed
image
---
Original (nested quotations):
image

Fixed:
image

CC @gares

### DIFF
--- a/pygments/lexers/elpi.py
+++ b/pygments/lexers/elpi.py
@@ -8,12 +8,13 @@
     :license: BSD, see LICENSE for details.
 """
 
-from pygments.lexer import RegexLexer, bygroups, include
+from pygments.lexer import RegexLexer, bygroups, include, using
 from pygments.token import Text, Comment, Operator, Keyword, Name, String, \
     Number, Punctuation
 
 __all__ = ['ElpiLexer']
 
+from pygments.lexers.theorem import CoqLexer
 
 class ElpiLexer(RegexLexer):
     """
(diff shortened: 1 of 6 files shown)


Anteru:
Good catch. Would you mind adding an example for this? Previously apparently the first parentheses was matched as whitespace.

gares:
Please run CI

pygments/lexers/elpi.py
```diff
@@ -152,21 +152,24 @@ class ElpiLexer(RegexLexer):
             (r'"', String.Double, '#pop'),
         ],
         'elpi-quote': [
-            (r'\{\{', Punctuation, '#push'),
             (r'\}\}', Punctuation, '#pop'),
-            (r"(lp:)((?=[A-Z_]){})".format(constant_re), bygroups(Keyword, Name.Variable)),
-            (r"[^l\}]+", Text),
-            (r"l|\}", Text),
+            (r"\s", Text.Whitespace),
```
Anteru, line 156: This should be `\s+`.

pygments/lexers/elpi.py
```diff
@@ -152,21 +152,24 @@ class ElpiLexer(RegexLexer):
...
+            (r"(lp:)({})".format("{{"), bygroups(Number, Punctuation), 'elpi-quote-exit'),
+            (r"(lp:)((?=[A-Z_]){})".format(constant_re), bygroups(Number, Name.Variable)),
+            (r".", Text),
```
Anteru, line 159: Also `.+`, otherwise you get one match per character (see your output - you want multiple consecutive characters to match?

Anteru (changes requested):

pygments/lexers/elpi.py
FissoreD: this is a good idea! Nonetheless, .+ is overly aggressive, as I don't intend to capture all characters up to the next \n. 
I can opt for a more nuanced approach, drawing inspiration from .+ but employing negative lookahead.

gares:
Please run ci

gares (approved):

gares:
Looks good to me

gares:
Thanks!

PR #2655, opened 2024-02-27
Update Macaulay2 symbols for version 1.23

d-torrance:
We add several new symbols to the Macaulay2 lexer that were introduced in version 1.23.


Anteru:
Merged, thanks!

PR #2658, opened 2024-03-03
Add support for HolyC

xslendix:


Anteru:
This needs an example file (with example output.)

PR #2659, opened 2024-03-07
Add lexer for Soong (Android.bp) config files

JonathonReinhart:
This adds a lexer for `Android.bp` files which are used by the *Soong* Android build system.

https://source.android.com/docs/setup/build

Here is an example of a full `Android.bp` file: [`platform/frameworks/base/Android.bp`](https://android.googlesource.com/platform/frameworks/base/+/55afce9107ea882fe236acc5653aef9d6fb2185b/Android.bp).

### DIFF
--- a/pygments/lexers/soong.py
+++ b/pygments/lexers/soong.py
@@ -1,0 +1,79 @@
+"""
+    pygments.lexers.soong
+    ~~~~~~~~~~~~~~~~~~~~~
+
+    Lexers for Soong (Android.bp Blueprint) files.
+
+    :copyright: Copyright 2006-2024 by the Pygments team, see AUTHORS.
+    :license: BSD, see LICENSE for details.
+"""
+from pygments.lexer import RegexLexer, bygroups, include
+from pygments.token import Comment, Name, Number, Operator, Punctuation, \
+        String, Whitespace
+
+__all__ = ['SoongLexer']
+
+class SoongLexer(RegexLexer):
+    name = 'Soong'
+    version_added = '2.18'
+    url = 'https://source.android.com/docs/setup/reference/androidbp'
+    aliases = ['androidbp', 'bp', 'soong']
+    filenames = [
+        'Android.bp',
+    ]
+
+    tokens = {
+        'root': [
+            # A variable assignment
+            (r'(\w*)(\s*)(\+?=)(\s*)',
+             bygroups(Name.Variable, Whitespace, Operator, Whitespace),
+             'assign-rhs'),
+
+            # A top-level module
+            (r'(\w*)(\s*)(\{)',
+             bygroups(Name.Function, Whitespace, Punctuation),
+             'in-rule'),
+
+            # Everything else
+            include('comments'),
+            (r'\s+', Whitespace),  # newlines okay
+        ],
+        'assign-rhs': [
+            include('expr'),
+            (r'\n', Whitespace, '#pop'),
+        ],
+        'in-list': [
+            include('expr'),
+            include('comments'),
+            (r'\s+', Whitespace),  # newlines okay in a list
+            (r',', Punctuation),
+            (r'\]', Punctuation, '#pop'),
+        ],
+        'in-map': [
+            # A map key
+            (r'(\w+)(:)(\s*)', bygroups(Name, Punctuation, Whitespace)),
+
+            include('expr'),
+            include('comments'),
+            (r'\s+', Whitespace),  # newlines okay in a map
+            (r',', Punctuation),
+            (r'\}', Punctuation, '#pop'),
+        ],
+        'in-rule': [
+            # Just re-use map syntax
+            include('in-map'),
+        ],
+        'comments': [
+            (r'//.*', Comment.Single),
+            (r'/(\\\n)?[*](.|\n)*?[*](\\\n)?/', Comment.Multiline),
+        ],
+        'expr': [
+            (r'(true|false)\b', Name.Builtin),
+            (r'0x[0-9a-fA-F]+', Number.Hex),
+            (r'\d+', Number.Integer),
+            (r'".*?"', String),
+            (r'\{', Punctuation, 'in-map'),
+            (r'\[', Punctuation, 'in-list'),
+            (r'\w+', Name),
+        ],
+    }
--- a/AUTHORS
+++ b/AUTHORS
@@ -196,6 +196,7 @@ Other contributors, listed alphabetically, are:
 * Kashif Rasul -- CUDA lexer
 * Nathan Reed -- HLSL lexer
 * Justin Reidy -- MXML lexer
+* Jonathon Reinhart, Google LLC -- Soong lexer
 * Norman Richards -- JSON lexer
 * Corey Richardson -- Rust lexer updates
 * Fabrizio Riguzzi -- cplint leder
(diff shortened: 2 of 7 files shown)


JonathonReinhart:
Hi @Anteru @jeanas, friendly ping! Is there anything I can do to advance this?

Anteru:
Just came back from travel, looking into this now.

JonathonReinhart:
Thanks @Anteru. I fixed up the `check` and `regexlint` errors.

pygments/lexers/soong.py
```diff
@@ -0,0 +1,70 @@
...
+            (r'\}', Punctuation, '#pop'),
+        ],
+        'in-rule': [
+            # Just re-use map syntax
+            include('in-map'),
+        ],
+        'common': [
+            (r'//.*?$', Comment.Single),
+            (r'/(\\\n)?[*](.|\n)*?[*](\\\n)?/', Comment.Multiline),
+            (r'\n', Whitespace),
```
Anteru, line 56: `\n+`? Should be no problem merging consecutive whitespace characters.

pygments/lexers/soong.py
```diff
@@ -0,0 +1,70 @@
...
+            (r' +', Whitespace),
```
Anteru, line 57: Does this language not support `\t` as a whitespace character?

pygments/lexers/soong.py
```diff
@@ -0,0 +1,70 @@
...
+        ],
+        'expr': [
+            (r'(true|false)\b', Name.Builtin),
+            (r'0x[0-9a-fA-F]+', Number.Hex),
+            (r'\d+', Number.Integer),
+            (r'".*?"', String),
+            (r'[\[\]]', Punctuation),
+            (r',', Punctuation),
```
Anteru, line 65: You can merge those two rules and use `+` to allow consuming multiple punctuation characters in one go.

Anteru (changes requested):

pygments/lexers/soong.py
JonathonReinhart: Replaced with `\s+`.

pygments/lexers/soong.py
JonathonReinhart: Replaced with `\s+`.

pygments/lexers/soong.py
JonathonReinhart: Obsolete. Square brackets actually move to a new `in-list` state.

JonathonReinhart:
Thanks for the review, @Anteru! Your comment about whitespace helped me realize there was a deficiency in my previous version (multi-line lists weren't handled correctly). I've corrected this and added a test.

Please take another look!

pygments/lexers/soong.py
```diff
@@ -0,0 +1,77 @@
...
+        'expr': [
+            (r'(true|false)\b', Name.Builtin),
+            (r'0x[0-9a-fA-F]+', Number.Hex),
+            (r'\d+', Number.Integer),
+            (r'".*?"', String),
+            (r',', Punctuation),
+            (r'\{', Punctuation, 'in-map'),
+            (r'\[', Punctuation, 'in-list'),
+            (r'\w+', Name),
+            include('comments'),
```
jeanas, line 75: Please add a rule

```python
default('#pop'),
```

here. IIUC, it's working right now because we have an undocumented fallback to restarting in the `'root'` state when nothing matches and there's a newline boundary, but we might want to remove that, so better not rely on it...

pygments/lexers/soong.py
```diff
@@ -0,0 +1,77 @@
...
+            (r'\s+', Whitespace),  # newlines okay in a map
+            (r',', Punctuation),
+            (r'\}', Punctuation, '#pop'),
+        ],
+        'in-rule': [
+            # Just re-use map syntax
+            include('in-map'),
+        ],
+        'comments': [
+            (r'//.*?$', Comment.Single),
```
jeanas, line 63: Unnecessary:

```suggestion
            (r'//.*?', Comment.Single),
```

pygments/lexers/soong.py
```diff
@@ -0,0 +1,77 @@
...
+            (r'\]', Punctuation, '#pop'),
+        ],
+        'in-map': [
+            # A map key
+            (r'(\w+)(:)(\s*)', bygroups(Name, Punctuation, Whitespace)),
+
+            include('expr'),
+            include('comments'),
+            (r'\s+', Whitespace),  # newlines okay in a map
+            (r',', Punctuation),
```
jeanas, line 55: You already have this in `'expr'`.

pygments/lexers/soong.py
```diff
@@ -0,0 +1,77 @@
...
+
+            # Everything else
+            include('comments'),
+            (r'\s+', Whitespace),  # newlines okay
+        ],
+        'in-list': [
+            include('expr'),
+            include('comments'),
+            (r'\s+', Whitespace),  # newlines okay in a list
+            (r',', Punctuation),
```
jeanas, line 45: Already in `'expr'`

jeanas (changes requested):

pygments/lexers/soong.py
JonathonReinhart: Unfortunately, I don't think it's that simple. I tried that and it broke everything.

Currently, `expr` is used with both an `include()` and a new state. (That seems like a bad idea, now that I think about it.)

So while it makes sense to pop out of the `expr` state, I don't want to leave the other states that `include('expr')`.

I'll try separating the actual state from the included group.

pygments/lexers/soong.py
JonathonReinhart: Actually, I think it probably makes sense to keep this one and remove the `,` from `expr`. (I think each item in a comma-separated list is an expression; the commas are between each expression).

I removed both `,` and 'comments' from `expr` with no breakage.

pygments/lexers/soong.py
JonathonReinhart: See [other comment](https://github.com/pygments/pygments/pull/2659/files/67edcd927cb0d61d5d54d874e9074ce085799508#r1544492882).

pygments/lexers/soong.py
JonathonReinhart: Removing `$` seems to break everything.

I think the reason is because `.*?` is non-greedy. If you take away the subsequent character, it will match nothing:

https://docs.python.org/3/howto/regex.html#greedy-versus-non-greedy

```Python Console
>>> re.match(r'//.*?', '// foo!')
<re.Match object; span=(0, 2), match='//'>
>>> 
>>> re.match(r'//.*?!', '// foo!')
<re.Match object; span=(0, 7), match='// foo!'>
```

---

I took this from one of the other lexers. If you search, there are others exactly like this:
```console
$ git grep "\/\/\.\*\?\$', Comment"
```

pygments/lexers/soong.py
JonathonReinhart: That seemed to do the trick. See b3b1e66. (We can squash it if you're happy with it.)

This introduces a new state called `'assign-rhs'` which includes expressions, and is explicitly terminated on a newline. (I wish there was something like `default(blow up)` to assert this.)

pygments/lexers/soong.py
JonathonReinhart: Oh in fact, this comes directly from [the docs](https://pygments.org/docs/lexerdevelopment/#changing-states).

jeanas: Oh, sorry, of course I meant to also remove the `?` at the end (and I've just made that change in the docs).

jeanas: Change applied.

JonathonReinhart: Got it. I made the change with no impact to the output.

pygments/lexers/soong.py
JonathonReinhart: Squashed.

JonathonReinhart:
Thanks @jeanas. All comments have been resolved.

jeanas (approved):
Looks good. Thanks!

PR #2661, opened 2024-03-09
add pyupgrade lint group

danieleades:


Anteru:
@jeanas , @birkenfeld : I'm generally in favor of being consistent in application of `f` strings instead of using `.format()` all over the place, but I'm not 100% convinced if we want this kind of sweeping change. Things I dislike here are:

* https://github.com/pygments/pygments/pull/2661/files#diff-19b41a2482f38f35c49309b6d25775ed9c2b059c2e5a4139d7c10a90bef1d9c9R601
* https://github.com/pygments/pygments/pull/2661/files#diff-972e36f207f7e653c50bdc7dd2bb0469f93eb3aa64558e4341a446fa147660d1R121

But overall, I think this has merit. Any strong feelings about this? Otherwise I'll merge this for 2.18.

birkenfeld:
Yeah the long lines should be broken, but otherwise, I think it's okay.

danieleades:
> @jeanas , @birkenfeld : I'm generally in favor of being consistent in application of `f` strings instead of using `.format()` all over the place, but I'm not 100% convinced if we want this kind of sweeping change. Things I dislike here are:
> 
> * https://github.com/pygments/pygments/pull/2661/files#diff-19b41a2482f38f35c49309b6d25775ed9c2b059c2e5a4139d7c10a90bef1d9c9R601

yep, will break this up

> * https://github.com/pygments/pygments/pull/2661/files#diff-972e36f207f7e653c50bdc7dd2bb0469f93eb3aa64558e4341a446fa147660d1R121

what's the issue with this one?

> But overall, I think this has merit. Any strong feelings about this? Otherwise I'll merge this for 2.18.

Anteru:
The issue there is that I find `groupname` to be a better name than `_re_id`, and I don't see a way to get to the same level of readability with `f`- strings, but it's not a strong argument.

Anteru:
Merged, thanks!

PR #2672, opened 2024-03-27
Remove an unnecessary apostrophe in this text

alexwlchan:
This is possessive "its" which doesn't need an apostrophe; "it's" is a contraction of "it is" which doesn't make sense here.


Anteru:
Merged, well spotted!

PR #2676, opened 2024-03-29
add ruff bugbear lint group

danieleades:


birkenfeld:
A lot of these look good, with a few things that I noted on cursory review:

* In many cases, `from None` is better than `from e`
* `assert False` is perfectly fine in tests

PR #2677, opened 2024-03-29
add ruff-specific lints

danieleades:


Anteru:
How was this PR created? Which ruff rules were used for this? I just ran ruff to fix an (unrelated) set of problems, and I'm wondering if we're chasing ruff versions here?

danieleades:
> How was this PR created? Which ruff rules were used for this? I just ran ruff to fix an (unrelated) set of problems, and I'm wondering if we're chasing ruff versions here?

this PR adds the `RUF` lint group, as you can see here - https://github.com/pygments/pygments/pull/2677/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711

danieleades:
closing. this should be revisited from scratch

PR #2681, opened 2024-04-08
Support multiline string in Swift

sebastianv1:
Adds support for multiline strings and refactors a bit of the shared lexing rules with single-line string.


sebastianv1:
Hi @Anteru, curious if there's a timeframe we might be able to expect this PR to reviewed/merged? We might be interested in bumping pygments to a pre-release version if it's merged prior to 2.18.0 release. 🙏

Anteru:
We're already past due the planned release time. I hope to get to it this month, but stuff is coming up :(

sebastianv1:
Sounds good, thank you for taking the time! Whenever this is merged I think we'll just look to point pygments at `HEAD` and not wait for the official release :)

PR #2682, opened 2024-04-08
Add "visual-basic" as alias of vb.net

verhovsky:


Anteru:
You need to recreate the mapfiles first.

verhovsky:
It would've been faster to close this PR and do it yourself instead of telling me figure out what that means.

PR #2697, opened 2024-04-27
Add alias for pycon lexer (python-console)

bmispelon:
Hi,

While writing documentation for the Django project (which uses reStructuredText) I noticed a `code-block:: pycon` and was really confused for a while. I genuinely assumed someone had made a typo because when I read "pycon", the only thing that comes to mind for me is the [python conference of the same name](https://pycon.org/). I had to look it up to figure out it was short for "PYthon CONsole".

This PR adds a new alias for the lexer with a name that's a bit more verbose, but also a bit less confusing (imo).

As it's a new alias, I believe this should be fully backwards compatible.

Thanks! ✨


birkenfeld:
Thanks!

bmispelon:
That was fast, thanks for the quick merge! 🚀

PR #2698, opened 2024-04-27
Added XonshLexer

anki-code:
Added [Xonsh shell](https://github.com/xonsh/xonsh) lexer.

Example:

```python
from pygments.lexers.python import XonshLexer
from pygments.formatters import HtmlFormatter

# Command from https://xon.sh/tutorial.html#nesting-subprocesses
code = "$[@$(which @($(echo ls).strip())) @('-' + $(printf 'l'))]"
print(highlight(code, XonshLexer(), HtmlFormatter()))
```

image

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

### DIFF
--- a/tests/snippets/xonsh/test_xonsh.txt
+++ b/tests/snippets/xonsh/test_xonsh.txt
@@ -1,0 +1,46 @@
+---input---
+$PATH.append(p"/home/anki-code/" / ".local/bin")
+$(echo @("Path updated"))
+bash -c! echo 1
+
+---tokens---
+'$PATH'       Name.Variable
+'.'           Operator
+'append'      Name
+'('           Punctuation
+'p'           Name
+'"'           Literal.String.Double
+'/home/anki-code/' Literal.String.Double
+'"'           Literal.String.Double
+' '           Text
+'/'           Operator
+' '           Text
+'"'           Literal.String.Double
+'.local/bin'  Literal.String.Double
+'"'           Literal.String.Double
+')'           Punctuation
+'\n'          Text.Whitespace
+
+'$'           Keyword
+'('           Punctuation
+'echo'        Name.Builtin
+' '           Text
+'@'           Keyword
+'('           Punctuation
+'"'           Literal.String.Double
+'Path updated' Literal.String.Double
+'"'           Literal.String.Double
+')'           Punctuation
+')'           Punctuation
+'\n'          Text.Whitespace
+
+'bash'        Name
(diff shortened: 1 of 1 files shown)


Anteru:
This needs an example file before we can consider merging it. ``version_added`` should be set to 2.19, we won't merge this for 2.18 any more.

pygments/lexers/python.py
```diff
@@ -1196,3 +1196,85 @@ def analyse_text(text):
...
+            include("mode_switch_brackets"),
+            (r"&&|\|\||\band\b|\bor\b", Operator, "subproc_start"),
+            (r'"(\\\\|\\[0-7]+|\\.|[^"\\])*"', String.Double),
+            (r"'(\\\\|\\[0-7]+|\\.|[^'\\])*'", String.Single),
+            (r"(?<=\w|\s)!", Keyword, "subproc_macro"),
+            (r"^!", Keyword, "subproc_macro"),
+            (r";", Punctuation, "subproc_start"),
+            (r"&|=", Punctuation),
+            (r"\|", Punctuation, "subproc_start"),
+            (r"\s+", Text),
```
Anteru, line 1271: Shouldn't this be whitespace? It's not a quoted string after all.

pygments/lexers/python.py
```diff
@@ -1196,3 +1196,85 @@ def analyse_text(text):
...
+            (r"\$\w+", Name.Variable),
+            (r"\(", Punctuation, "py_bracket"),
+            (r"\{", Punctuation, "py_curly_bracket"),
+            include("mode_switch_brackets"),
+            inherit,
+        ],
+        "subproc_start": [
+            (r"\s+", Whitespace),
+            (COMMAND_TOKEN_RE, Name.Builtin, "#pop"),
+            (r"", Whitespace, "#pop"),
```
Anteru, line 1259: Pardon me, but I don't see this used anywhere else in Pygments, nor do I fully understand why you'd want to match nothing to pop? Can you elaborate?

Anteru (changes requested):

pygments/lexers/python.py
jeanas: Comment from the sidelines: Use the `default("#pop")` idiom?

Anteru: Yes. I just realized it is on the "writing your lexer" page ... hmm.

PR #2707, opened 2024-05-14
Add new symbols for Macaulay2 1.24.05

d-torrance:
We add several new symbols from the latest Macaulay2 release to the lexer.


Anteru:
As usual, thanks!

PR #2712, opened 2024-05-18
Fix Sphinx build and check example file existence

kdeldycke:
There's an issue that was introduced in #2442 that breaks the Sphinx build. This is a hotfix.


kdeldycke:
Sorry @Anteru, but an issue was introduced with the #2442 PR you just merged. Which breaks the building of the Sphinx documentation. This PR fixes this.

And also, to prevent the problem to be triggered again, I added a safeguard in the unit tests.

Anteru:
Good catch, thanks!

PR #2713, opened 2024-05-19
Aligns example files and lexer's filename pattern matching

kdeldycke:
This PR update all example files so they match their lexer's file extension.

It also fixes inconsistencies between example files' folders and their matching lexer's alias.

To prevent any further inconsistencies to be introduced, a couple of tests are added in the test suite to enforce this alignments.

This is a follow up on #2442 and #2712.


Anteru:
Thanks for all the work, that helps quite a bit with ad-hoc testing if XML files for example end with `.xml` :)

kdeldycke:
Thanks for the quick merge!

PR #2714, opened 2024-05-19
Test lexer metadata uniqueness

kdeldycke:
To prevent future contributors to introduce inconsistencies, I propose to add a unit test which enforce uniqueness of some lexer's metadata:
- name
- aliases
- mimetypes


Anteru:
That's a good idea, and I guess we could extend this in the future to make sure the version is not out of whack, etc. Thanks!

kdeldycke:
> That's a good idea, and I guess we could extend this in the future to make sure the version is not out of whack, etc. Thanks!

Yes, that's the idea. But for this PR I played it safe and only validated the metadata I was familiar with from my last weeks experiments! :)

PR #2715, opened 2024-05-20
Test example files and snippets conventions

kdeldycke:
Another followup on #2713, this PR adds test to check the naming convention of folders in which test snippets resides.

Also use proper Pytest's `marks` to skip example files not strictly adhering to these conventions.


Anteru:
You're on a roll!

Issue #986, opened 2019-08-31
Incorrect highlighting for C# hexadecimal integer literals


Anteru:
(Original issue 1282 created by lesderid on 2016-09-03T15:32:38.546743+00:00)

C# hexadecimal integer literals don't highlight properly. This is probably a bug in the lexer.

Here's an example:  
[![xkuzmv.png](https://pst.moe/paste/xkuzmv?format=png)](https://pst.moe/paste/xkuzmv)

PR #2727, opened 2024-06-09
Fix for #986

adluu2k1:
Fix for https://github.com/pygments/pygments/issues/986

Before:
image

After:
image

I also created rules for numerical literals based on [C# Lexical Structure](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/lexical-structure#6453-integer-literals) in microsoft docs.
Appreciate any comments and feedbacks!


Anteru:
Merged, thanks!

Issue #2671, opened 2024-03-26
No recognition of JS/TS private identifiers


isc-aray:
Pygments does not recognize JS/TS private identifiers (which start with a pound sign, e.g. `#page`).

Minimal example:
``` ts
class test {
    #testprop: string
}
```

Expect no highlighting for bad syntax, but the `#` symbol is highlighted as a syntax error.

PR #2729, opened 2024-06-12
Issue 2671 fixed. The private identifiers can be recognized correctly.

QSsoul:
Hello, here is the change of the error of "#". We add a new rule for private identifier and make it can be recognized. 

328522807-0347592c-f210-4890-8ea5-2c4180ee6a51

We also pass the CI test.


ZuxingGit:
link ->  #2671

Anteru:
Merged, thanks!

Issue #1880, opened 2021-08-07
Please add a JSON5 lexer


purple512:
[JSON5 specification](https://spec.json5.org/) - a superset of JSON

It would be nice to be able to format JSON5 files.

ct456:
Hi, I'm a student from the University of Adelaide, I'm in a group of students working on contributing to this github repository. We'd like to take on this issue if its still available.

Jean-Abou-Samra:
Feel free. Since JSON5 looks like a not-too-big extension to JSON, I'd recommend extending the existing JSON lexer. Out of curiosity, who else is there in your student group? Is your university asking you to do this?

ct456:
Hi, thank you for the prompt reply and tips. I'll be the one primarily working on the code, there are other members who will be working on other non-git related work. And yes this is part of an assessment piece for a subject.

BuildMonkey:
Hello @Jean-Abou-Samra,

I am the lecturer for the course. This is a student team working on this issue. The team has been notified that they will need to collaborate on code contributions.

We are from the University of Adelaide, where students have an eight-week project contributing to an OSS project. Having briefly reading this issue, it seems like this might not be a good issue for a team to work on over weeks. 

Perhaps there is another issue that might be better suited for a team in the time period provided.

Regards,
Dr. Rita Garcia

Jean-Abou-Samra:
The bulk of Pygments is lexers for various languages. I think working on lexers is not a bad idea, as it will make for nicely self-contained projects. On the other hand, if it's a team of five and a project of eight weeks, just one such issue might not be enough. Luckily, there are dozens of them in the tracker ...

I think the project this team could tackle really depends on who they are. @ct456 What is your academic level? Did any of you contribute to open source software before? Is it your first year with programming or are you more advanced? Do you study computer science as a main subject or as part of a curriculum in another field? Do you have some experience of version control (whether Git, Bazaar, SVN, ...)?

Besides improving all the lexers, here are a few items I can think of, which are not necessarily captured in tracker issues:

- Improving tests. Under `scripts/`, you can find a few tests that have not yet been made required, and would ideally be, fixing existing failures as much as possible or (if not feasible all at once) adding lists of known failures. This can seem less attractive because it doesn't directly impact users, but for development tests are paramount.
- Adding new styles or improving existing formatters (the HTML one for example has some feature requests in the tracker).
[...]

PR #2734, opened 2024-06-18
Add JSON5 lexer

n00dlr:
Implemented a JSON5 lexer, as requested in #1880.

This lexer was mostly derived from the jsonnet lexer, which already implements lexing for features such as unquoted and single quoted object names and string values, comments and trailing commas in objects and arrays. This lexer covers the other JSON5 features such as mulitline strings, hexadecimal numbers, exponentials, leading and trailing decimal points, optional plus/minus signs and Infinity/NaN keywords while removing most of the jsonnet-specific code.

I'm not familiar with pygments or regex, so any feedback would be appreciated.


Anteru:
Merged, thanks!

Issue #2543, opened 2023-10-18
java: method definition aren't correctly recognized when the return type has a parameter type


wcazzola:
Consider this piece of code 

```java
public class ClassLoader {
   protected ClassLoader() { ... }
   protected ClassLoader(ClassLoader parent) { ... }
  
   public final ClassLoader getParent() { ... } 
   public static ClassLoader getSystemClassLoader() 
             throws SecurityException, IllegalStateException { ... }
   public static ClassLoader getPlatformClassLoader() throws SecurityException { ... }
   
   protected Class<?> findClass(String name) throws ClassNotFoundException { ... }
   protected Class<?> findClass(String moduleName, String name) 
             throws ClassNotFoundException { ... }
   protected Class<?> findLoadedClass(String name) { ... }
   protected Class<?> findSystemClass(String name) throws ClassNotFoundException { ... }
   
   public Class<?> loadClass(String name) throws ClassNotFoundException { ... }

   protected final void resolveClass(Class<?> c) throws NullPointerException { ... }
   protected Class<?> defineClass (String name, byte[] b, int off, int len) 
             throws ClassFormatError { ... }
                       ...
}
```
the name of the methods in their declarations should be recognized as such but this doesn't happen when the return type has a type parameter as <?> 

This is what I get with pygmentize 2.16.1

![image](https://github.com/pygments/pygments/assets/3992503/c15f9b9e-9675-48d1-96e8-5460d1c056c7)

As you can notice getParent, getSystemClassLoader, resolveClass also the constructors are correctly colored but those methods whose return type is Class<?> as loadClass or defineClass are not recognized.

PR #2737, opened 2024-06-20
Fix Issue #2543

JasonZhao5:
…eturn types like ‘Class<?>’.


JasonZhao5:
Add ‘?’ to the symbols that the expression can match to accommodate return types like ‘Class<?>’.
image

Anteru:
Merged manually. Thanks!

PR #2740, opened 2024-06-21
Fix << operator parsing in RubyLexer and add test

Cyberunk:
Summary

This PR fixes an issue in `RubyLexer` where `<<` is misinterpreted as the beginning of a heredoc when it is used in certain contexts, such as `[]<<a`.

Changes

- Added a new rule in `RubyLexer` to correctly interpret `<<` as an operator when it is not part of a heredoc.

Testing

- Added/Updated test case `testShiftLeftWithoutSpace` to cover this scenario.
- Ran the entire test suite to ensure no other tests are broken.


Anteru:
Merged, thanks a lot!

Anteru:
Sorry, bailing out again. The test doesn't test anything actually, and it fails the automated tests.

Issue #2293, opened 2022-12-03
MATLAB: Comments not recognized as nestable


Keyacom:
The MATLAB lexer does not recognize nested multiline comments. The below code is proof that Highlight.js (the lib used by GitHub) gets it right:

```matlab
% These are the derivatives for each term
d = [0 -1 0];

%{
  %{
    (Example of a nested comment, indentation is for cosmetics (and ignored).)
  %}
  We form the sequence, following the Taylor formula.
  Note that we're operating on a vector.
%}
seq = d .* (x - c).^n ./(factorial(n))

% We add-up to get the Taylor approximation
approx = sum(seq)
```

As for Pygments:
![obraz](https://user-images.githubusercontent.com/70766223/205461614-b10c9de5-1263-482c-8cdc-50cda7c733d4.png)

The black section of text between `%}` and `%}` should be highlighted like a comment, not like normal text.

VermouthDry:
![image](https://github.com/pygments/pygments/assets/141050364/adc51052-3398-4c33-8ff1-f02650e703ac)
I think the issue is caused by the code of 'blockcomment'. In this situation, the comment directly stoped when it comes to the first '%}' because of '#pop' and the state returns to 'root'. As a result, the following text failed to be highlighted like a comment since the state is incorrect.

PR #2741, opened 2024-06-22
Sloving issue #2293

VermouthDry:
The MATLAB lexer does not recognize nested multiline comments. The code provided in the issue description demonstrates that Highlight.js, the library used by GitHub, correctly handles nested comments.

This pull request modifies the MATLAB lexer to correctly handle nested multiline comments.

Changes Made:
Modified pygments/lexers/matlab.py
Changed the handling of block comments to use #push and #pop states for nested comments.
Added logic to enter a new block comment state on encountering the %{ marker and to return to the previous state on encountering the %} marker.

Justification:
When encountering the multi-line comment start marker %{, use #push to push the current state onto the stack and enter a new comment state. This allows handling nested multi-line comments. Use #pop to pop the current state from the stack and return to the previous state. This ensures that nested comments are correctly closed and the lexer returns to the appropriate state.

### DIFF
--- a/pygments/lexers/matlab.py
+++ b/pygments/lexers/matlab.py
@@ -2671,7 +2671,8 @@ class MatlabLexer(RegexLexer):
             include('expressions')
         ],
         'blockcomment': [
-            (r'^\s*%\}', Comment.Multiline, '#pop'),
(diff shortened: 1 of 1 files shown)


Anteru:
Thanks for your contribution. Would you mind adding a test for the new functionality?

PR #2743, opened 2024-06-25
Use alias_filenames instead of filenames for HTML/XML+Evoque lexers.

renfeisong:
This PR moves the `.html` and `.xml` extensions from `filenames` to `alias_filenames` for the HTML/XML+Evoque lexers, which is now consistent with all other “template+html/xml” lexers.


Anteru:
Merged, thanks!

Issue #2082, opened 2022-03-07
SLexer - break, next, else, in, repeat as Name.Function, not Name


vladbazon:
By SLexer, 'break(' IS a Name.Function token, but 'break' (without '()') is not.

For highlighting 'break', 'else' etc. -- we exclude him from the key 'keywords' (in r.py):
        'keywords': [
            (r'(if|for|while|return|switch|function)'
             r'(?![\w.])',
             Keyword.Reserved),
        ],

and we specify him on the _first scanned state in 'root':
        'root': [
            # calls:
            **(r'\b(else|next|break|repeat|in)\b', Name.Function),**
            (r'(%s)\s*(?=\()' % valid_name, Name.Function),
            include('statements'),
            # blocks:
            (r'\{|\}', Punctuation),
            # (r'\{', Punctuation, 'block'),
            (r'.', Text),
        ],

Indeed, a very simple solution...

Jean-Abou-Samra:
Would you be willing to prepare a PR?

mjkallen:
I found this issue, because I was surprised that the SLexer doesn't identify `function` as a reserved keyword, but as a function name. The list of keywords in https://github.com/pygments/pygments/blob/f64833d9d0443ff2443f8f8a4357c00196734d6c/pygments/lexers/r.py#L95 looks good to me, but my understanding is that the lexer identifies `function` in something like `f <- function(x) x^2` as a (regular) function name. Even though `function` and other language constructs like `if` and `break` look like functions to R, for highlighting purposes, it makes sense to treat these as reserved keywords. In case of `break` this also includes `break()`. So @jeanas's comment in PR #2087  about simply moving `include('keywords')` to come before`include('valid_name')` makes sense to me.

PR #2750, opened 2024-07-04
Improve SLexer

mjkallen:
This fixes the issue with keywords never being highlighted as such, because they were recognized as (regular) functions. See #2082.

Also adds the "new" forward pipe operator ('|>') which was introduced in R version 4.1


Anteru:
Thanks a lot, merged!

PR #2751, opened 2024-07-09
Add a TableGen lexer

frasercrmck:
This commit adds a basic lexer for LLVM's TableGen DSL[1][2].

The major design decision is about how to tokenize "code blocks" since those are, at a TableGen language level, string literals. In practice, however, in LLVM 99% of the time they are snippets of C++ which are processed by a TableGen backend. This would ideally be a configurable property of the lexer. For the initial implementation, it is assumed that code blocks contain C++ code and so the CppLexer is invoked on their contents.

[1] https://llvm.org/docs/TableGen/index.html
[2] https://llvm.org/docs/TableGen/ProgRef.html

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -50,6 +50,7 @@ Other contributors, listed alphabetically, are:
 * Ian Cooper -- VGL lexer
 * David Corbett -- Inform, Jasmin, JSGF, Snowball, and TADS 3 lexers
 * Leaf Corcoran -- MoonScript lexer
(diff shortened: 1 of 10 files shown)


frasercrmck:
Some example output:

image

josh146:
Just want to add that this will be a great addition, as we are currently adding a lot of TableGen code blocks to our sphinx project!

Anteru:
I'm sure everyone working on LLVM will appreciate it, merged for 2.19, thanks!

PR #2754, opened 2024-07-16
Use better multiline comment for Java

Moosems:
The old one didn't match with the following code:

```python
from re import MULTILINE, Match, compile
code: str = """/**
 * Code Multiline Comment
 */"""
match: Match[str] | None = compile(r'/\\*.*?\\*/', flags=MULTILINE).search(code)
print(match)
```


birkenfeld:
The lexer compiles all expressions with `re.DOTALL` which has the same effect as doing `(.|\n)`. Did you actually observe a wrongly highlighted comment somewhere?

On the other hand, your change introduces a bug since it removes the escaping of the opening asterisk.

Moosems:
I didn't realize the DOTALL, closing issue :D

PR #2755, opened 2024-07-16
Update LLVM lexer

frasercrmck:
* `splat` is now a keyword
* Hexadecimal floating-point constants can begin with `0x[KLMHR]` which denote different floating-point types: various long doubles, `half`-typed and `bfloat`-typed, respectively.


birkenfeld:
LGTM, thanks!

PR #2757, opened 2024-07-19
Add 'Change-By' field to Debian control lexer

sblondon:
The content of the 'Change-by' field has the 'Maintainer' field format: https://www.debian.org/doc/debian-policy/ch-controlfields.html#changed-by
so it should use it instead of the default parsing.

I added a test with a fake e-mail address ending with '.test' as top level domain instead of '.tld' previously because someone learn me the existence of the [RFC 2606](https://datatracker.ietf.org/doc/html/rfc2606) (Reserved Top Level DNS Names). I can modify the other occurrences in the file but I'm sure it's really useful. What do you think on this point?

### DIFF
--- a/pygments/lexers/installers.py
+++ b/pygments/lexers/installers.py
@@ -278,7 +278,7 @@ class DebianControlLexer(RegexLexer):
     tokens = {
         'root': [
             (r'^(Description)', Keyword, 'description'),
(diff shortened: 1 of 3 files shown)


Anteru:
Merged, thanks!

PR #2758, opened 2024-07-29
Fix regex for triple quoted strings in Turtle lexers

### LINKED ISSUE #2744: Turtle highlighting does not properly terminate on multi-line strings
See title. If a string has a newline character in it, it'll keep highlighting the rest of the file like the string keeps going.

Reproduction example: `curl https://health-ri.sandbox.semlab-leiden.nl/dataset/fdf426ca-0c23-4554-aee9-501e0bb61dae\?format\=ttl | pygmentize -l turtle`

Markus92:
The test cases for the turtle lexer did not include triple quoted strings (neither single nor double). The regex used to match the string contents matched every single character except escape characters, but crucially it also matched the triple quotes used to delineate the termination of such a string. The updated regex specifically excludes these characters.

Fixes #2744

Note: the same issue might also occur in the SPARQL lexer. However, I never really use SPARQL so I cannot verify if the same logic would apply.

This is my first contribution to Pygments, forgive me for any mistakes I made. I confirm I agree, am willing and am able to put my contributions under the BSD 2-clause license.

Before:
afbeelding

After:
afbeelding

### DIFF
--- a/pygments/lexers/rdf.py
+++ b/pygments/lexers/rdf.py
@@ -280,7 +280,7 @@ class TurtleLexer(RegexLexer):
         ],
         'triple-double-quoted-string': [
             (r'"""', String, 'end-of-string'),
-            (r'[^\\]+', String),
+            (r'[^\\]+(?=""")', String),
             (r'\\', String, 'string-escape'),
         ],
         'single-double-quoted-string': [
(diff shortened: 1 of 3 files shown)


Markus92:
@birkenfeld Any chance you could take a look at this? Thanks!

birkenfeld:
LGTM. Thanks!

PR #2761, opened 2024-08-05
feat(styles): add a Srcery theme

jamietanna:
This is largely adapted from the colours used by Gruvbox Dark, and then
modified to work for Srcery's respective colours.

Determining `Generic.Prompt` and `Generic.Output` was a little more
difficult, and aren't quite one-for-one in terms of the colours used by
Gruvbox.

[0]: https://srcery.sh

### DIFF
--- a/pygments/styles/srcery.py
+++ b/pygments/styles/srcery.py
@@ -1,0 +1,70 @@
+"""
+    pygments.styles.srcery
+    ~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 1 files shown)


MindTooth:
Sad to not see this merged yet. 😔   How do you think we go about getting it merged?

jamietanna:
No I don't, sorry - don't wanna tag maintainers, but I do at least see maintainers are active and busy, so hopefully at some point!

Issue #1815, opened 2021-05-23
[tsx] TypescriptLexer does not support tsx


sk-:
Currently `tsx` is supposed to be supported via the [`TypescriptLexer`](https://github.com/pygments/pygments/blob/0da640a6bf65654bf28940119bd379d2887e93af/pygments/lexers/javascript.py#L444-L453), however there are several issues when parsing tsx files. I'd suggest removing that support for the `TypescriptLexer`, just like it is for the `JavascriptLexer`, which does not support `jsx`. Additionally prioritizing `tsx` and `jsx` support would be highly appreciated, so that we don't need to rely on external packages like https://github.com/richardbann/pygments-lexer-babylon. (See issues #868 and #709.) 

## Errors
- Text inside nodes is parsed as `Token.Name` or `Token.Keyword`
- `<`, `>`, and `/` are parsed as operators
- Closing tags generate errors

Below is a comparison of the same markup parsed with the html and typescript parser.

## HTML
```html
    <div>
      this is some
      <p>text</p>
    </div>
``` 

pygments tokens

```
 (Token.Punctuation, '<'),
 (Token.Name.Tag, 'div'),
 (Token.Punctuation, '>'),
 (Token.Text, '\n      this is some\n      '),
 (Token.Punctuation, '<'),
 (Token.Name.Tag, 'p'),
 (Token.Punctuation, '>'),
 (Token.Text, 'text'),
 (Token.Punctuation, '<'),
 (Token.Punctuation, '/'),
 (Token.Name.Tag, 'p'),
 (Token.Punctuation, '>'),
 (Token.Text, '\n    '),
 (Token.Punctuation, '<'),
 (Token.Punctuation, '/'),
 (Token.Name.Tag, 'div'),
 (Token.Punctuation, '>'),
```

## TSX

```tsx
function f() {
  return (
    <div>
      this is some
      <p>text</p>
    </div>
  )
}
```

pygments tokens

```
 (Token.Operator, '<'),
 (Token.Name.Other, 'div'),
 (Token.Operator, '>'),
 (Token.Text, '\n      '),
 (Token.Keyword, 'this'),
 (Token.Text, ' '),
 (Token.Name.Other, 'is'),
 (Token.Text, ' '),
 (Token.Name.Other, 'some'),
 (Token.Text, '\n      '),
 (Token.Operator, '<'),
 (Token.Name.Other, 'p'),
 (Token.Operator, '>'),
 (Token.Name.Other, 'text'),
[...]

PR #2762, opened 2024-08-06
Add support for tsx to jsx lexer

englut:
I noticed that my `.tsx` files are not supported by Pygments currently. Adding the file type to the `jsx` lexer works well, at least for my purposes!

image

ran `tox` and everything seems to pass okay!

Should address #1815


Anteru:
Thanks for the contribution. I'm closing this as  I've added a "proper" TSX lexer, given you're absolutely right, it is a fairly trivial addition. See df8d3eaa55a5ca818c47b5ced99c282446fe6cad

PR #2767, opened 2024-08-15
Fix quoted string containing comment character

### LINKED ISSUE #2720: Incorrect Rendering of .gitconfig File Contents
Description:
When attempting to display the contents of a .gitconfig file in INI format, there is an issue with the rendering of comments. Currently, anything after a semicolon is treated as a comment, even if it is not intended to be. This problem becomes apparent when using tools like mkdocs to display the file contents.

Steps to Reproduce:
1. Create a .gitconfig file with the following contents:

~~~
```cfg
[alias]
    test = "some stuff with a semi colon ; and then more stuff"
    test2 = "some more stuff with a semi colon ; and then more stuff" # (1)!
```
~~~

2. Use mkdocs or a similar tool to display the file contents.

Expected Result:
The .gitconfig file should be displayed accurately, showing the contents as intended without any rendering issues.

Actual Result:
The rendered display of the .gitconfig file, using mkdocs, shows incorrect rendering. The comments associated with the "test2" alias are consumed by an annotation, leading to missing text in the rendered output.

image

Additional Information:
Here's a real-world example of a config value where the semicolon does not indicate a comment:
~~~
```cfg
[alias]
   fetch-reset = "!git fetch ; git reset --hard @{u}"
```
~~~

sblondon:
The comment character ('#' or ';') included in a quoted value does not be interpreted as beginning of comment but as a character of the quoted value.

This PR fixes #2720.

### DIFF
--- a/tests/snippets/ini/test_quoted_entries.txt
+++ b/tests/snippets/ini/test_quoted_entries.txt
@@ -1,0 +1,36 @@
+---input---
+[section]
+    key 1 = "value1"
+    key 2 = "value2" # comment
+    key 3 = "value3 ; value 3bis" ; comment
+
+---tokens---
+'[section]'   Keyword
+'\n    '      Text.Whitespace
+'key 1'       Name.Attribute
+' '           Text.Whitespace
+'='           Operator
+' '           Text.Whitespace
+'"'           Literal.String
+'value1'      Literal.String
+'"'           Literal.String
+'\n    '      Text.Whitespace
+'key 2 = '    Literal.String
+'"'           Literal.String
+'value2'      Literal.String
+'"'           Literal.String
+' '           Text.Whitespace
+'# comment'   Comment.Single
+''            Literal.String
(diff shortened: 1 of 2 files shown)


tests/snippets/ini/test_quoted_entries.txt
```diff
@@ -0,0 +1,36 @@
...
+'\n    '      Text.Whitespace
+'key 1'       Name.Attribute
+' '           Text.Whitespace
+'='           Operator
+' '           Text.Whitespace
+'"'           Literal.String
+'value1'      Literal.String
+'"'           Literal.String
+'\n    '      Text.Whitespace
+'key 2 = '    Literal.String
```
kartben, line 18: this is actually highlighting a bug in the current implementation, right?

Anteru: Yes. Thanks for spotting this. Tracked in #2835

PR #2769, opened 2024-08-23
Agda: add `opaque` and `unfolding` keywords

ncfavier:
The keywords were added in Agda 2.6.4, see https://agda.readthedocs.io/en/v2.6.4/language/opaque-definitions.html

The current list of keywords is at https://github.com/agda/agda/blob/84b0c60b418476d63d5287a5751da43ceab33735/src/full/Agda/Syntax/Parser/Lexer.x#L159-L200


Anteru:
Merged, thanks!

PR #2772, opened 2024-08-25
Reduce analyse_text priority for Mojo relative to Python

### LINKED ISSUE #2771: MojoLexer's analyse_text should not have higher priority for "import " than PythonLexer
The `analyse_text` method for `MojoLexer` in https://github.com/pygments/pygments/blob/master/pygments/lexers/mojo.py#L701 is a near-copy of that from `PythonLexer` in https://github.com/pygments/pygments/blob/master/pygments/lexers/python.py#L408.

A code sample that contains `import foo` thus evaluates as Mojo rather than Python. Python is more likely, so Mojo should have a lower priority.

dlazin:
Score "import " as 0.9 for Mojo. Python returns True for this (implicitly 1.0). Resolves #2771.

### DIFF
--- a/pygments/lexers/mojo.py
+++ b/pygments/lexers/mojo.py
@@ -699,6 +699,9 @@ class MojoLexer(RegexLexer):
     }
 
     def analyse_text(text):
-        return (
-            shebang_matches(text, r"mojo?") or "import " in text[:1000]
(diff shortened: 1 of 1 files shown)


birkenfeld:
Looks good, thanks!

PR #2774, opened 2024-08-28
Handle / in postgres explain

anayrat:
Hello,
I noticed pygments do not handle the `/` in : 
```
I/O Timings: shared/local read=109734.296
```


birkenfeld:
Thanks!

PR #2781, opened 2024-09-13
Drop executable bit on non-executable files

Fryguy:
We have a security scanner that noticed that these files have the executable bit set, but they don't have a shebang and so couldn't be executed.


birkenfeld:
Thanks!

PR #2782, opened 2024-09-15
Initial Gleam support

aognio:
- Initial support for the Gleam programming language (gleam.run)
- Based on grammar rules from https://github.com/aognio/tree-sitter-gleam
- Also tried to improve the lexer taking some bits from https://github.com/gleam-lang/gleam-highlight.js/blob/main/index.js

Here's an screenshot of how it looks:

gleam-pygments-lexer-github-style

We can compare this output to the same file on Github here:

https://github.com/gleam-lang/example-todomvc/blob/main/src/todomvc.gleam

### DIFF
--- a/pygments/lexers/_gleam_builtins.py
+++ b/pygments/lexers/_gleam_builtins.py
@@ -1,0 +1,25 @@
+"""
+    pygments.lexers._gleam_builtins
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+    Gleam builtins.
+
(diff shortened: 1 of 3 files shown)


Anteru:
This is missing a test file. Can you please add a test and make sure it passes?

Anteru:
Realized Gleam was merged already: https://github.com/pygments/pygments/blob/master/pygments/lexers/gleam.py, and this PR was opened after the previous one was merged? I also guess that means the changelog is actually wrong on this one.

Issue #2747, opened 2024-06-29
Add `deb822` format lexer for `*.sources` files


chrysle:
`deb822` is a newer format to specify system-wide Debian package repositories on Ubuntu in specific fragments (under **/etc/apt/sources.list.d/*.sources**) and an alternative to **sources.list**, for which a lexer does already exist. It is supported by APT itself since version _1.1_, but lacked support in most of Ubuntu's graphical package management utilities – until Ubuntu decided to remove the **sources.list** in favour of the new format in the recently released version 24.04. Most tools have since been catching up with the decision, thus it should be quite useful to create a new lexer for the format (technical specification [here](https://discourse.ubuntu.com/t/spec-apt-deb822-sources-by-default/29333)) and possibly deprecate the **sources.list** lexer.

sblondon:
`sources.list` is currently used in Debian so deprecating seems to be a premature action. I don't know if the replacement of `sources.list` by `*.sources` files is an Ubuntu-project only.

sblondon:
After discussing with Debian Developers, `deb822` is currently supported by Debian (Bullseye and perhaps older versions too) but is not enabled by default. `deb822` will replace `sources.list` format in the future for repository.

birkenfeld:
Even if the format itself is deprecated by Debian, that's no reason to deprecate its lexer in Pygments. It's going to be in use for a long time still. (Also, we currently have no mechanism to deprecate lexers in the first place...)

Regarding the new format, would it make sense to add `*.sources` to the control file lexer for now?

chrysle:
> Even if the format itself is deprecated by Debian, that's no reason to deprecate its lexer in Pygments. It's going to be in use for a long time still. 

Agreed.

> Regarding the new format, would it make sense to add `*.sources` to the control file lexer for now?

Not really, as a Debian `control` file uses `deb822` syntax, but has [completely different fields](https://www.debian.org/doc/debian-policy/ch-controlfields.html#list-of-fields) than `*.sources` files; these are listed in the [`sources.list` manpage](https://manpages.debian.org/bookworm/apt/sources.list.5.en.html#THE_DEB_AND_DEB-SRC_TYPES:_GENERAL_FORMAT). But the current control file lexer may well be used as a template, probably I'll have some time to work on that soon.

sblondon:
> > Even if the format itself is deprecated by Debian, that's no reason to deprecate its lexer in Pygments. It's going to be in use for a long time still.

[...]

PR #2788, opened 2024-10-05
Add DebianSourcesLexer lexer

sblondon:
Debian and derivatives have a new *.sources file based on deb822 format to set up system-wide packages repositories.
Ubuntu uses already it since [24.04 release](https://discourse.ubuntu.com/t/spec-apt-deb822-sources-by-default/29333).
This format will replace the sources.list format in the future.

I added the next release (2.19) in `version_added` but, of course, it should be changed if it will be release in another release. I can remove it too if a maintainer wants to manage it.

closes: #2747


Anteru:
Merged for 2.19! Thanks!

PR #2789, opened 2024-10-07
Various improvements to the BQN lexer

slotThe:
EDIT: Apologies for the somewhat messy PR; perhaps conceptually some of these should go into different PRs. However, the changes are so small that I figured it was perhaps best to include them here as well.

#### Fix BQN lexer treating special characters as word chars

`\w` matches all alphanumeric Unicode characters, including ones (e.g., 𝕊) that BQN treats special. This is especially troublesome for variables; previously, something like

    𝕊i

would have returned

    (Token.Operator, '𝕊'),
    (Token.Error, 'i'),

instead of

    (Token.Operator, '𝕊'),
    (Token.Name.Variable, 'i').

This extends to special sequences like `\b`, which care about the difference between `\w` and `\W`.

---

~~There are still some `\b`'s in the code, but from what I can tell these don't produce any undesirable effects. E.g., for 2-modifiers, the BQN grammar already disallows writing something like `F_𝕣_G` instead of `F _𝕣_ G`. No guarantees that I caught everything, of course.~~

This is, of course, nonsense, there are plenty of situations where this could arise that are accepted by BQN; e.g., `_F ← {𝔽_𝕣}`, `_D ← {𝔽_F𝕩}`, etc. See the fixup commit.

#### BQN Lexer: Replace function * with ⋆

BQN does not actually use * (ASTERISK) anywhere, but there is a primitive function ⋆ (STAR OPERATOR).

#### BQN Lexer: Allow underscores in numbers

As per the spec.

### DIFF
--- a/pygments/lexers/bqn.py
+++ b/pygments/lexers/bqn.py
@@ -26,6 +26,10 @@ class BQNLexer(RegexLexer):
     mimetypes = []
     version_added = '2.16'
 
+    # An inter_word_char. Necessary because \w matches all alphanumeric
+    # Unicode characters, including ones (e.g., 𝕊) that BQN treats special.
+    _iwc = r'((?=[^𝕎𝕏𝔽𝔾𝕊𝕨𝕩𝕗𝕘𝕤𝕣])\w)'
+
     tokens = {
         'root': [
             # Whitespace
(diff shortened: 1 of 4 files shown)


Anteru:
All in one PR is fine for this kind of small stuff, merged!

PR #2794, opened 2024-10-13
Add Rego lexer

sebastien-rosset:
Add support for [Rego](https://www.openpolicyagent.org/docs/latest/policy-language/), the policy language for the Open Policy Agent (OPA) framework.

Create a new `rego.py` module.


florendelacruz (commented):
Si

PR #2798, opened 2024-10-22
Add operator ^ to Matlab, Octave and Scilab lexers

kcs:
Currently the Matlab (including octave and SciLab) lexers do not recognize the `^` operator. Only the element-wise operator `.^` is recognized. Although most of the time the element-wise operator is used, there are some cases for the normal operator to be used, like for example scalar power operation (`2^16`).

I also added a test case for this operator in the matlab sample code.


birkenfeld:
LGTM. Thanks!

PR #2800, opened 2024-11-01
Update Macaulay2 symbols for verson 1.24.11

d-torrance:
Also re-sort the symbols in ASCII order, as this is the new behavior of the script that generates this file.


birkenfeld:
LGTM, thanks!

Issue #2805, opened 2024-11-08
Out of context highlighing of contextual keywords in csharp (starting with "file")


jofish920:
Following on from #2710, which pointed out that previously the csharp lexer was not properly identifying new contextual keywords as keywords, I looked at some contexts in which the current solution might overgeneralise.  Here is an example (not good C#, but an effective illustration):

![image](https://github.com/user-attachments/assets/390eeacd-7fb4-477b-b41d-0ec0624d5f19)

As can be seen (and confirmed by viewing the output through the raw formatter), the current code is treating almost every instance of "file" as a keyword.

In case it isn't clear in the image, the code used to produce the image above is as follows:

```python
file record X(int file);
file record Y(int blah);

public static class Program {
    static int file(Func<int,int> file) => file(13);
    static int blah(Func<int,int> blah) => blah(13);

    public static void Main(string[] args) {
        int file = Program.file(file => 42);
        int blah = Program.blah(blah => 12);
        X x = new(9);
        Console.WriteLine($"{file}{blah}{x}{args}");
    }
}
```

I wonder if the results could be improved by only recognising "file" as a keyword if it is followed by something like "record", "class", or "enum", and how many of the other contextual keywords would be susceptible to similar handling.

PR #2806, opened 2024-11-09
tests issue #2805 - overgeneralised recognition of file as keyword in csharp lexer

jofish920:
This PR adds some unit tests that point out instances where `pygments.lexers.dotnet.CSharpLexer` currently overgeneralises in recognising "file" as a keyword (issue #2805).  These tests are currently set up as expected fails under pytest, but can be made active on a case-by-case basis by changing a flag.  The issue of calls following "await" being incorrectly tagged with `Token.Name.Function` rather than `Token.Name` is actually a different issue, but was considered as a candidate and left in as a problematic case.

It is my intention to try fixing these and related issues at some point in the near future, but for now these tests identify cases that I think shold be addressed.


jofish920:
On second thoughts, I will mark this as a draft for now and actually try to implement some fixes first.

jofish920:
As far as I can see, the pull request now adequately deals with overgeneralised tagging of 'file' as a keyword.
The unit test covers at least the basic cases.

The one expected fail is due to a different issue, involving expressions of the form `await Name(...)` being misidentified as method definitions (I think).  Looking into that and the classification of the remaining contextual keywords is something for later.

Anyway, I think that this is now ready for review; aside from the additional tests, only a couple of lines ended up being changed.

birkenfeld:
Thanks for the PR! The fix looks good to me.

In general, we'd prefer tests that check for desired token output in the existing "snippet" format to custom test functions/methods. Are you ok with changing that over?

jofish920:
I'll certainly have a look at it.  While I've made use of pygmentize for a while, my first dive into the code was just over a week ago.  However, from what I remember, converting the tests to the snippet format shouldn't be too difficult.

I mostly wrote the test the way I did because I wasn't sure how long it was going to take me to get around to working on the issue, and I wanted to write some test cases in advance; the warm fuzzy feeling that it gave me is probably less important than having something that fits in with the rest of the code :)

Thanks for the quick response to the submission.  I'm going to get some sleep before I work on anything, though.  It's just past 2:30am where I am, and I have things to get done tomorrrow.

jofish920:
The commit message really says it all.  I've switched testing over to the "snippet" format and run the tests to ensure that everyting still works.  Sorry about bailing last night, but I was really getting too tired to think properly.

birkenfeld:
Very nice, thank you!

PR #2809, opened 2024-11-12
fix: get fonts inside subdirectories on macOS

cherryblossom000:
I use [Home Manager](https://github.com/nix-community/home-manager) with Nix on macOS to install new fonts, which stores the fonts in `~/Library/Fonts/HomeManager/{opentype,truetype}`. However, the `FontManager` was not able to get these fonts so I couldn’t use them when rendering as an image.

This PR fixes this by adding all the fonts inside subdirectories of `~/Library/Fonts`, `/Library/Fonts`, etc. to the font map.


Anteru:
Merged, thanks.

PR #2810, opened 2024-11-12
Fixes #2692: xml lexer mistags doctype with inline dtd

### LINKED ISSUE #2692: Wrong highlight when using inline dtd in xml document
When trying to highlight following xml, the `]>` at the end of `DOCTYPE` gets highlighted as normal text.
Possible because, in dtd the `ELEMENT` does not have a closing `/`.
```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE hello [<!ELEMENT hello (#PCDATA)>]>
<hello>Hello Word!</hello>
```
Using:
```python
from pygments.lexers import XmlLexer
from pygments.formatters import HtmlFormatter

code = '<?xml version="1.0" encoding="UTF-8"?><!DOCTYPE hello [<!ELEMENT hello (#PCDATA)>]><hello>Hello Word!</hello>'
print(hightlight(code, XmlLexer(), HtmlFormatter()))
```

Results in:
grafik

https://pygments.org/demo/?lexer=xml&style=default&formatter=html&code=%3C%3Fxml+version%3D%221.0%22+encoding%3D%22UTF-8%22%3F%3E%0A%3C%21DOCTYPE+hello+%5B%3C%21ELEMENT+hello+%28%23PCDATA%29%3E%5D%3E%0A%3Chello%3EHello+Word%21%3C%2Fhello%3E

jofish920:
This fixes the bug that was mentioned in #2692, and also adds support for highlighting the inline DTD.
The latter is  achieved using a `DelegatingLexer`, which has been made a new base class for `XmlLexer`.
I did end up removing the MIME types associated with the two new lexers, mostly to get `tox` to stop
complaining, but also because I really don't see the output as something that should be available except
on request.

While I feel this is ready for review, I am more hoping for a little feedback than anticipating that the fix is
going to be included in the main branch.


jofish920:
OK.  So some fixes are required.  I'll need to look at this later; the error messages weren't the most informative in this case.

jofish920:
OK.  The unnecessary imports are where I removed a unit test that I had added earlier, and are easily fixed.  The fact that `regexlint` dies with an uncaught exception instead of producing a response is slightly concerning.  The error is as follows (slightly edited to remove long paths):

    multiprocessing.pool.RemoteTraceback: 
    """
      File "/usr/lib/python3.12/multiprocessing/pool.py", line 125, in worker
        result = (True, func(*args, **kwds))
                        ^^^^^^^^^^^^^^^^^^^
      File "XXX/pygments/.tox/regexlint/lib/python3.12/site-packages/regexlint/cmdline.py", line 183, in check_lexer_map
        return check_lexer(*args)
               ^^^^^^^^^^^^^^^^^^
      File "XXX/pygments/.tox/regexlint/lib/python3.12/site-packages/regexlint/cmdline.py", line 221, in check_lexer
        reg = Regex.get_parse_tree(pat[0], cls.flags)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "XXX/pygments/.tox/regexlint/lib/python3.12/site-packages/regexlint/parser.py", line 350, in get_parse_tree
        return cls._get_parse_tree(s, flags, effective_flags)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "XXX/pygments/.tox/regexlint/lib/python3.12/site-packages/regexlint/parser.py", line 428, in _get_parse_tree
        c = open_stack[-1].children.pop()
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    IndexError: pop from empty list
    """

Currently I'm unsure what to do about this; my brief attempt at debugging `regexlint` indicates that tracking down the bug could take some time, and I know that the code being linted works as intended.  Given that the bug doesn't occur on the "master" branch, it must be triggered by something that I've done in updatig the lexers.  I'm just not sure what - it all looks rather straight-forward to me, with the possible exceptions of non-greedy matches (seen elsewhere) and a single character negative lookahead.

jofish920:
I was going to split the handling of comments and '<![...[...]>' constructs into different sections, but I don't think that the current expressions should cause any problems.  For the moment I'll consider this ready for review.  It is a relatively minor change that fixes #2692 and allows for a possible extension in hilighting iline DTDs that I don't anticipate will be used very often.

Anteru:
I just came across this, and it looks good to me, but I'm not sure I follow the problem with regexlint. Does it fail for you? I'm trying to set up a release right now and I'm inclined to skip merging this if there are problem with regexlint, mostly because I don't have the time right now to debug this, and I'd be happy to resolve this _after_ the release.

jofish920:
The problem was that regexlint fails when it encounters a regular expression that sets flags using the syntax '(?i)...' rather than '(?i:...)'.  I was using the former, mostly out of habit, and worked around the problem by switching to the latter in commit [a513cdc].  It has been a while since I looked at it, but the current version of the PR should be OK.

This is a general workaround, so a fix to regexlint would be nice but is not essential.

[a513cdc]: https://github.com/pygments/pygments/pull/2810/commits/a513cdca08ca7b0440bb5dc7d3fd4769026a7a42

PR #2812, opened 2024-11-13
fixes #2811: exit from bad utf-8 in check_sources

### LINKED ISSUE #2811: Decoding error causes abnormal termination in "scripts/check_sources.py" when running "tox -e check"
Running "tox -e check" resulted in an uncaught exception of type `UnicodeDecodeError` raised by code in `main` within "scripts/check_sources.py" due to a failure in decoding the contents of one of the files in my ".venv" directory.  The code catches exceptions due to the file being missing, but not improperly encoded.  Dealing with the two  types of exception in the same way would be my preferred solution.  A different issue is why the files in ".venv" were being checked in the first place?

jofish920:
Came across a case where running "tox -e check" resulted in the script "test/check_sources.py" exiting due to an uncaught exception rather than reporting an error in the file being checked.  This adds an exception handler almost identical to the one right above it that deals with unopenable  files.  Still an apparent problem with scanning files in ".venv".

### DIFF
--- a/scripts/check_sources.py
+++ b/scripts/check_sources.py
@@ -179,6 +179,10 @@ def main(argv):
                 print(f"{fn}: cannot open: {err}")
                 num += 1
                 continue
+            except UnicodeDecodeError as err:
+                print(f"{fn}: error decoding: {err}")
+                num += 1
+                continue
 
             for checker in checkerlist:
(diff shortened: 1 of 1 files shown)


Anteru:
Merged, thank you!

PR #2826, opened 2024-11-30
gnuplot lexer: change comment token to use regex

### LINKED ISSUE #2825: `LatexFormatter`'s `mathescape` does not escape backslashes inside `gnuplot` comment
`LatexFormatter`'s `mathescape` does not escape backslashes inside `gnuplot` comment. This causes an unexpected error when compiling with LaTeX.

For example, let's consider the following `gnuplot` code:
```tex
# $\alpha$
```

`pygmentize` this with the following command:
```sh
python -m pygments -f latex -l gnuplot -O mathescape test.gnuplot
```

Expected result is:
```tex
\begin{Verbatim}[commandchars=\\\{\},codes={\catcode`\$=3\catcode`\^=7\catcode`\_=8\relax}]
\PY{c}{\PYZsh{} $\alpha$}
\end{Verbatim}
```

However, actual result is:
```tex
\begin{Verbatim}[commandchars=\\\{\},codes={\catcode`\$=3\catcode`\^=7\catcode`\_=8\relax}]
\PY{c}{\PYZsh{}}\PY{c}{ $}\PY{c}{\PYZbs{}}\PY{c}{alpha$}
\end{Verbatim}
```

zyoshoka:
Changed comment token to use regex in gnuplot lexer.

This will fix #2825.


birkenfeld:
This might fix the latex math issue, but it will stop recognizing comments with line continuations (backslash before the newline).

PS. What do you mean by "use regex"?

zyoshoka:
You are right. I've fixed the line continuation problem.

PS. Please ignore my comment above because I was lacking in understanding of the codebase...

PR #2609, opened 2023-12-11
add coffee theme

TruncatedDinoSour:
adds [coffee theme](https://github.com/coffee-theme/coffee/), based on its vim theme

image

PR #2830, opened 2024-12-18
[styles/coffee] Improve the colour of `builtin` & `PreProc`

ar1ja:
Before the colour of built-ins was plain. Now it is changed to be highlighted and reflected in the source code.

Previously added at https://github.com/pygments/pygments/pull/2609 :)

Before:

image

After:

image

It's essentially a change of one colour.

### DIFF
--- a/pygments/styles/coffee.py
+++ b/pygments/styles/coffee.py
@@ -31,7 +31,7 @@ class CoffeeStyle(Style):
     styles = {
         Comment: "#70757A",
         Comment.Hashbang: "#8f9f9f",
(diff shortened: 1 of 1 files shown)


TruncatedDinoSour:
Contrast seems to not have changed (still 2.9, verified) and since all the tests pass, I think this is final. Let me know if you have any changes to be made :)

TruncatedDinoSour:
Okay, I also decided to make the pre-processor tokens a little darker considering the ViM theme:

image

Issue #2834, opened 2025-01-05
WARNING: Lexing literal_block '...' as "ini" resulted in an error at token: '}'


ThomasWaldmann:
I originally filed this at sphinx, but guess it better fits here. It might be related to the 2.19 release as it first popped up today.

### Describe the bug

A part of our docs that didn't change since quite a while suddenly causes a warning from the ini lexer and thus sphinx terminates with rc 1 (which leads to termination of our CI pipeline).

```
data-structures.rst:971: WARNING: Lexing literal_block '...' as "ini" resulted in an error at token: '}'.
```
Details see there:

https://github.com/borgbackup/borg/pull/8625

I didn't see anything problematic in that ini section, even the {} were nested correctly.

### Problematic text

```
[cache]
version = 1
repository = 3c4...e59
manifest = 10e...21c
timestamp = 2017-06-01T21:31:39.699514
key_type = 2
previous_location = /path/to/repo

[integrity]
manifest = 10e...21c
files = {"algorithm": "XXH64", "digests": {"HashHeader": "eab...39e3", "final": "e2a...b24"}}
```

### Environment Information

```text
Platform:              darwin; (macOS-15.2-arm64-arm-64bit-Mach-O)
Python version:        3.13.1 (main, Dec  3 2024, 17:59:52) [Clang 16.0.0 (clang-1600.0.26.4)])
Python implementation: CPython
Sphinx version:        8.1.3
Docutils version:      0.21.2
Jinja2 version:        3.1.5
Pygments version:      2.19.0
```

ThomasWaldmann:
Maybe related: #2818 ?

potiuk:
Same in Airflow - our CI tests started to fail with errors when rendering our docs where `ini` example contains dictionary.

This seems to be a side-effect of (from the changelog https://pygments.org/docs/changelog/#version-2-19-0 :

> Ini: Fix quoted strings with embedded comment characters ([#2767](https://github.com/pygments/pygments/issues/2767), [#2720](https://github.com/pygments/pygments/issues/2720))

The errors we have can be seen here: https://github.com/apache/airflow/actions/runs/12622351781/job/35170297671#step:6:35195

For example:

```python
  29 | 
  30 |    Retrieves Connection or Variables from AWS SSM Parameter Store.
  31 | 
  32 |    Configurable via ``airflow.cfg`` like so:
  33 | 
> 34 |    .. code-block:: ini
  35 | 
  36 |        
[...]

PR #2835, opened 2025-01-05
Fix quoted string handling

kartben:
Commit 2dad7c0e (PR #2767) introduced regression in the handling of strings with double quotes.

The bug was even "featured" in the golden for test_quoted_entries.txt since `key 2` attribute was improperly parsed as literal string "key 2 = ".

https://github.com/pygments/pygments/blob/2dad7c0e1c83cace532ee469c885e7d07ed4a894/tests/snippets/ini/test_quoted_entries.txt#L1-L18


Anteru:
Thanks a lot for the PR!

kartben:
> Thanks a lot for the PR!

Oh wow, didn't expect this to be merged as it was still in draft :) There's, I think, several other issues with the current implementation that I was hoping to address, in addition to adding the snippets allowing to add the missing test coverage

ex:

```
    key 4 = { "abc" = "def" }
    key 5 = "hello 'world'"
```

"key 4" would be to have coverage for what @ThomasWaldmann mentioned in #2834, which is still not quite right, by a fair margin :)

image

Anteru:
But that's not really an `ini` file at that stage. Feels more like you'll want TOML or something else at that point?

kartben:
> But that's not really an `ini` file at that stage. Feels more like you'll want TOML or something else at that point?

ya, true -- given there's no common / agreed upon grammar for .ini files anyway, I think it's fair to say there will always be corner cases

PR #2837, opened 2025-01-07
Clean up sql.py

dlazin:
This is a no-op change that standardizes the imports in sql.py, which makes the file easier to work with for future improvements.


Anteru:
Good one, thanks a lot! Didn't realize how messy this got over time.

PR #2838, opened 2025-01-08
Add Ruff "RUF" Rules

reteps:
- I have a downstream project that cannot understand that some of the variables in the codebase are supposed to be mutable globals.

The addition of these lint rules adds annotations to all sets and dictionaries, marking them as mutable class variables. Additionally, all lists were converted to tuples when appropriate (e.g. a static list).


Anteru:
I'm not convinced that those changes actually help a lot and we're not using typing information either. Adding @jeanas , @birkenfeld - should we track this as a "Pygments 3.0" idea, to add typing information? I feel like this is a big enough change to warrant a bigger bump, and then we could also do things like this here.

reteps:
Closing as I won't work on this.

PR #2845, opened 2025-01-20
Implemented a lexer for UCSD's ELSA lambda calculus evaluator

Ant-28:
Created lexer for Elsa, [UCSD's lambda calculus evaluator](https://github.com/ucsd-progsys/elsa?tab=readme-ov-file). Documentation does not show multiline comments are possible, but they can be executed in Elsa.


Ant-28:
Following up on this

PR #2854, opened 2025-02-02
Add missing 'conf' and 'generate_buildrequires' directives to RPMSpecLexer. Fixes #2853.

### LINKED ISSUE #2853: RPMSpecLexer: directives 'conf' and 'generate_buildrequires' are missing
Directives 'conf' and 'generate_buildrequires' from [spec](https://rpm-software-management.github.io/rpm/manual/spec.html) are missing.

termim:
Add missing spec file directives from [rpm](https://rpm-software-management.github.io/rpm/manual/spec.html).

### DIFF
--- a/pygments/lexers/installers.py
+++ b/pygments/lexers/installers.py
@@ -156,7 +156,7 @@ class RPMSpecLexer(RegexLexer):
     url = 'https://rpm-software-management.github.io/rpm/manual/spec.html'
     version_added = '1.6'
 
-    _directives = ('(?:package|prep|build|install|clean|check|pre[a-z]*|'
+    _directives = ('(?:package|prep|conf|build|install|clean|check|generate_buildrequires|pre[a-z]*|'
                    'post[a-z]*|trigger[a-z]*|files)')
 
     tokens = {


alliasgher:
Drive-by note, since this has been open a while: I went to fix #2853 myself, found this PR, and stopped. But I did hit something worth flagging before it lands.

Adding a bare `conf` alternative shadows the `%config` and `%configure` macros, because `_directives` is matched at line start and the alternation is not anchored to a word boundary. On this branch:

```
%config(noreplace) /etc/foo.conf -> Name.Decorator '%conf' + Text 'ig(noreplace) /etc/foo.conf'
%configure                       -> Name.Decorator '%conf' + Text 'igure'
```

Both should stay `Keyword` (they are handled by the macro rule further down), and `%configure` in particular is very common in real spec files, so this trades one bug for a more visible one.

A lookahead is enough to fix it:

```python
_directives = ('(?:package|prep|conf(?![a-z])|generate_buildrequires|'
               'build|install|clean|check|pre[a-z]*|'
               'post[a-z]*|trigger[a-z]*|files)')
```

With that, `%conf` and `%generate_buildrequires` become `Name.Decorator` as this PR intends, while `%config` and `%configure` stay `Keyword`. `generate_buildrequires` needs no guard, as nothing else in the alternation starts with `g`.

I verified both variants against `master` and ran the full suite (`5330 passed, 16 skipped`) with the lookahead version. Happy to leave this entirely to you @termim, it is your PR and your fix, I just did not want the regression to slip through. I am not opening a competing PR.

PR #2882, opened 2025-04-05
fix lua regex causing runaway backtracking.

### LINKED ISSUE #2839: Lua code using spaces for indentation causes infinite loop.
The recent update to pygments (version 2.19) causes the library to hang indefinitely when processing Lua code that uses spaces for indentation (rather than tabs).

Since [mkdocs](https://github.com/mkdocs/mkdocs) implicitly converts indentation to spaces, anyone using it for Lua documentation is likely to encounter build failures after updating their dependencies. It happened to us at the [MWSE](https://github.com/MWSE/MWSE) repository, where downgrading pygments to 2.18 resolved the failures.

Here is a code snippet from our documentation that will cause [the pygments demo](https://pygments.org/demo/) to hang (when Lua is specified as the language):
```lua
-- This is a working example. You are encouraged to create a new file
-- main.lua, paste this code and see the results in-game!
-- Use `u` key to bribe an actor you are looking at with items from your inventory.


local function bribe()
    local actorReference = tes3.getPlayerTarget()
    -- This function returns nil value if the player isn't aiming at a reference,
    -- so we skip that case
    if not actorReference then return end

    tes3ui.showInventorySelectMenu({
        reference = tes3.player,
        -- .. is Lua operator of concatenation. It joins 2 strings together.
        title = "Bribe " .. actorReference.object.name,
        callback = function(e)
            if e.item then
                -- If e.item exist, that means that the player picked an
                -- item in the  menu. It up to us to do something with it.
                tes3.transferItem({
                    from = tes3.player,
                    to = actorReference,
                    item = e.item,
                    itemData = e.itemData,
                    count = e.count,
                })
                -- Here we calculate the total gold value of the transfered item(s), since that
                -- can be a stack of items. e.count holds the amount of the items selected.
                local itemWorth = e.item.value * e.count

                -- At last! Now the actual persuasion part. We use `modifier` argument.
                -- The higher the value we pass there the higher the disposition change.
                tes3.persuade({
                    actor = actorReference,
                    modifier = math.log10(itemWorth)
                })
            end
        end,
        -- At first it's counter intuitive that this filter selects all the non-enchanted items
        -- This illusion disappears soon as we relize that the game uses this filter in the
        -- enchanting menu to select items viable for enchanting.
        filter = tes3.inventorySelectFilter.enchanted
    })

end

-- We registered our bribe function on keyDown event, and filtered it for `u` key,
-- so our bribe function is only called once `u` key is pressed
event.register(tes3.event.keyDown, bribe, { filter = tes3.scanCode.u })
```

thavelick:
Use possessive quantifier if possible, otherwise fallback to old, buggy regex.

Fixes #2839

For reference, see: 
*  https://www.regular-expressions.info/catastrophic.html 
* https://www.regular-expressions.info/possessive.html

### DIFF
--- a/pygments/lexers/scripting.py
+++ b/pygments/lexers/scripting.py
@@ -57,7 +57,7 @@ class LuaLexer(RegexLexer):
 
     _comment_multiline = r'(?:--\[(?P<level>=*)\[[\w\W]*?\](?P=level)\])'
     _comment_single = r'(?:--.*$)'
-    _space = r'(?:\s+)'
+    _space = r'(?:\s+(?!\s))'
     _s = rf'(?:{_comment_multiline}|{_comment_single}|{_space})'
     _name = r'(?:[^\W\d]\w*)'
 


pygments/lexers/scripting.py
```diff
@@ -57,7 +58,13 @@ class LuaLexer(RegexLexer):
...
     _comment_multiline = r'(?:--\[(?P<level>=*)\[[\w\W]*?\](?P=level)\])'
     _comment_single = r'(?:--.*$)'
-    _space = r'(?:\s+)'
+    if sys.version_info >= (3, 11):
+        # Use a possessive quantifier to prevent greediness from causing runaway backtracking.
+        _space = r'(?:\s++)'
+    else:
+        # Possessive quantifiers are not available in Python < 3.11. Complex Lua is likely to
+        # cause any regex that includes _space to take an exponentially long time.
+        _space = r'(?:\s+)'
```
andersk, line 67: A negative lookahead assertion works in all Python versions.

```python
    _space = r'(?:\s+(?!\s))'
```

thavelick: Nice that makes sense. I fixed it up and it works as far as I can tell. Thanks for helping simplify.

Anteru: Tested, works here - appreciate the investigation!

Anteru:
Queued for Pygments 2.19.2, thanks guys!

PR #2884, opened 2025-04-17
Add Bicep lexer

RobertoPrevato:
Add Lexer for [Bicep](https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/overview?tabs=bicep), for .bicep and [.bicepparam](https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/parameter-files?tabs=Bicep) files.

The Lexer was added to `config.py` together with Terraform, as Bicep belongs to the same category of tools like Terraform.

Please let me know if the PR is valid.


deogratias-saidi:
is it possible to merge this?

PR #2887, opened 2025-05-03
Update MATLAB lexer

watermarkhu:
- Update MATLAB lexer with keywords up to r2024b. 
- Fixed highlighting for argument blocks
- Ruff formatted


watermarkhu:
Favor #3091

PR #2895, opened 2025-05-22
rrt style: add more styles for Generic and Name.Tag

rrthomas:
Primarily aimed at Markdown support.


Anteru:
Merged, thanks again!

PR #2900, opened 2025-05-28
Support headline tags on org-mode headlines with keywords

jbradberry:
e.g. TODO.  The original regexps had left out this combination.

Also, support rendering folded drawers.  This isn't proper Org syntax per se since it is an overlay that Emacs manages for you, but it seems likely that people will want to support syntax highlighting for it (I know I certainly do).

### DIFF
--- a/pygments/lexers/markup.py
+++ b/pygments/lexers/markup.py
@@ -637,18 +637,18 @@ class OrgLexer(RegexLexer):
             (r'^# .*', Comment.Single),
 
             # Headings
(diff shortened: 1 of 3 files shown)


jbradberry:
@Anteru tagging you here.  I think this one is least likely of the 3 I've submitted to need changes.

jbradberry:
@Anteru re-requesting review.

jbradberry:
@Anteru @goshatch pinging again for review.

jbradberry:
@birkenfeld maybe this one as well?

PR #2905, opened 2025-06-02
Update the ObjDump and Gas lexer for ARM/AArch64.

gerph:
The ObjDump lexer was suitable for processing x64 code, but didn't do very well when handling AArch64 (or ARM, for that matter). In GNU binutils' objdump the format differs from that of the LLVM objdump - the former makes the 32bit instruction 8 character hex string, whilst the latter splits it up into bytes. We now recognise these.

Registers weren't recognised in the ARM/AArch64 lexer, as the disassembly only recognised `%...` as a register name. This change introduces the ARM and AArch64 registers as names, and makes them get identified before the identifiers (which they would otherwise look like).

Constants are introduced by a `#<number>` sequence in both, but only the `#<decimal>` was understood; hex constants (eg `#0x100`) were not recognised, and so would be processed as a `#0` number then `x100` identifier.

Undefined instructions are identified with `.inst <value>` in AArch64 and ARM, with the binutils objdump. This is now recognised as a valid format.


gerph:
Failing golden results check; will investigate.

gerph:
Fixed CI - I'd left some bits commented out.

PR #2907, opened 2025-06-03
Add rules for interpolated strings in MoonScript lexer

omnivelociraptor:
This change updates the MoonScript lexer with rules for interpolated strings (eg `"the #{color} car"`).


Anteru:
Merged, thanks!

PR #2910, opened 2025-06-16
More Gleam improvements

aognio:
# Gleam Lexer Improvements (Pygments)

## Additions

- Added support for multiline strings using triple quotes:
  ```gleam
  let x = """
    this is a
    multiline string
  """
  ```

- Added support for bit arrays (`<< ... >>`) as `String.Other`:
  ```gleam
  let bits = <<1, 2, 3>>
  ```

- Added support for discard variables (`_name`) as `Name.Other`.

- Added support for PascalCase type constructors as `Name.Class`:
  ```gleam
  Ok(value)
  Some(result)
  ```

- Added support for function attributes / annotations as `Name.Decorator`:
  ```gleam
  @external("native")
  ```

- Added explicit handling for `pub fn` declarations, grouped by tokens:
  - `Keyword`, `Whitespace`, `Keyword`, `Whitespace`, `Name.Function`

## Modifications

- Renamed `keywords` helper to `_keywords` to avoid naming collisions and clarify intent.

- Reorganized `let` and `fn` rules to:
  - Highlight discard vars in `let`
  - Use `bygroups` for precise token classification

- Enhanced operator and punctuation rules (no changes needed, but retained as-is for completeness and clarity)

## Style & Organization

- Used `words(..., suffix=r'\b')` for keyword matching for robustness
- Maintained consistent token grouping via `bygroups`
- Preserved all original number format rules with no changes

## Ready for Testing

- Compatible with `pygmentize -l gleam` once registered
- Valid for both terminal and HTML highlighting
- Covers most Gleam syntax idioms in use today (v1.0+)

### DIFF
--- a/pygments/lexers/gleam.py
+++ b/pygments/lexers/gleam.py
@@ -17,7 +17,7 @@ __all__ = ['GleamLexer']
 
 class GleamLexer(RegexLexer):
     """
-    Lexer for the Gleam programming language (version 1.0.0).
+    Lexer for the Gleam programming language (version >= 1.0.0).
     """
 
     name = 'Gleam'
@@ -27,7 +27,7 @@ class GleamLexer(RegexLexer):
     mimetypes = ['text/x-gleam']
     version_added = '2.19'
 
-    keywords = words((
+    _keywords = words((
         'as', 'assert', 'auto', 'case', 'const', 'delegate', 'derive', 'echo',
         'else', 'fn', 'if', 'implement', 'import', 'let', 'macro', 'opaque',
         'panic', 'pub', 'test', 'todo', 'type', 'use',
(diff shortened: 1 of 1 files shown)


aognio:
Most of the changes in action here:

Screenshot from 2025-06-15 23-22-01

Gleam sources in the example:

```gleam
// Single-line comment
/// Doc comment for the module

import gleam/io

@external("my_js_target")
pub fn main() {
  let _unused_variable = 123_456
  let name = "Gleam Lexer"

  // Multiline string
  let banner = """
    Welcome to the Lexer Test!
    This is a multiline string.
    """

  // Bit string
  let bits = <<1, 255, "hi"/utf8>>

  // Pattern match
  case name {
    "Gleam" -> io.print("Hello, Gleam!")
    _other -> io.print("Hello, stranger.")
  }

  // Numbers
  let decimal = 42
  let float = 3.1415
  let hex = 0xDEAD_BEEF
  let bin = 0b1010_0101
  let oct = 0o755

  // Function call and pipe
  name
  |> String.upcase
  |> io.print
}

// Function with pub + fn
pub fn compute(value: Int) -> Result(Int, Nil) {
  if value > 0 {
    Ok(value * 2)
  } else {
    Error(Nil)
  }
}

```

PR #2911, opened 2025-06-18
Update lexer for Fennel.

technomancy:
* Add new filename pattern for macro modules.
* Add a mime type (Trac ignores languages that lack this for some reason).
* Update list of special forms for the latest Fennel version.
* Add backtick as a punctuation character.


Anteru:
Merged, thanks!

PR #2913, opened 2025-06-20
Add Yacc parser

el-remph:
Also supports as many GNU Bison extensions as I could find


taminomara:
Can we please get this merged?

PR #2914, opened 2025-06-20
Add Rell lexer

alexj136:
Add a lexer for the [Rell](https://docs.chromia.com/rell/rell-intro) language.


pygments/lexers/rell.py
```diff
@@ -0,0 +1,62 @@
...
+
+    tokens = {
+        'root': [
+            (r'(big_integer|boolean|byte_array|decimal|gtv|integer|json|list|'
+             r'map|mutable|set|text|virtual)\b',
+             Keyword.Type),
+            (r'(false|true|null)\b', Keyword.Constant),
+            (r'(entity|enum|namespace|object|struct)\b', Keyword.Declaration),
+            (r'(function|operation|query)\b', Keyword.Declaration, 'function'),
+            (r'(abstract|and|break|continue|create|delete|else|for|if|import|'
```
Anteru, line 39: Please use `words()` instead of providing the words concatenated manually using `|`.

pygments/lexers/rell.py
```diff
@@ -0,0 +1,62 @@
...
+            (r'//.*?$', Comment.Single),
+            (r'/\*(.|\n|\r)*?\*/', Comment.Multiline),
+            (r'"(\\\\|\\"|[^"])*"', String.Double),
+            (r'\'(\\\\|\\\'|[^\\\'])*\'', String.Single),
+            (r'-?[0-9]*\.[0-9]+([eE][+-][0-9]+)?', Number.Float),
+            (r'-?[0-9]+([eE][+-][0-9]+|[lL])?', Number.Integer),
+            (r'x(\'[a-fA-F0-9]*\'|"[a-fA-F0-9]*")', String.Binary),
+            (r'(\.)([ \n\t\r]*)(' + ident + ')',
+                bygroups(Punctuation, Whitespace, Name.Attribute)),
+            (r'[{}():;,.]', Punctuation),
```
Anteru, line 51: Use `+` here to merge multiple punctuation tokens together.

Anteru (changes requested):

pygments/lexers/rell.py
```diff
@@ -0,0 +1,62 @@
...
+class RellLexer(RegexLexer):
+    """
+    A Lexer for Rell.
+    """
+    name = 'Rell'
+    url = 'https://docs.chromia.com/rell/rell-intro'
+    aliases = ['rell']
+    filenames = ['*.rell']
+    mimetypes = ['text/x-rell']
+    version_added = '2.19.2'
```
Anteru, line 27: `2.20.0`, sorry.

Anteru (changes requested):

alexj136 (commented):
Many thanks for the review @Anteru. I addressed 2 of your 3 comments, just need a bit of clarification on the last item.

pygments/lexers/rell.py
Anteru: Which is probably also wrong and needs updated ... the ask is to use `[{}():;,.]+` (notice the trailing `+` which means that consecutive punctuation tokens will produce a single token, instead of producing multiple tokens of the same type.

alexj136: Ah, that makes sense. It's fixed now. Though I had to move `.` to a separate rule, because the change would otherwise have meant that the `.` in e.g. `1+.99` would be punctuation rather than part of the number.

alexj136:
Thanks again @Anteru, I've addressed your last comment and fixed the `tox check` failure.

Anteru:
Merged, thanks!

RespectTH (approved):

agngnw1-ops (commented):
DDDD

PR #2915, opened 2025-06-21
Terraform: Add support for provider-defined functions

hfern:
Terraform supports calling provider-defined functions via `provider::providername::functionname(...)`. This PR adds Pygment support for that syntax.

The test case is lifted from their example provider function: https://github.com/hashicorp/terraform-provider-time/blob/main/docs/functions/rfc3339_parse.md#example-usage


hfern:
@Anteru any chance I can get a review on this? We are having to work around this in documentation.

PR #2916, opened 2025-06-22
Switch license metadata to the PEP 639 format

SpecLad:
License classifiers and the table form of `license` are now deprecated.


Anteru:
Thanks!

PR #2919, opened 2025-06-25
meson: Add *, /, % operators definition

dnicolodi:


Anteru:
Merged, thanks a lot!

PR #2920, opened 2025-06-29
[lexers] FStab, Interfaces, SSH, and SSHD lexers.

ar1ja:
Added support for:

* `/etc/fstab`
* `/etc/networking/interfaces` (Alpine Linux and alike)
* `/etc/ssh/sshd_config`
* `~/.ssh/sshd_config`

This is useful for sysadmin blog posts, configuration highlighting, etc. Speaking from experience.

mpv-shot0004
mpv-shot0003
mpv-shot0002
mpv-shot0001

Signed-off-by: Arija A. \<ari@ari.lt\>


ar1ja:
Note that the screenshot where `Hostname` is not highlighted was fixed in the case-insensitive keywords fix :)

PR #2929, opened 2025-07-10
fixed protobuf namespace tokenization (namespaces can include .)

pseusys:
In protobuf syntax, namespace definitions are allowed to be separated by "." and actually this feature is used in many generators. It would be nice if it was implemented correctly :)


birkenfeld:
Thanks for the fix!

PR #2930, opened 2025-07-11
FlatBuffers lexer

pseusys:
[FlatBuffers](https://flatbuffers.dev/) is yet another PSLS, kind of similar to protobof and Cap'N Proto, although it might be a little less popular...


pseusys:
Help needed here, I couldn't run `--update-goldens` properly on the sample file... maybe docs are out of date?

Anteru:
It's breaking on other checks - can you take a stab at those first, please? Should be fairly easy to fix I hope. FWIW I did just verify locally that `tox -- --update-goldens` works, so it could be failing due to some test failure for you.

pseusys:
@Anteru, could you please restart the tests? I think I have fixed them.

pseusys:
@Anteru, it would be great if you could restart the tests, so that we understand if my fix works on actions or not :)

PR #2934, opened 2025-07-19
Add lexer for Kakoune

ficcdaf:
This commit adds a lexer for Kakoune's command language. It's implemented as a subclass of the BashLexer due to the very similar syntax. Its output is _very_ similar to Kakoune's highlighting of its own syntax.


ficd0:
This same lexer has been merged into chroma: https://github.com/alecthomas/chroma/pull/1150

I've been using it to highlight code on my website with no issues ([example in blog post](https://ficd.sh/blog/kak-secure-local-config/#hooks-and-commands).

So this PR is ready for review :)

PR #2946, opened 2025-08-11
Fix GetElem in Lean 4

mdbrnowski:
I have added `]'`, `]?`, and `]!` as operators, since they are used as [Data Lookups](https://lean-lang.org/doc/reference/latest/Type-Classes/Basic-Classes/#GetElem___mk).


birkenfeld:
Fun-looking operators, but sure!

PR #2951, opened 2025-08-16
Tera Term macro: Add commands introduced in v5.3 and higher

kurtmckee:
This PR adds support for additional Tera Term macro commands that were introduced in v5.3 of the product [[ref](https://teratermproject.github.io/manual/5/en/macro/command/index.html)].


kurtmckee:
_Rebased on `main` and force-pushed to resolve merge conflicts._

birkenfeld:
Thanks for the update and the patience!

PR #2954, opened 2025-08-22
Reimplement SuperCollider lexer and add tests

### LINKED ISSUE #2953: Proposal: Reimplementation of SuperCollider lexer
## TLDR

The current SuperCollider lexer appears to have several significant limitations, including misclassifications, error tokens on valid input, and missing support for certain literal notations. To improve accuracy and maintainability, I’d like to propose a full reimplementation that A) resolves these issues and B) streamlines the lexer for easier long-term maintenance.

### Misclassification

1. Identifiers like `SynthDef` are incorrectly classified as `Name.Other`. Should be `Name.Class`. In SuperCollider, capitalized identifiers are per definition class names. See https://doc.sccode.org/Reference/Classes.html and https://doc.sccode.org/Reference/Literals.html.
2. `\n  ` is classified as `Text` but should be `Whitespace`. Other combinations of space and newline share this problem. Addressing this would resolve #1905 for SuperCollider.
3. Identifiers like `~tone` are incorrectly classified as two distinct tokens, `Operator` and `Name.Other`. It should be `Name.Variable`, since `~tone` is a variable name scoped to an instance of SuperCollider's `Environment` class. `~` is not an operator in SuperCollider. See https://doc.sccode.org/Tutorials/JITLib/proxyspace_examples.html.
4. Strings like `'hello'` are incorrectly classified as `String.Single`, but should be `String.Symbol`. See https://doc.sccode.org/Reference/Literals.html#Symbols.
5. `inf` is incorrectly classified as a `Keyword.Constant`. SuperCollider's documentation defines it as a float, so we should classify as `Number.Float`. See https://doc.sccode.org/Reference/Literals.html#Floats
6. `<!--` is incorrectly classified as `Comment`. It has no special meaning in SuperCollider.
7. These words are incorrectly classified as `Keyword`, `Name.Builtin`, or `Keyword.Declaration`. They have no special meaning in SuperCollider:
	- in
	- break
	- return
	- continue
	- default
	- else
	- throw
	- catch
	- finally
	- new
	- delete
	- typeof
	- instanceof
	- void
	- Number
	- RegExp
	- Packages
	- isFinite
	- isNan
	- parseFloat
	- parseInt
	- super
	- this
	- let
	- with
	- function
8. These words are classified as `Name.Builtin` but should arguably be tokenized as `Keyword.Builtin.Pseudo`. The SuperCollider documentation describes these as "global pseudo-variables", see e.g. https://doc.sccode.org/Classes/Thread.html#.thisThread.
	- thisFunctionDef
	- thisFunction
	- thisMethod
	- thisProcess
	- thisThread
### Lacking tokenization

1. Char literals such as `$A` should be classified as `String.Char`. See https://doc.sccode.org/Reference/Literals.html#Characters
2. Scale degree accidental notation like `1b2` or `3ss` (flat/sharp notation). See https://doc.sccode.org/Reference/Literals.html#Scale%20Degrees
3. Radix notation like `12r4A.ABC` (base-12 number) should be classified as `Number.Float`. See https://doc.sccode.org/Reference/Literals.html#Radix
4. The keyword `pi` should be classified as `Number.Float`. See https://doc.sccode.org/Reference/Literals.html#Floats
5. Operator handling does not accurately match the operators in SuperCollider, and some operators are missing (`|==|`, `**`, `<!`). Ideally, instead of matching all operators at once, we should first try to match multicharacter operators and then single character ones.
### Errors and other problems

1. `$\\` should be tokenized as `String.Char` (escaped backslash) but produces `Token.Error`.
2. The current lexer assumes that SuperCollider has regex literals like JavaScript's `/pattern/flags`. This causes the lexer to produce a `Token.Error` when parsing nested multiline comments (see below). SuperCollider does not use this type of regex notation, which means we can simplify the state machine logic quite a bit.
3. In SuperCollider, nested multiline comments are allowed, which is not processed correctly by the current lexer. See [https://doc.sccode.org/Reference/Comments.html].
4. The current lexer has a method override on `analyse_text()` which probably should be removed as it does not seem to be doing anything relevant.
5. I don't believe we need the `re.DOTALL` flag.

To produce a few `Token.Error`s as mentioned above:

```python
from pygments.lexers import SuperColliderLexer
from pygments.token import Token
lexer = SuperColliderLexer()

code = """
/* Nested
/* multiline */
comments */

$\\ // Legal syntax
"""

tokens = list(lexer.get_tokens(code))

for token_type, value in tokens:
  print(f"{token_type}: {repr(value)}")
```

Prints this:

```
Token.Text: ''
Token.Comment.Multiline: '/* Nested\n/* multiline */'
Token.Text: '\n'
Token.Name.Other: 'comments'
Token.Text: ' '
Token.Operator: '*'
Token.Text: ''
Token.Error: '/'
Token.Text: '\n'
Token.Text: ''
Token.Text: '\n'
Token.Name.Other: '$'
Token.Error: '\\'
Token.Text: ' '
Token.Comment.Single: '// Legal syntax\n'
```

But should print this:

```
Token.Comment.Multiline: '/*'
Token.Comment.Multiline: ' Nested\n'
Token.Comment.Multiline: '/*'
Token.Comment.Multiline: ' multiline '
Token.Comment.Multiline: '*/'
Token.Comment.Multiline: '\ncomments '
Token.Comment.Multiline: '*/'
Token.Text.Whitespace: '\n\n'
Token.Literal.String.Char: '$\\'
Token.Text.Whitespace: ' '
Token.Comment.Single: '// Legal syntax'
Token.Text.Whitespace: '\n'
```

## Proposal

I have a full reimplementation written from scratch that resolves the issues mentioned above. To show the differences in tokenization, I would be happy to produce examples and showcase the output from the current lexer with and without my proposed changes.

aeskildsen:
1. First commit adds tests to specify expected behaviour.
2. Second commit reimplements the SuperCollider lexer.
3. Third commit updates changelog and authors.
4. Fourth commit updates example file to expect new behaviour.

Fixes #2953

### DIFF
--- a/pygments/lexers/supercollider.py
+++ b/pygments/lexers/supercollider.py
@@ -10,9 +10,9 @@
 
 import re
 
-from pygments.lexer import RegexLexer, include, words, default
-from pygments.token import Text, Comment, Operator, Keyword, Name, String, \
-    Number, Punctuation
+from pygments.lexer import RegexLexer, words, include
+from pygments.token import Whitespace, Comment, Operator, Keyword, Name, String, \
+    Number, Punctuation, Text
 
 __all__ = ['SuperColliderLexer']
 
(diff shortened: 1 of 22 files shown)


pygments/lexers/supercollider.py
```diff
@@ -29,66 +29,126 @@ class SuperColliderLexer(RegexLexer):
...
-        """We're searching for a common function and a unique keyword here."""
-        if 'SinOsc' in text or 'thisFunctionDef' in text:
-            return 0.1
+            (r'\d+', Number.Integer),
+        ],
+        'punctuation': [
+            (r'[{}()\[\],\.;:]+', Punctuation),
+        ],
+        'operators': [
+            (r'[!@%&*\-+=|<>?/]{1,4}', Operator),
```
nhthn, line 122: `{1,4}` should be `+` here, and the comment below is a bit inaccurate. All strings made of the characters `!@%&*\-+=<>?/` are syntactically valid and can be user-defined operators. (Exceptions are strings starting with `//` `/*` and the string `<-`, which isn't a binop.)

pygments/lexers/supercollider.py
```diff
@@ -29,66 +29,126 @@ class SuperColliderLexer(RegexLexer):
...
+
+            # Operators in SuperCollider are somewhat complex.
+            # Rather than trying to match every possible operator manually,
+            # this pattern matches operators based on the character set that is
+            # allowed in binary operators in SuperCollider, as documented here:
+            # https://doc.sccode.org/Classes/Char.html#*binaryOpCharacters
+            # This may match some invalid operators, but that is preferable
+            # to missing valid ones.
+
+            # In addition, there are many word operators in SuperCollider,
```
nhthn, line 132: These are "key binary operators" and can be any alphanumeric string, so they can be matched with the following rule (which should take precedence over the rule matching variable names):

```
(r'[a-z][a-zA-Z0-9_]*:', Name),
```

pygments/lexers/supercollider.py
```diff
@@ -29,66 +29,126 @@ class SuperColliderLexer(RegexLexer):
...
-    }
-
-    def analyse_text(text):
-        """We're searching for a common function and a unique keyword here."""
-        if 'SinOsc' in text or 'thisFunctionDef' in text:
-            return 0.1
+            (r'\d+', Number.Integer),
+        ],
+        'punctuation': [
+            (r'[{}()\[\],\.;:]+', Punctuation),
```
nhthn, line 119: The characters `#^`, the backtick, and the string `<-` should also be parsed as punctuation.

pygments/lexers/supercollider.py
```diff
@@ -29,66 +29,126 @@ class SuperColliderLexer(RegexLexer):
...
-            (r'[0-9][0-9]*\.[0-9]+([eE][0-9]+)?[fd]?', Number.Float),
+        'multiline_comment': [
+            # Nested multiline comment support
+            # https://doc.sccode.org/Reference/Comments.html
+            (r'[^*/]+', Comment.Multiline),
+            (r'/\*', Comment.Multiline, '#push'),
+            (r'\*/', Comment.Multiline, '#pop'),
+            (r'[*/]', Comment.Multiline)
+        ],
+        'keywords': [
```
nhthn, line 62: Missing keywords: `const`, `this`, and `context`.

nhthn (commented):
Thanks for doing this! Few comments from a former SC dev here. I actually had rewritten this lexer locally as well but yours is better.

pygments/lexers/supercollider.py
```diff
@@ -29,66 +29,126 @@ class SuperColliderLexer(RegexLexer):
...
+        'identifiers': [
+            (r'\b[A-Z]\w*\b', Name.Class),
+            (r'~[a-z]\w*\b', Name.Variable),
+        ],
+        'strings': [
+            (r'"', String.Double, 'double_quoted_string'),
+
+            # https://doc.sccode.org/Reference/Literals.html#Symbols
+            (r"'(?:[^']|\\[^fnrtv])*'", String.Symbol),
+            (r'\\\w+\b', String.Symbol),
```
nhthn, line 84: The empty symbol `\\\b` is also valid in sclang, and is missing here.

nhthn:
Sorry, one more thing: make sure to match primitives with the rule `_\w+`.

aeskildsen:
> Thanks for doing this! Few comments from a former SC dev here. I actually had rewritten this lexer locally as well but yours is better.

Thank you for looking this over! Great feedback, I will implement the changes.

pygments/lexers/supercollider.py
aeskildsen: I'm not sure how to categorize `<-`. What is more suitable - Punctuation, Keyword, Text, Name?

pygments/lexers/supercollider.py
aeskildsen: Okay, that makes sense, thank you for explaining. I'm going to leave this as is, then - `gcd` will be matched as a Name and `:` as Punctuation. Since we don't use the colon when invoking `gcd` in method form (`3.gcd(4)`), I think this is a reasonable choice for a syntax highlighter.

pygments/lexers/supercollider.py
aeskildsen: Resolved in commit abe5868

pygments/lexers/supercollider.py
aeskildsen: Resolved in commit abe5868

pygments/lexers/supercollider.py
aeskildsen: Resolved in commit abe5868

pygments/lexers/supercollider.py
aeskildsen: Resolved in commit abe5868 (code comment adjusted)

aeskildsen:
> Sorry, one more thing: make sure to match primitives with the rule `_\w+`.

Should be resolved in commit abe5868 👍

PR #2957, opened 2025-08-27
Update Mathematica lexer

rafal-c:
Add new aliases, extensions, and mime-types. 

These changes are motivated by the fact that the programming language itself is called the [Wolfram Language](https://www.wolfram.com/language/), hence the new aliases `wolfram` and `wl`. I also added the `*.wls` extension which is used for Wolfram Language scripts.


birkenfeld:
Looks good to me, thanks!

PR #2966, opened 2025-09-15
uz and zu Integer Literal Suffixes in C and C++

insertinterestingnamehere:
In C++23, the `uz` literal suffix was added to allow making `size_t` typed integers. See [P0330R8](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p0330r8.html). There's a similar change proposed but not yet accepted for the C standard. See [N3485](https://thephd.dev/_vendor/future_cxx/papers/C%20-%20Literal%20Suffixes%20for%20size_t.html). This updates the lexer here to recognize that suffix for integer literals.


birkenfeld:
LGTM.

PR #2969, opened 2025-09-24
Fix #2926 issue in BashLexer

### LINKED ISSUE #2926: Coloured keywords in bash lexer at beginning of a longer name
Hi,

if I have following bash code (script.sh):
```bash
make -f Makefile
do-release-upgrate
lets-do-it
if-else-script
do
    true;
done
until [ false ]
```
with `pygmentize -f html -O style=colorful,linenos=1 -l bash script.sh` I receive:

Image

 where `do` and `if` in liens 2 and 4 are coloured while they sgould not be.

styx0x6:
This PR improves the Bash lexer by fixing an issue where keywords at the beginning of certain names were incorrectly highlighted, and updates the Bash documentation link. It also adds a new test file `doif.sh` and its expected output to verify the fix.

Lexer improvements:

* Fixed the Bash lexer so that keywords (like `do`, `if`, `until`) are not incorrectly highlighted when they appear at the start of variable or command names containing special characters (e.g., `do-release-upgrade`, `do@arobase`). It just excludes some problematic chars as other attempt to do it better will be more heavy changes on logic with high risj of side effects.
* Updated the Bash lexer documentation URL to point to the official GNU Bash manual.

Fixes pygments/pygments#2926


styx0x6:
Hi @birkenfeld @Anteru ,

Any chance to merge this one as it fixes a small colorisation issue in the Bash lexer?

birkenfeld:
Hm, so now we have two assertions there - first your new lookahead and then the `\b` after optional space. I think they should be combined.

styx0x6:
Hi @birkenfeld ,

You are right, I commited an update. I let you check and merge.

birkenfeld:
👍

PR #2970, opened 2025-09-24
Update the mysql lexer keywords

kurtmckee:
This PR introduces the following changes:

* The `CHAR BYTE` alias is now supported.

  Previously, `BYTE` had not been understood to be part of a type alias and was registered only as a reserved keyword.

* Many additional keywords and reserved keywords have been found by changing the mysql GitHub branch from `8.0` to `trunk`.

  A number of existing keywords [have been removed in the mysql keyword list](https://dev.mysql.com/doc/refman/8.4/en/keywords.html#keywords-removed-in-current-series) since mysql 8.0, but for backwards compatibility these removals were discarded in the pygments keywords list.

  The result is a number of additions to the list of mysql keywords that pygments recognizes, with zero removals. _Note that `byte` appears to be a removal, but is actually moved up to the tuple of `MYSQL_DATATYPES`, as noted in the first bullet point above._


birkenfeld:
LGTM, thanks!

PR #2973, opened 2025-09-26
Template string support in Python formatting

bartbroere:
[Python 3.14 adds support for template strings](https://docs.python.org/3.14/library/string.templatelib.html#template-strings). The syntax of t-strings is the same as that of f-strings, so it's relatively easy to format it properly.

I didn't rename the name of the state, nor create different states for f-strings and template strings. I could do that if it's preferred.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -11,6 +11,10 @@ Version 2.20.0
 
   * Rell (#2914)
 
+- Updated lexers:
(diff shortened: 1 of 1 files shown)


bartbroere:
This PR is no longer necessary, also implemented in #3009

PR #2977, opened 2025-10-12
Add generic styles for onedark

Yutsuten:
I use one dark theme myself, but I noticed that there wasn't styling for `.diff` files.
After a bit of a research, I found that some missing entries in the `styles/onedark.py` was the issue.

I filled the missing entries with what makes sense to me, while comparing with what I get with `git diff`.


birkenfeld:
Sure! Thanks for the PR.

PR #2980, opened 2025-10-23
Turtle: Allow "empty" comment lines

zyzzyxdonta:
Comments may consist of only a `#` symbol immediately followed by a newline character. Comments like this are likely to appear for formatting purposes in larger blocks of comments.


zyzzyxdonta:
Relevant section in the spec, for reference: https://www.w3.org/TR/turtle/#sec-grammar-comments

birkenfeld:
Thanks for the PR!

PR #2987, opened 2025-11-09
Add support for Python 3.14 and drop EOL 3.8-3.9

hugovk:
* 3.14 released in October 2025
* 3.8 was EOL in October 2024
* 3.9 was EOL in October 2025

https://devguide.python.org/versions/


Anteru:
RHEL still supports Python 3.9, so we can't drop that. Dropping 3.8 though is fine, and I've added 3.14. Thanks!

PR #2997, opened 2025-12-03
Recognise markdown block tokens on last line of text

joouha:
Currently the markdown lexer will fail to recognise block tokens on the final line of text where there is no trailing new-line.

This can be fixed by matching either a new-line or the end-of-string after a block token.

### DIFF
--- a/pygments/lexers/markup.py
+++ b/pygments/lexers/markup.py
@@ -541,26 +541,26 @@ class MarkdownLexer(RegexLexer):
     tokens = {
         'root': [
             # heading with '#' prefix (atx-style)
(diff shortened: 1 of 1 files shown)


birkenfeld:
You should probably not set the `ensurenl=False` option for the lexer, then this should not be necessary.

joouha:
Hi - thanks. I think the issue I'm encountering actually comes from `prompt_toolkit`'s use of pygments, where lines of text get split and joined prior to lexing, which results in stripping of final newlines even if `ensurenl=True`. So I'll close this.

PR #2999, opened 2025-12-11
Update shell.py

7heo:
Reflect the fact that this is a "Lexer for various shells" as described in the top comment, and not just a Lexer for Bash.

Bash is not just shell, it is definitely a more advanced language, with array support, better loops, the `function` keyword, the `select` command, file slurping, more expansions, more conditionals, etc.

Assuming that any shell script would be a "Bash script" is error inducing at best, novices might try, not knowing better, to introduce so called bashisms in a POSIX shell script, and even seasoned developers might be mislead and assume the wrong thing.

Calling it a "Shell" script costs nothing, is correct, and will prevent any such issues.

Note: I left the references to "Bash" in function names and such, as the functionality is independent of those, but to be perfectly correct, those should also be corrected.


birkenfeld:
Not a bad idea, but the new name is a bit too generic, since e.g. tcsh with its own lexer later in the file is also a similar shell :)

7heo:
> Not a bad idea, but the new name is a bit too generic, since e.g. tcsh with its own lexer later in the file is also a similar shell :)

I honestly thought it was appropriate to make the general case generic, and any specific case specific. Would you like a specific case for Bash, or what do you mean? I can change it, I just want to be sure of what I should do.

Edit: if you want to make a specific case for bash, it would even make more sense to make one for zsh.

PR #3002, opened 2025-12-20
php: Add new reserved keywords

supertassu:
`readonly` and `yield from` have been added as language keywords in PHP 8.1 or later. Update the lexer accordingly.


birkenfeld:
Thanks!

Issue #2852, opened 2025-02-01
[LESS] [SCSS/SASS] Add support for single‑line comments


ExE-Boss:
**LESS** and **SCSS**/**SASS**, unlike **CSS**, all allow single‑line comments, but **Pygments** fails to highlight those as a comment:

<dl>
<dt>LESS</dt>
<dd>

```less
// single-line comment
```
</dd>
<dt>SCSS</dt>
<dd>

```scss
// single-line comment
```
</dd>
<dt>SASS</dt>
<dd>

```sass
// single-line comment
```
</dd>
</dl>

PR #3005, opened 2025-12-24
fix(lexers.less): Fix support for single‑line comments in LESS

ExE-Boss:
This fixes the **LESS** part of <https://github.com/pygments/pygments/issues/2852>.

--------------------------------------------------------------------------------

Unfortunately, the single‑line comment definitions have to be copied in every location instead of defining them once in `basics`, as `include('basics')` doesn’t take overrides into account.


Anteru:
Merged, thanks!

PR #3008, opened 2026-01-06
Add support for the Papyrus scripting language

TranslationLookasideBuffer:
[Papyrus](https://ck.uesp.net/wiki/Category:Papyrus) is the scripting language built for the Creation Engine by Bethesda Softworks and is used in several of their games, notably _The Elder Scrolls V: Skyrim_, _Fallout 4_, and _Starfield_. Each game shipped with a different iteration of the language, so this change adds support for the first of those, _Skyrim_.

I agree to this code being released under the BSD 2-clause license.

### DIFF
--- a/pygments/lexers/papyrus.py
+++ b/pygments/lexers/papyrus.py
@@ -214,3 +214,17 @@ class PapyrusSkyrimLexer(RegexLexer):
             (r'[^"\\]+', String.Double),
         ],
     }
+
+    def analyse_text(text):
+        score = 0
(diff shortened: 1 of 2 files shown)


Anteru:
Thanks for the contribution. This is missing an `analyse_text` function and as a result fails the CI checks. Would you mind adding that? It doesn't have to be super accurate.

TranslationLookasideBuffer:
Alright, should be good to go!

PR #3010, opened 2026-01-09
Python lexer: support t-strings

### LINKED ISSUE #3009: Support Python 3.14's t-strings
I think we can reuse f-string support for t-strings, their syntax is pretty-much identical.

taminomara:
Fix #3009.

I didn't duplicate all tests for f-strings because f- and t-strings use the same rules.


Anteru:
Merged, thank you!

PR #3012, opened 2026-01-11
Add Python 3.14 to Trove classifiers

hugovk:
Follow on from https://github.com/pygments/pygments/pull/2987 and https://github.com/pygments/pygments/commit/cfca37efebd16fa3e8fce60bd5b9650c8587e4d9.


Anteru:
Good catch, thanks.

Issue #2992, opened 2025-11-18
test_lexer_classes fails with pytest 8.4


hrnciar:
Hello,

I am working on updating pytest in Fedora and pygments is one of the affected packages that fails to build. The `test_lexer_classes` is failing with:

~~~~
=================================== FAILURES ===================================
_____________________ test_lexer_classes[BashSessionLexer] _____________________

cls = <class 'pygments.lexers.shell.BashSessionLexer'>

    @pytest.mark.parametrize('cls', lexers._iter_lexerclasses(plugins=False))
    def test_lexer_classes(cls):
        # test that every lexer class has the correct public API
        assert isinstance(cls.name, str)
        for attr in 'aliases', 'filenames', 'alias_filenames', 'mimetypes':
            assert hasattr(cls, attr)
            assert isinstance(getattr(cls, attr), list), \
                f"{cls}: {attr} attribute wrong"
        assert isinstance(cls.url, str), \
            (f"Lexer class {cls.__name__} is missing the `url` attribute. "
             "Please add it to provide a link to the language's homepage "
             "for the Pygments documentation (or set it to an empty "
             "string if this doesn't make sense for the lexer).")
        assert isinstance(cls.version_added, str), \
            (f"Lexer class {cls.__name__} is missing the `version_added` attribute. "
             f"Please add `version_added = '{major}.{minor+1}'` to the class definition.")
        if cls.version_added:
            assert re.fullmatch(r"\d+\.\d+", cls.version_added), \
                (f"Lexer class {cls.__name__} has a wrong version_added attribute. "
                 "It should be a version number like <major>.<minor> (but not "
                 "<major>.<minor>.<micro>).")
        if cls._example is not None:
            assert isinstance(cls._example, str)
            p = (
                pathlib.Path(inspect.getabsfile(pygments)).parent.parent
                / "tests"
                / "examplefiles"
                / cls._example
            )
>           assert p.is_file(), f"Example file {p} not found"
E           AssertionError: Example file /builddir/build/BUILD/python-pygments-2.19.1-build/BUILDROOT/usr/lib/python3.14/site-packages/tests/examplefiles/console/example.sh-session not found
E           assert False
E            +  where False = is_file()
E            +    where is_file = PosixPath('/builddir/build/BUILD/python-pygments-2.19.1-build/BUILDROOT/usr/lib/python3.14/site-packages/tests/examplefiles/console/example.sh-session').is_file

tests/test_basic_api.py:71: AssertionError
[...]

PR #3016, opened 2026-01-13
Fix test_lexer_classes search path

yselkowitz:
"Test the existence of the example file" (commit afd62f5e6ac, #2712) caused
the test to search for the example files relative to the import location of
pygments rather than relative to the tests.  In normal in-tree usage there
may be no practical difference, but in downstream packaging there may be.

Resolves: #2992


birkenfeld:
LGTM, thanks.

PR #3021, opened 2026-01-22
Add support for overlay/fragments in DevicetreeLexer and update example file

kartben:
The DTS lexer had issues with recognizing overlay/fragment syntax, whereby in the following snippet i2c1 wouldn't be seen as a "node"

```devicetree
&i2c1 {
	serlcd@72 {
		compatible = "sparkfun,serlcd";
		reg = <0x72>;
		columns = <16>;
		rows = <2>;
		command-delay-ms = <10>;
		special-command-delay-ms = <50>;
	};
};
```

and "i2c1" was being tokenized as a Name.Attribute instead of a Name.Function, causing the rest of the lexing to fail as it encountered the "@" character in an unexpected location.

This fixes the issue by adding proper support for overlay/fragment syntax, both `&label` form as well as `&{node/full/path}`.

Note that as can be seen from the diff this doesn't change anything else in the tokenization, so there's that :)

Before:

image

After:

image

### DIFF
--- a/pygments/lexers/devicetree.py
+++ b/pygments/lexers/devicetree.py
@@ -71,7 +71,10 @@ class DevicetreeLexer(RegexLexer):
         'root': [
             include('whitespace'),
             include('macro'),
-
+            # Overlay/fragment: &label { ... } or &{/path/to/node} { ... }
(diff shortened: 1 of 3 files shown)


birkenfeld:
Thanks, LGTM!

PR #3022, opened 2026-01-22
Add bytestring support to DevicetreeLexer and update example DTS file

kartben:
Add support for bytestring which, as per Devicetree Specification:

-  A bytestring is enclosed in square brackets [ ] with each byte represented by two hexadecimal digits. Spaces between each byte are optional. Example:

        local-mac-address = [00 00 12 34 56 78];

   or equivalently:

        local-mac-address = [000012345678];


birkenfeld:
LGTM. Thanks!

PR #3023, opened 2026-01-23
LLVM: Support C-style comments

### LINKED ISSUE #2978: LLVM: No highlighting for block comments
A fairly new feature in LLVM IR is C-style block comments, like `/* ... */`.

See https://llvm.org/docs/LangRef.html#identifiers

> Comments are delimited with a ‘;’ and go until the end of line. Alternatively, comments can start with /* and terminate with */.

These don't get highlighted by Pygments yet, which results in error highlighting within the LLVM docs:

Image

kurtmckee:
Fixes #2978


birkenfeld:
:+1:

PR #3031, opened 2026-02-09
Fix analyse_text: use re.search instead of re.match in HaxeLexer and ActionScript3Lexer

### LINKED ISSUE #3030: HaxeLexer.analyse_text and ActionScript3.analyse_text re.match bug
```
    def analyse_text(text):
            return 0.3
```

daniel7an:
## Summary

`HaxeLexer.analyse_text` and `ActionScript3Lexer.analyse_text` use `re.match`, which only checks at the beginning of the string. This means the heuristic fails for files that start with comments, package imports, or any non-matching content.

This PR changes both to `re.search`, which scans the entire text — matching the intended behavior.

Fixes #3030

### DIFF
--- a/pygments/lexers/actionscript.py
+++ b/pygments/lexers/actionscript.py
@@ -195,7 +195,7 @@ class ActionScript3Lexer(RegexLexer):
     }
 
     def analyse_text(text):
-        if re.match(r'\w+\s*:\s*\w', text):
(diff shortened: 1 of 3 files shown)


birkenfeld:
Duplicate AI slop, not even marked as such.

PR #3035, opened 2026-02-10
Fix re.match vs re.search bug in HaxeLexer and ActionScript3Lexer analyse_text

### LINKED ISSUE #3030: HaxeLexer.analyse_text and ActionScript3.analyse_text re.match bug

veeceey:
## Summary

- Fix `analyse_text` in `HaxeLexer` and `ActionScript3Lexer` which use `re.match` instead of `re.search` to look for type annotation patterns (`\w+\s*:\s*\w`). Since `re.match` only checks at the beginning of the string and source files typically start with comments, package declarations, or imports, the pattern almost never matches, making `analyse_text` effectively non-functional for language detection.
- Lower the priority score from 0.3 to 0.1 because this pattern is very generic and matches many languages (Python dict comprehensions, Carbon type annotations, etc.), avoiding false positives when guessing the lexer.

Fixes #3030

## Changes

- `pygments/lexers/haxe.py`: `re.match` -> `re.search`, score `0.3` -> `0.1`
- `pygments/lexers/actionscript.py`: `re.match` -> `re.search`, score `0.3` -> `0.1`

## Test plan

- [x] All 678 tests in `test_guess.py` pass (no regressions)
- [x] All 34 Haxe/ActionScript-related tests pass
- [x] `test_guess_carbon_lexer` still correctly identifies Carbon code (CarbonLexer returns 0.2, which beats HaxeLexer's 0.1)
- [x] HaxeLexer and AS3Lexer now correctly return 0.1 for typical source files that start with comments/imports

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pygments/lexers/actionscript.py
+++ b/pygments/lexers/actionscript.py
@@ -195,7 +195,7 @@ class ActionScript3Lexer(RegexLexer):
     }
 
     def analyse_text(text):
-        if re.search(r'\w+\s*:\s*\w', text):
+        if re.search(r'import\s+flash\.', text):
             return 0.1
         return 0
 
--- a/pygments/lexers/haxe.py
+++ b/pygments/lexers/haxe.py
@@ -890,7 +890,7 @@ class HaxeLexer(ExtendedRegexLexer):
     }
 
     def analyse_text(text):
-        if re.search(r'\w+\s*:\s*\w', text):
+        if re.search(r':\s*(Int|Float|Bool|Void|Dynamic)\b', text):
             return 0.1
 
 


veeceey:
Friendly ping - any chance someone could take a look at this when they get a chance? Happy to make any changes if needed.

birkenfeld:
Looking at those patterns, they're way too generic to be useful. If you want to do something here, please find a better, more unique pattern for those languages. Otherwise I'm going to remove those two.

veeceey:
Thanks for the feedback @birkenfeld. You're right that `\w+\s*:\s*\w` is too generic even at 0.1 — it matches way too many languages.

I'll update the PR with more specific patterns:

- **HaxeLexer**: Look for Haxe-specific keywords like `package`, `import`, `class` combined with Haxe-style type annotations (e.g., `var x:Int`, `function foo():Void`), or Haxe-specific types like `haxe.`, `Std.`, `StringTools`
- **ActionScript3Lexer**: Look for AS3-specific patterns like `package {`, `import flash.`, `import mx.`, or `[Bindable]`/`[Embed]` metadata tags

If you'd prefer to just remove both `analyse_text` methods entirely, I'm fine with that too — just let me know which approach you'd prefer.

birkenfeld:
Detecting Haxe-specific types sounds good (the other are still too generic), and `import flash.` for Actionscript.

veeceey:
Updated — HaxeLexer now checks for Haxe-specific types (`Int`, `Float`, `Bool`, `Void`, `Dynamic`) and ActionScript3Lexer uses `import flash.` for detection. Much more targeted than the old generic pattern.

birkenfeld:
Well, #3038 was better. Sorry Claude.

PR #3038, opened 2026-02-17
Fix analyse_text using re.match instead of re.search

### LINKED ISSUE #3030: HaxeLexer.analyse_text and ActionScript3.analyse_text re.match bug

worksbyfriday:
## Summary
- Fixes #3030
- `HaxeLexer.analyse_text` and `ActionScript3Lexer.analyse_text` used `re.match` which only matches at the start of the string
- Haxe and ActionScript3 files typically start with comments or package imports, so the type annotation pattern `\w+\s*:\s*\w` was never found
- Changed to `re.search` and lowered the score from 0.3 to 0.1 since this pattern is common across many languages

## Test plan
- [x] All 678 guess tests pass
- [x] All 16 Haxe/ActionScript-specific tests pass
- [x] The `test_guess_carbon_lexer` test (mentioned in #3030 as a potential conflict) continues to pass since the Haxe score is now 0.1 (below Carbon's 0.2)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pygments/lexers/actionscript.py
+++ b/pygments/lexers/actionscript.py
@@ -195,8 +195,8 @@ class ActionScript3Lexer(RegexLexer):
     }
 
     def analyse_text(text):
-        if re.search(r'\w+\s*:\s*\w', text):
-            return 0.1
+        if re.search(r'\bimport\s+flash\.', text):
+            return 0.2
         return 0
 
 
--- a/pygments/lexers/haxe.py
+++ b/pygments/lexers/haxe.py
@@ -890,8 +890,14 @@ class HaxeLexer(ExtendedRegexLexer):
     }
 
     def analyse_text(text):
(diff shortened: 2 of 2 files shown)


birkenfeld:
Well Claude, looks like you're very active here in #3035 and this. If you want to compete, update your PR to fix the issue the same way I advised you in #3035.

worksbyfriday:
Updated to use language-specific patterns per your advice in #3035:

- **HaxeLexer**: detects `import haxe.`, `Std.`, and `trace()`
- **ActionScript3Lexer**: detects `import flash.`

These replace the generic `\w+\s*:\s*\w` pattern that could match many languages.

birkenfeld:
LGTM. Good bot

PR #3040, opened 2026-02-17
fix(regexopt): Fix os.path.commonprefix deprecation warning in python3.15

lambda-abstraction:
python/cpython#144436 deprecates `os.path.commonprefix` in 3.15

i copied the implementation of `commonprefix` (https://github.com/python/cpython/blob/957f9fe162398fceeaa9ddba8b40046b8a03176d/Lib/genericpath.py#L115-L129) and removed the unused `fspath`, as its only used with strings, not paths

### DIFF
--- a/pygments/regexopt.py
+++ b/pygments/regexopt.py
@@ -11,12 +11,23 @@
 
 import re
 from re import escape
-from os.path import commonprefix
 from itertools import groupby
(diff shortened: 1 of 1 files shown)


birkenfeld:
LGTM, thanks!

PR #3043, opened 2026-02-20
Fix xterm-256 color table: wrong cube count and missing grayscale entries

bysiber:
## Problem

The xterm-256 color table in `Terminal256Formatter` has two off-by-one errors:

1. **Color cube**: `range(217)` generates 217 entries instead of 216 (6x6x6 = 216). The extra entry at index 232 contains `(0,0,0)` (black) instead of the correct grayscale start.

2. **Grayscale ramp**: `range(1, 22)` generates only 21 entries instead of the correct 24 (colors 232-255). Three grayscale values are missing entirely.

Total entries: 16 + 217 + 21 = 254 instead of the correct 256. The extra cube entry shifts all grayscale indices by one, and the missing entries cause wrong color matching for grays.

Additionally, `_closest_color` iterates `range(0, 254)` which doesn't cover all entries.

## Fix

- Color cube: `range(216)`
- Grayscale: `range(24)`
- Closest color search: `range(256)`

### DIFF
--- a/pygments/formatters/terminal256.py
+++ b/pygments/formatters/terminal256.py
@@ -169,19 +169,19 @@ class Terminal256Formatter(Formatter):
         self.xterm_colors.append((0x00, 0xff, 0xff))  # 14
         self.xterm_colors.append((0xff, 0xff, 0xff))  # 15
 
-        # colors 16..232: the 6x6x6 color cube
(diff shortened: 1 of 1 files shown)


birkenfeld:
LGTM, thanks!

PR #3045, opened 2026-02-20
Add HAProxy configuration file lexer

gaborbernat:
HAProxy is one of the most widely used open-source reverse proxies and load balancers. The new `HAProxyLexer` (alias `haproxy`, matching `haproxy.cfg`) covers the full configuration syntax: section headers (`global`, `frontend`, `backend`, `listen`, `defaults`, `resolvers`, `peers`, etc.), directive keywords, preprocessor conditionals (`.if`/`.elif`/`.else`/`.endif`), inline ACL expressions (`{ ... }`), quoted strings with escape sequences, numeric values with time and size suffixes (`30s`, `10m`, `4k`), environment variable expansion (`${VAR}`), IP address literals, and ACL flags (`-i`, `-m`). An `analyse_text` heuristic is included so configs can be auto-detected without requiring an explicit lexer name.

The lexer was built against the official HAProxy configuration documentation:
- [Configuration Manual](https://docs.haproxy.org/3.1/configuration.html)
- [Management Guide](https://docs.haproxy.org/3.1/management.html)
- [HAProxy Starter Guide](https://docs.haproxy.org/3.1/intro.html)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -9,6 +9,7 @@ Version 2.20.0
 
 - New lexers:
 
+  * HAProxy
   * Rell (#2914)
 
 - Updated lexers:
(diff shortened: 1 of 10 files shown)


RealOrangeOne:
Whilst it's pretty old, you may also be interested in https://github.com/haproxytech/haproxy-pygments-lexer

PR #3047, opened 2026-02-22
Fix catastrophic backtracking in Lua/Luau lexer

### LINKED ISSUE #3036: Lua lexer bug
I've experienced Pygments "hanging" when processing a Lua file, I've narrowed it down to the following synthetic example:

```
function outer()
  for i = 1, 10 do
    for j = 1, 5 do
      local node3 = {}
      for k = 1, 3 do
            local item4 = collection.items[ref4]
            local item4data = extractFromItem(document, item4)
            local refs5 = item4.itemRefs or {}
            node4.children = {}
            -- has no children but sequences
            node4.sequences = {}
            node4.name = item4data.name or "unnamed node"
            node4.item_refs = refs5
            manager.rootgroups[#manager.rootgroups + 1] = node4
            manager.nodes[#manager.nodes + 1] = node4
            node3.children[#node3.children + 1] = node4
            -- print(("Level4 Item %s:" ):format(ref4))
            -- util.dumpTable(data.items)
          end
      end
  end
end
```

The stack trace pointed me at this line taking forever https://github.com/pygments/pygments/blob/master/pygments/lexer.py#L712 and the regular expression causing this seems to be:

```
[A-Za-z_]\w*(?=(?:(?:--\[(?P<level>=*)\[[\w\W]*?\](?P=level)\])|(?:--.*$)|(?:\s+))*[.:])
```

Edit: after half an hour, it progressed to the next expression:

```
[A-Za-z_]\w*(?=(?:(?:--\[(?P<level>=*)\[[\w\W]*?\](?P=level)\])|(?:--.*$)|(?:\s+))*\()
```

Let me know if I can provide more details.

worksbyfriday:
## Summary

Fixes #3036 — the Lua and Luau lexers hang indefinitely on certain inputs due to catastrophic regex backtracking.

**Root cause**: The `_s` pattern matches whitespace, single-line comments, and multi-line block comments. Its `_comment_multiline` component contains `[\w\W]*?`. When `_s` is used in lookaheads with a `*` quantifier — e.g. `(?={_s}*[.:])` — the regex engine tries all possible partitions of the input between `[\w\W]*?` and the outer `*`, causing exponential backtracking.

**Fix**: Introduce `_s_la` (lookahead-safe) that matches only whitespace (`\s`), avoiding the ambiguous alternation entirely. This changes tokenization only for the rare case of multi-line block comments between an identifier and its following `.`/`:`/`(` — those identifiers are now classified as plain `Name.Variable` instead of being tracked through the varname/funcname state.

Changes:
- Added `_s_la = r'\s'` to both `LuaLexer` and `LuauLexer`
- Replaced `_s` with `_s_la` in all lookahead patterns (6 locations in LuaLexer, 2 in LuauLexer, 1 in `_luau_make_expression`)
- Updated golden test outputs for the minor reclassification
- Added regression test snippet (`tests/snippets/lua/test_no_hang_comments.txt`)

**Before**: The reproduction case from #3036 hangs indefinitely
**After**: Tokenizes in <1ms (201 tokens)

### DIFF
--- a/pygments/lexers/scripting.py
+++ b/pygments/lexers/scripting.py
@@ -59,6 +59,12 @@ class LuaLexer(RegexLexer):
     _comment_single = r'(?:--.*$)'
     _space = r'(?:\s+(?!\s))'
     _s = rf'(?:{_comment_multiline}|{_comment_single}|{_space})'
+    # A lookahead-safe version of _s that avoids catastrophic backtracking.
+    # The _comment_multiline pattern contains [\w\W]*? which, when used
+    # inside a lookahead with a * quantifier, causes exponential blowup.
+    # This version skips only whitespace; comments between an identifier
+    # and a following [.:] or ( are rare enough to sacrifice.
+    _s_la = r'\s'
     _name = r'(?:[^\W\d]\w*)'
 
     tokens = {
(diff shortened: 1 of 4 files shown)


birkenfeld:
Looks good to me! Thanks for the fix.

PR #3052, opened 2026-02-28
Improve architecture parsing in 'Debian control' lexer

sblondon:
The dependancy fields accept specification of architectures and forbidden architectures. For examples:
- `Depends: foo [i386], bar [amd64]`
- `Build-Depends: foo [!i386] | bar [!amd64]`

The regexp modification splits the '[]' characters, the optional '!' and the architecture names.

The examples are copied from Debian Policy documentation: https://www.debian.org/doc/debian-policy/ch-relationships.html


birkenfeld:
Sure, LGTM!

PR #3057, opened 2026-03-06
lexers: devicetree: Fix catastrophic backtracking bug

grahamroff-dev:
The regex for parsing devicetree statements contains a property name lookahead that results in a near infinite loop if there are a lot of whitespace characters in the property value.

Problem can be seen by running this example script:

```python
from pygments.lexers import DevicetreeLexer
from pygments.formatters import NullFormatter

def test():
    dts_contents = '''
/ {
	soc {
		plic: interrupt-controller@c000000 {
			interrupts-extended = < &hlic0
			                        0x9 >;
		};
    }
}
    '''
    print('Starting highlight')
    highlight(dts_contents, DevicetreeLexer(), NullFormatter())

if __name__ == "__main__":
    test()
```

The script will appear to hang forever (technically it should eventually finish, maybe in a day or so).

**Root Cause: Catastrophic Backtracking**

The hang is caused by catastrophic backtracking in the property name lookahead regex inside `lexers/devicetree.py`.

The problematic pattern in the `statements` token rule is:
```python
(r'[a-zA-Z_][\w-]*(?=(?:\s*,\s*[a-zA-Z_][\w-]*|(?:' + _ws + r'))*\s*[=;])',
 Name),
```

The second alternative inside the outer `*` is `\s*(?:/[*][^*/]*?[*]/\s*)*` (the `_ws` pattern), which **can match an empty string**. This creates the classic catastrophic backtracking pattern `(A|B)*` where `B` can match empty.

**More details**: In the example the lookahead is applied to `\n\t\t\t                        0x9 >;` (~28 whitespace characters before `0x9`). Since `0x9 >` is not `=` or `;`, the lookahead must fail — but the regex engine tries to split those 28 whitespace characters among the outer `*` iterations in **2²⁸ ≈ 268 million** different ways before giving up. When `0x9` is on the same line without the extra indentation, the whitespace is just one space (2¹ = 2 attempts), so it completes instantly.

**Fix applied**: restructured the lookahead so the outer `*` can only match when a `,` is present (it can never match empty string):
```python
(r'[a-zA-Z_][\w-]*(?=(?:\s*,' + _ws + r'[a-zA-Z_][\w-]*)*' + _ws + r'[=;])',
 Name),
```

This correctly handles all cases (comma-separated property names, comments before `=`/`;`) while eliminating the exponential backtracking.

### DIFF
--- a/tests/examplefiles/devicetree/example.dts
+++ b/tests/examplefiles/devicetree/example.dts
@@ -205,5 +205,10 @@
 			reg = < 0x200000 0x1000 >;
 			qcom,start-offset = < 0x90010 >;
 		};
+
+		interrupt-controller@c000000 {
+			interrupts-extended = < &gic 0x1 &gic 0x2 &gic
+			                        0x3 >;
+		};
 	};
 };
--- a/tests/examplefiles/devicetree/example.dts.output
+++ b/tests/examplefiles/devicetree/example.dts.output
@@ -1602,6 +1602,49 @@
 ';'           Punctuation
 '\n'          Text.Whitespace
 
+'\n'          Text.Whitespace
+
+'\t\t'        Text.Whitespace
+'interrupt-controller' Name.Function
+'@'           Operator
+'c000000'     Literal.Number.Integer
+' '           Comment.Multiline
+'{'           Punctuation
+'\n'          Text.Whitespace
+
+'\t\t\t'      Text.Whitespace
+'interrupts-extended' Name
+' '           Text.Whitespace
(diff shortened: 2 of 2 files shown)


birkenfeld:
Thanks for the PR! Can you add the example to the test cases?

grahamroff-dev:
> Thanks for the PR! Can you add the example to the test cases?

I updated the test case to include the problematic syntax, thanks!

birkenfeld:
Great! :)

PR #3069, opened 2026-03-26
Fix Haskell lexer: handle escape sequences in character literals

### LINKED ISSUE #1795: Haskell lexer doesn't work for '\n'
The character literal '\n' (same for '\x', for other x that I've tried) appears to be handled incorrectly. Using the LaTeX minted package with BW style, which is built on Pygments, it puts '\ in boldface and then n' in normal (not-bold) face.

This appears to be related to https://github.com/pygments/pygments/issues/71.

mvanhorn:
## Summary

Fixes incorrect tokenization of Haskell escape character literals like `'\n'`, `'\t'`, `'\\'`.

## Why this matters

The root-state regex `'[^\\]'` only matched character literals containing a single non-backslash character. Escape sequences like `'\n'` were split across tokens - `'\` became `Keyword.Type` and `n'` became `Name` - producing wrong highlighting.

## Changes

`pygments/lexers/haskell.py` line 57: Added pattern `'\\.'` to match escape character literals. Placed after the existing non-escape pattern so simple chars like `'a'` still match first.

## Testing

Verified all common cases tokenize as `Token.Literal.String.Char`:
- `'\n'` (newline), `'\t'` (tab), `'\\'` (backslash), `'a'` (simple), `'A'` (uppercase)

Fixes #1795

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/tests/examplefiles/haskell/example.hs.output
+++ b/tests/examplefiles/haskell/example.hs.output
@@ -194,9 +194,7 @@
 "ESC'"        Keyword.Type
 ','           Punctuation
 ' '           Text.Whitespace
-"'\\"         Keyword.Type
-"'"           Literal.String.Char
-"'"           Literal.String.Char
+"'\\''"       Literal.String.Char
 ','           Punctuation
 ' '           Text.Whitespace
 "' '"         Literal.String.Char
--- a/tests/examplefiles/lhs/Sudoku.lhs.output
+++ b/tests/examplefiles/lhs/Sudoku.lhs.output
@@ -2536,8 +2536,7 @@
 '('           Punctuation
 'showChar'    Name
 ' '           Text.Whitespace
-"'\\"         Keyword.Type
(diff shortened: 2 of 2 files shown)


birkenfeld:
Thanks for the PR, can you add a test case?

mvanhorn:
Added in 0088b19 - snippet test covering `'\n'`, `'\t'`, `'\\'`, and `'a'` all tokenizing as `Literal.String.Char`.

birkenfeld:
Looks like outputs for the existing tests need to be adjusted as well.

mvanhorn:
Regenerated the golden outputs for example.hs and Sudoku.lhs in 35edbc8. All Haskell tests passing locally.

birkenfeld:
LGTM now, thanks!

mvanhorn:
Thanks for the merge!

PR #3078, opened 2026-03-30
Fix #3076 - Handle None object before HTML escaping

### LINKED ISSUE #3076: mkdocs build fails in HtmlFormatter with 'NoneType' object has no attribute 'replace'
We encountered a bug when building our HTML documentation via mkdocs build.
It seems that the README.md could not be parsed anymore.
The README.md was not changed.
The only change was upgrading to the new pygements version 2.20.0 as part of our automated regression tests.
After pining the version to 2.19.2 the build is not failing anymore.

```
ERROR   -  Error reading page 'README.md': 'NoneType' object has no attribute 'replace'
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:\Users\<User>\AppData\Local\pypoetry\Cache\virtualenvs\<lib-env>-py3.12\Scripts\mkdocs.exe\__main__.py", line 7, in <module>
  File "C:\Users\<User>\AppData\Local\pypoetry\Cache\virtualenvs\<lib-env>-py3.12\Lib\site-packages\click\core.py", line 1485, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\<User>\AppData\Local\pypoetry\Cache\virtualenvs\<lib-env>-py3.12\Lib\site-packages\click\core.py", line 1406, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "C:\Users\<User>\AppData\Local\pypoetry\Cache\virtualenvs\<lib-env>-py3.12\Lib\site-packages\click\core.py", line 1873, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\<User>\AppData\Local\pypoetry\Cache\virtualenvs\<lib-env>-py3.12\Lib\site-packages\click\core.py", line 1269, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\<User>\AppData\Local\pypoetry\Cache\virtualenvs\<lib-env>-py3.12\Lib\site-packages\click\core.py", line 824, in invoke
    return callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\<User>\AppData\Local\pypoetry\Cache\virtualenvs\<lib-env>-py3.12\Lib\site-packages\mkdocs\__main__.py", line 288, in build_command
    build.build(cfg, dirty=not clean)
  File "C:\Users\<User>\AppData\Local\pypoetry\Cache\virtualenvs\<lib-env>-py3.12\Lib\site-packages\mkdocs\commands\build.py", line 310, in build
    _populate_page(file.page, config, files, dirty)
  File "C:\Users\<User>\AppData\Local\pypoetry\Cache\virtualenvs\<lib-env>-py3.12\Lib\site-packages\mkdocs\commands\build.py", line 167, in _populate_page
    page.render(config, files)
  File "C:\Users\<User>\AppData\Local\pypoetry\Cache\virtualenvs\<lib-env>-py3.12\Lib\site-packages\mkdocs\structure\pages.py", line 285, in render
    self.content = md.convert(self.markdown)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\<User>\AppData\Local\pypoetry\Cache\virtualenvs\<lib-env>-py3.12\Lib\site-packages\markdown\core.py", line 367, in convert
    newRoot = treeprocessor.run(root)
              ^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\<User>\AppData\Local\pypoetry\Cache\virtualenvs\<lib-env>-py3.12\Lib\site-packages\pymdownx\highlight.py", line 527, in run
    code.highlight(
  File "C:\Users\<User>\AppData\Local\pypoetry\Cache\virtualenvs\<lib-env>-py3.12\Lib\site-packages\pymdownx\highlight.py", line 400, in highlight
    formatter = html_formatter(
                ^^^^^^^^^^^^^^^
  File "C:\Users\<User>\AppData\Local\pypoetry\Cache\virtualenvs\<lib-env>-py3.12\Lib\site-packages\pymdownx\highlight.py", line 179, in __init__
    HtmlFormatter.__init__(self, **options)
  File "C:\Users\<User>\AppData\Local\pypoetry\Cache\virtualenvs\<lib-env>-py3.12\Lib\site-packages\pygments\formatters\html.py", line 434, in __init__
    self.filename = html.escape(self._decodeifneeded(options.get('filename', '')))
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\<User>\AppData\Local\Programs\Python\Python312\Lib\html\__init__.py", line 19, in escape
    s = s.replace("&", "&amp;") # Must be done first!
        ^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'replace'
```

styx0x6:
My proposal to fix issue #3076.

Handle None object before HTML escaping regarding commit 0f97e7c.

Various text-only attributes using `html.escape` may also be NoneType objects, which seems not to be supported by `html.escape` which uses built-in `replace` function.

Handling None object will avoid further of this kind of error.

Fixes #3076.


birkenfeld:
Thanks for working on a fix! 

I don't think this will work as expected, we don't want a passed None to appear in the output as stringified "None".

Better to skip the escape() call if it's not a string.

birkenfeld:
Also, looks like a few new tests are needed 😁

styx0x6:
Hi,

You are right, also tought about this solution afterwards, string casting might have other side-effect. It's effectivelly more clean to skip the escape on None values. Will come back with an update.

styx0x6:
Here it is. Ready to run tests again @birkenfeld .

2bndy5 (commented):
Just to be painfully clear:
An empty string is not identical to a `None` type.

```py
>>> any_str = ""
>>> True if any_str else False
False
>>> True if any_str is not None else False
True
```
```py
>>> no_val = None
>>> True if no_val else False
False
>>> True if no_val is not None else False
False
```
I'm not sure if this bears any significance in the usage here. But the issue you're trying to solve is a dynamic typing problem. It makes sense to be stricter with your type checking, no?

birkenfeld:
It doesn't make sense to gratuitously break people's code, no.

styx0x6:
Hi @2bndy5 ,

That's right, an empty string is not the same as a `NoneType` object, but whatever, regarding the business logic of the code line here, it will just say that an empty string will not pass through the `html.escape`, which is not a security issue.

The important point here is to not let `None` being treated by `html.escape`. And that will fix the bug.

We might force the type checking but it would change nothing. Let's see that as avoiding useless code treatment and performance optimization 🙂 .

Anteru:
Just to be clear, it's people passing in `None`, not getting `None` from the options, as those all default to `''`, right? I'd rather have a helper function, `html_escape` or similar, in our utility functions, which is:
```python
def html_escape(string):
    """Return a safe version of the passed string, and '' if None"""
    if string is not None:
        return html.escape(string)
    return ''
```

So we can test this in isolation. Eventually, we can even add strong types to it (`str | None`) in this case. Then we can test this function (as trivial as it is ...) and test it in combination.

2bndy5:
Using a blanket `true if cond else false` makes sense as long as `html.escape()` has no special handling of empty strings. And I see that this is not a concern (see [`html.escape()` source](https://github.com/python/cpython/blob/99f333de7f0f62baca0649af6ea6673f81e56e9e/Lib/html/__init__.py#L12)). However, this could become a future regression if `html.escape()` ever changes about handling an empty string.

I agree that treating `None` and an empty string similarly could prove to be an optimization. Setting aside how empty iterators are handled, we could go further in the name of optimization: The `html.escape()` call should be after the `else` keyword, for the sake of various Python interpreters.

```py
# ensure `html.escape()` is only processed when needed by the interpreter
self.filename = _filename if not _filename else html.escape(_filename)
# interpreter stops at `else` when it is None or an empty string
```

I'm not trying to block anything up here. I'm only trying to be helpful.

birkenfeld:
@2bndy5 the order of the clauses does not matter, the if-else expression first evaluates the condition and then exactly one of the cases, depending on the result.

@styx0x6 the function suggested by @Anteru sounds like a good idea.

styx0x6:
Thanks all for your help and discussion.

I will try to offer a new version asap.

birkenfeld:
No problem, and thank you for actually contributing instead of letting Claude run wild. It's a breath of fresh air these days 🤣

styx0x6:
Here is an update, with trivial unit tests this time.

Also I discovered another clean-up code to do in `html.py` and `svg.py`, using an old `escape_html` function hardcoded directly, which seems useless now. I think it's better to clean-up and use the util function now. I'll set a new PR for the next feature release for that in order to not interact with this bugfix.

(@birkenfeld, I understand you, thinking the same. Claude is like a colleague, helpful to ask a question sometimes, but not here to do the job for you :) ).

Any comment is welcome.

Anteru:
Looks good to me. This will go into a 2.20.1 release, thanks again! Also good catch on the `escape_html` thing. I agree that should go away or be unified.

PR #3084, opened 2026-04-01
Add annotations/attributes (for C23 and C++26)

mikomikotaishi:
Attributes have existed since C++11, and were added to C23. Furthermore, annotations were added to C++26 with the inclusion of reflection.

Pygments currently does not support highlighting attributes on C/C++, so this pull request adds this long-needed feature as well as adds support for annotations for the upcoming C++26.

### DIFF
--- a/pygments/lexers/c_cpp.py
+++ b/pygments/lexers/c_cpp.py
@@ -204,6 +204,9 @@ class CFamilyLexer(RegexLexer):
             (r'.*?\n', Comment),
         ],
         'classname': [
+            include('whitespace'),
+            # C/C++ attributes before type name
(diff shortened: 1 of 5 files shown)


mikomikotaishi:
@birkenfeld @Anteru Any thoughts on this change? Would it be possible to get this in for 2.20.1?

Anteru:
Maybe. Need to take a closer look. I'm fairly risk-averse at this point, given the recent fallout.

mikomikotaishi:
Any update on this?

Anteru:
Thanks! Merged, but it won't go into a point release.

PR #3088, opened 2026-04-04
Add a Fountain lexer

mcepl:
## Summary
- add a new `FountainLexer` with context-aware parsing for screenplay blocks such as title pages, scene headings, character cues, dialogue, transitions, centered text, notes, boneyard comments, and sections
- support inline Fountain markup including emphasis, escapes, and scene numbers, and register the lexer for `.fountain` and `.spmd` files
- add real-world and focused test coverage with `Big-Fish.fountain`, a smaller example file, and snippet-based golden tests
## Testing
- `python3 -m pytest tests/snippets/fountain tests/examplefiles/fountain tests/test_guess.py`

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -12,6 +12,7 @@ Version 2.21.0
   * BitBake (#3103)
   * CEL (#3048)
   * PureScript (#1077, #3054)
+  * Fountain
 
 - Updated lexers:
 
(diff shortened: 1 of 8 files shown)


Anteru:
Can you please remove copyrighted screen plays from the PR? I think then we can move ahead with this.

mcepl:
> Can you please remove copyrighted screen plays from the PR? I think then we can move ahead with this.

Are you sure? I took those examples from https://fountain.io/

mcepl:
> Can you please remove copyrighted screen plays from the PR? I think then we can move ahead with this.

@Anteru  Done. Should I find some public domain (even mine ;)) piece of screenplay, or we would just go without?

PR #3092, opened 2026-04-06
Add additional keywords coming in C2Y

mikomikotaishi:
This pull request adds support for keywords coming to C2Y, including `defer` and `countof`.

### DIFF
--- a/pygments/lexers/c_cpp.py
+++ b/pygments/lexers/c_cpp.py
@@ -113,8 +113,8 @@ class CFamilyLexer(RegexLexer):
         'keywords': [
             (r'(struct|union)(\s+)', bygroups(Keyword, Whitespace), 'classname'),
             (r'case\b', Keyword, 'case-value'),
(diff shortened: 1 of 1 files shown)


mikomikotaishi:
@Anteru @birkenfeld I would also like to furthermore suggest adding as an additional option a `commonaliaseshighlight` for common primitive aliases used in C++. I have seen aliases like `u8`/`u16`/`u32`/`u64`/`i8`/`i16`/`i32`/`i64`/`usize`/`isize`/etc. used in some projects, for Rust-style aliases for the fixed-width types from the standard library. If such a feature is worthy of inclusion, I think it should be by default off.

Anteru:
Thanks for the contribution! I'm not convinced of `u8` and others, my code for example using `uint8` and now we'd have to make a fairly broad list of "common" aliases, and I don't think that's going to serve people well. We won't be able to agree on a common list of aliases.

PR #3093, opened 2026-04-07
Add SVG filenames

roskakori:
While SVG was already part of the mime types, its filename was not recognized yet.


birkenfeld:
LGTM, thanks!

PR #3099, opened 2026-04-17
Add support for Python 3.15

hugovk:
3.15 is virtually in beta and 3.9 went EOL last October:

https://devguide.python.org/versions/


birkenfeld:
I'm not sure we need to drop 3.9 as long as we don't use any 3.10 only feature?

hugovk:
Sure, I've removed that commit.

birkenfeld:
Thanks!

PR #3100, opened 2026-04-17
Monokai: improve `pytb` output

hugovk:
The CPython docs use the default Pygments theme for light mode and Monokai for dark.

Compare the traceback at https://docs.python.org/3.15/whatsnew/3.13.html#improved-error-messages

Light looks good:

image

But dark is almost monochrome:

image

This PR adds some colour:

image

They're all above the minimum ratio of the WCAG AA guidelines.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -2,7 +2,7 @@ name: Pygments
 
 on:
   push:
-    branches: [master]
+    branches: ['*']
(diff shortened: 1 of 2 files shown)


Anteru:
Why is this trying to change the workflow file? The style change is fine.

Anteru:
Merged, thanks!

PR #3110, opened 2026-05-01
Require a word break after `SET` as a MySQL keyword

rolandwalker:
Per a bug report in the Mycli project:

 * https://github.com/dbcli/mycli/issues/1872

`SET` is tokenizing as a `Keyword` even when part of a `Name` in `MySqlLexer`.

The fix proposed is to require a word break after `SET`, and before the negative lookahead assertion.

There is an identical construct in `GoogleSqlLexer`, which is not changed.

### DIFF
--- a/pygments/lexers/sql.py
+++ b/pygments/lexers/sql.py
@@ -764,7 +764,7 @@ class MySqlLexer(RegexLexer):
             (r'[!%&*+/:<=>^|~-]+', Operator),
 
             # Exceptions; these words tokenize differently in different contexts.
-            (r'\b(set)(?!\s*\()', Keyword),
+            (r'\b(set)\b(?!\s*\()', Keyword),
             (r'\b(character)(\s+)(set)\b', bygroups(Keyword, Whitespace, Keyword)),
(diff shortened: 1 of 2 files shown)


tests/test_mysql.py
```diff
@@ -171,6 +171,19 @@ def test_exceptions(lexer, text, expected_types):
     assert all(t[0] == e for t, e in zip(tokens, expected_types))
 
 
+@pytest.mark.parametrize(
+    'text, expected_types',
+    (
+        # SET not-keyword-exceptions
+        # Name is the default expectation
+        ('SET_NONSENSE', (Name,)),
```
birkenfeld: IF there's only one case, `parametrize` is not needed. In general, it's better to use the new snippet-based test system by now.

birkenfeld (changes requested):

tests/test_mysql.py
rolandwalker: Done!

birkenfeld (approved):
Thank you!

PR #3113, opened 2026-05-01
Add TextWithCommentLexer for plain text with hash comments

Leistungsabfall:
Adds a minimal `TextWithCommentLexer` that highlights `#` comments while treating everything else as plain text.

**Possible use cases:**
This lexer provides minimal useful highlighting for files that are essentially plain text but use `#` for annotations — especially in terminal-based editors and preview tools.

### `.gitignore`
Line-based files where only comments need visual separation from content.
gitignore_example

### `git rebase -i`
The instructions are `#` comments, the rest is user-editable content. Highlighting comments helps distinguish actionable lines from help text.
rebase_example

### Notes and scratch files
Quick TODO files where `#` marks section headers or meta-notes vs. actual content. Inline `#` comments annotate individual lines.
notes_example

**Changes:**
  - `pygments/lexers/textfmts.py` — new `TextWithCommentLexer` class
  - `tests/snippets/text-with-comments/test_comments.txt` — golden test
  - `pygments/lexers/_mapping.py` — regenerated mapping

### DIFF
--- a/tests/snippets/text-with-comments/test_comments.txt
+++ b/tests/snippets/text-with-comments/test_comments.txt
@@ -1,0 +1,28 @@
+---input---
+text # inline comment
+# full line comment
+#full line comment
+text#invalid inline comment
+text #inline comment
+text# invalid inline comment
+
+---tokens---
+'text '       Text
+'# inline comment' Comment.Single
+'\n'          Text.Whitespace
+
+'# full line comment' Comment.Single
+'\n'          Text.Whitespace
+
+'#full line comment' Comment.Single
+'\n'          Text.Whitespace
+
+'text#invalid inline comment' Text
+'\n'          Text.Whitespace
+
+'text '       Text
+'#inline comment' Comment.Single
+'\n'          Text.Whitespace
+
+'text# invalid inline comment' Text
+'\n'          Text.Whitespace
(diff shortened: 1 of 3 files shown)


birkenfeld:
Thanks for the PR! The implementation here is a little inefficient; you should also match `[^\s#]+` as `Text` before the catch-all `.` expression.

Leistungsabfall:
Thanks for the hint @birkenfeld.
I updated the PR.

birkenfeld:
Thanks! One more thing I didn't realize, this module (`special`) should remain for actually special lexers. Can you move the new one to its own?

Leistungsabfall:
Sure!
I checked the codebase for similar lexers, hope `textfmts.py` is fine.

tests/snippets/text-with-comments/test_comments.txt
```diff
@@ -0,0 +1,29 @@
...
+text # inline comment
+# full line comment
+#full line comment
+text#invalid inline comment
+text #inline comment
+text# inline comment
+
+---tokens---
+'text'        Text
+' # inline comment' Comment
```
birkenfeld: So... do we want to start `Comment` at the hash or the whitespace?

Leistungsabfall: With the last commit I wanted to reduce the amount of tokens produced. Is it problematic if `Comment` starts on Whitespace? If yes, I'd just revert the last commit.

Leistungsabfall:
@birkenfeld I took a look at other lexers that handle comments and how they deal with Whitespace characters.
I rewrote the whole lexer to follow those conventions.
IMHO it's better now, but if something needs to be changed, please let me know.

PR #3114, opened 2026-05-01
Python lexer: add support for 3.15 'sentinel' built-in

rodrigogiraoserrao:
Adds the built-in 'sentinel' coming in Python 3.15 as per PEP 661, https://peps.python.org/pep-0661/


birkenfeld:
Nice, thanks a lot for the PR!

PR #3115, opened 2026-05-01
Python lexer: add support for lazy imports in 3.15

rodrigogiraoserrao:
Adds support for explicit lazy imports, coming in Python 3.15 as per PEP 810, https://peps.python.org/pep-0810/.

I wasn't sure whether `lazy` should be a token of the type `Keyword` or `Keyword.Namespace`.

If this PR is well-received, I'll make a similar one to add support for the soft keyword `type`, used in [type statements](https://docs.python.org/3/reference/simple_stmts.html#type).


birkenfeld:
Thanks for the PR! I think `Keyword.Namespace` will read nicer.

rodrigogiraoserrao:
> Thanks for the PR!

My pleasure!

> I think `Keyword.Namespace` will read nicer.

I guessed `Keyword` maybe wasn't enough. I definitely don't disagree, so I went ahead and made that change in 186057d. Is there something I could read somewhere to help me understand how you pick the exact token types? Or is it something you just develop with time and through familiarity with the `pygments` project?

birkenfeld:
Nothing specific, just the feeling that it's better to not have the import lines too "flashy" with many different styles.

rodrigogiraoserrao:
> Nothing specific, just the feeling that it's better to not have the import lines too "flashy" with many different styles.

Fair enough, makes sense to me @birkenfeld. The PR has been updated!

birkenfeld:
Thanks, LGTM!

PR #3117, opened 2026-05-02
Python lexer: add support for type statements

rodrigogiraoserrao:
Highlights the soft keyword 'type' as a keyword instead of a built-in in scenarios that look like type statements, added in Python 3.12 (https://docs.python.org/3/reference/simple_stmts.html#the-type-statement).

Using the theme 'dracula', before the change `type` always has the same colour:

Screenshot 2026-05-02 at 13 39 33

After the change, `type` has the colour of a keyword and the type alias is coloured like a class name (inspiration taken from how VS Code highlights type statements):

Screenshot 2026-05-02 at 13 39 47

(This is the snippet of code used in the test I added.)

P.S. I know pygments isn't supposed to _validate_ code. The regex I used to check if `type` is supposed to be highlightd like a keyword is something like “we're at the beginning of the line, we see the keyword, then whitespace, then a valid name, then further down the line we find an equals sign”. I think this is _enough_ while not being too restrictive, but I could be wrong.

### DIFF
--- a/pygments/lexers/python.py
+++ b/pygments/lexers/python.py
@@ -227,6 +227,12 @@ class PythonLexer(RegexLexer):
                                        # pattern matching (but None/True/False is ok)
              r'|'.join(k for k in keyword.kwlist if k[0].islower()) + r')\b))',
              bygroups(Text, Keyword), 'soft-keywords-inner'),
+            # `type` soft keyword
+            (r'(^[ \t]*)'                    # beginning of line + indentation
+             r'(type)\b'                     # the keyword 'type' followed by
+             r'([ \t]+)(' + uni_name + r')'  # the type alias name
+             r'(?=[ \t]*(?:[^\n]*)?=)',      # followed by an eventual assignment
+             bygroups(Text, Keyword, Text, Name.Class)),
         ],
         'soft-keywords-inner': [
             # optional `_` keyword
--- a/pygments/lexers/python.py
+++ b/pygments/lexers/python.py
@@ -231,7 +231,7 @@ class PythonLexer(RegexLexer):
             (r'(^[ \t]*)'                    # beginning of line + indentation
              r'(type)\b'                     # the keyword 'type' followed by
              r'([ \t]+)(' + uni_name + r')'  # the type alias name
-             r'(?=[ \t]*(?:[^\n]*)?=)',      # followed by an eventual assignment
+             r'(?=[^\n]*=)',                 # followed by an eventual assignment
              bygroups(Text, Keyword, Text, Name.Class)),
         ],
         'soft-keywords-inner': [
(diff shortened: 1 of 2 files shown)


pygments/lexers/python.py
```diff
@@ -227,6 +227,12 @@ def fstring_rules(ttype):
                                        # pattern matching (but None/True/False is ok)
              r'|'.join(k for k in keyword.kwlist if k[0].islower()) + r')\b))',
              bygroups(Text, Keyword), 'soft-keywords-inner'),
+            # `type` soft keyword
+            (r'(^[ \t]*)'                    # beginning of line + indentation
+             r'(type)\b'                     # the keyword 'type' followed by
+             r'([ \t]+)(' + uni_name + r')'  # the type alias name
+             r'(?=[ \t]*(?:[^\n]*)?=)',      # followed by an eventual assignment
```
birkenfeld: The additional `?` is unnecessary, `*` already allows 0 repetition, and matching the spaces separately is unneeded.

```suggestion
             r'(?=[^\n]*=)',      # followed by an eventual assignment
```

birkenfeld (changes requested):

birkenfeld:
I agree this is enough sanity checking for Pygments' sake.

rodrigogiraoserrao:
> I agree this is enough sanity checking for Pygments' sake.

My regex isn't good enough, though... If I split my generics across multiple lines, then `type` goes back to being highlighted as the built-in:

Screenshot 2026-05-06 at 14 19 37

What do you reckon? Is it worth creating a more complex system that handles this situation? Do we just go with the current solution, which is already an _improvement_ over what's currently available?

rodrigogiraoserrao:
Maybe the regex could follow a different approach... Instead of looking for an equals sign, that looks like an assignment, it could just check if there's 1+ whitespace chars and a letter or underscore (the valid chars for the first character of a Python identifier), which should be enough to distinguish from when `type` is used as a built-in name/valid variable name.

birkenfeld:
> Maybe the regex could follow a different approach... Instead of looking for an equals sign, that looks like an assignment, it could just check if there's 1+ whitespace chars and a letter or underscore (the valid chars for the first character of a Python identifier), which should be enough to distinguish from when `type` is used as a built-in name/valid variable name.

Sounds good to me. I don't think we need to recognize `"type \\\nxyz"`.

rodrigogiraoserrao:
> Sounds good to me. I don't think we need to recognize `"type \\\nxyz"`.

Thanks for the guidance. I confirm snippets of code with line-continuation characters in type statements, such as 

```py
type \
    Pair[T] = tuple[T, T]
```

will be mistokenised. I left a note in 2ef12f8.

This PR should be ready for your review/acceptance.

rodrigogiraoserrao:
@birkenfeld let me know if there's anything I can do to unblock this PR :)

rodrigogiraoserrao:
Dear Georg @birkenfeld, let me know how I can help you take this across the finish line :)
Thank you for your time 🙏

PR #3131, opened 2026-05-14
Add *.http filename extension to HttpLexer

### LINKED ISSUE #2842: please associate `*.http` for HttpLexer
I'm using https://github.com/facelessuser/pymdown-extensions plugins and https://github.com/squidfunk/mkdocs-material to write docs. I don't know how to let it highlight HTTP files because there aren't any filenames associated with HttpLexer I guess (https://pygments.org/docs/lexers/#pygments.lexers.textfmts.HttpLexer). 
Maybe, add `*.http`.

Github can also highlight HTTP file

```http
### basic get statement
GET https://api.github.com/users/boltlessengineer
user-agent: neovim
```

Trung6405:
## Summary
Adds `*.http` filename association to the `HttpLexer`.

Previously, there were no filenames associated with `HttpLexer`, so tools that rely on filename-based lexer detection (such as pymdown-extensions and mkdocs-material) had no way to automatically highlight `.http` files. For example, a file containing:

```http
GET https://api.github.com/users/boltlessengineer
user-agent: neovim
```

would not be syntax highlighted when saved as a `.http` file.

## Changes
* Added `filenames = ['*.http']` to `HttpLexer` in `pygments/lexers/textfmts.py`
* Updated the lexer mapping cache in `pygments/lexers/_mapping.py`
* Renamed existing HTTP example files to use the `.http` extension to match the new filename pattern

## Result
Files with a `.http` extension are now correctly identified and highlighted by `HttpLexer`, consistent with how GitHub highlights HTTP files.

Fixes #2842


tests/examplefiles/http/http_request_example.http
```diff
@@ -1,15 +1,15 @@
-POST /demo/submit/ HTTP/1.1
```
birkenfeld, line 1: Why has the content changed here?

Trung6405: Sorry about that, it was caused by core.autocrlf converting the \r\n line endings during the rename. I've fixed it by disabling autocrlf and re-copying the files.

randomstuff:
Warning, `.http` is already used by tools (some IDEs) for something which is not strictly an HTTP message (which could have its own lexer in Pygments at some point):

https://timdeschryver.dev/bits/http-files

(FWIW, I do think that it would be more logical to use `.http` for actual real HTTP message …)

Trung6405:
@randomstuff thanks for raising this! I hadn't considered the overlap with IDE tooling. I'll leave this decision to the maintainers since it affects the broader project. @birkenfeld happy to switch to an alternative extension or adjust the approach however you see fit.

PR #3137, opened 2026-05-18
Fix(Ruby): Reorder operators to correctly tokenize <=>

### LINKED ISSUE #2791: The Ruby operator "<=>" is sometimes parsed as "<" and "=>" or "<", "=", ">", other operators longer than one can also be confused with other operators, as is the case in C/C++

fiona3758:
## Summary
Fixes #2791 

Fixes operator tokenization in the Ruby lexer where `<=>` was being incorrectly split into `<` + `=>` or `<` + `=` + `>` due to shorter patterns matching first. Fixed by reordering the regex alternations so that longer operators like `<=>` are matched before shorter overlapping ones (`<`, `<=`, `>`). Also, added `=>` as a recognized operator token which was previously missing.


birkenfeld:
Thanks for the PR! Looks like the test outputs for existing tests weren't updated. Also, I don't think we need both added tests - either the two-file version or the one-file version is enough.

PR #3138, opened 2026-05-19
(Fish) Fix fish single quote backslash escape

### LINKED ISSUE #2821: Single Quote Escaping Inside Single-Quoted String in Fish Doesn't Work
I have the following snippet of fish shell code:
```fish
echo 'couldn\'t find a virtual environment to activate' >&2
```

Unfortunately, Pygments appears to not understand the [backslash quote escape](https://fishshell.com/docs/current/language.html#quotes):

image

987cj:
Fixes #2821 

Implements a fix for backslash quote escapes, and a small snippet test for them.

In regards to the ending backslashes, I’ve made the following changes:
    `'abc\\'` works (backslash escape)
    `'abc\'` doesn’t work (registers as a quote escape, string not properly closed
    `'abc\''` works (properly escaped quote, string closed)

This is tested within the code snippet : )

This works by recognising whether a quote is escaped, rather than recognising all escape characters (`'abc\\\'` registers as a full string, despite the escapes not actually closing the string properly). As fish has some settings for handling escape characters, I felt it best to do it this way so it doesn't accidentally break some functionality I'm not aware of.


birkenfeld:
Hm, I think this is better handled without lookarounds. Something like `r"'(\\\\|\\[^\\]|[^'\\])*'"` - we already have that in use a lot.

987cj:
Changed the regex to that :) Thanks!

birkenfeld:
Nice, thanks!

scolby33:
Thanks!

PR #3139, opened 2026-05-19
Fix mention regex in MarkdownLexer to support hyphens in usernames

### LINKED ISSUE #3135: Markdown lexer should include "-" as a valid character in mentions
"-" is a valid character in usernames on GitHub, but the regex that matches it for mentions currently only checks for word characters. To get mentions properly highlighted "-" should be added to the regex.

Example: `@user-name`

Trung6405:
## Summary
Fixes the mention regex in `MarkdownLexer` to support hyphens in usernames.

Previously, the regex that detects mentions only matched word characters (`\w`),
which covers letters, numbers, and underscores but not `-`. This caused hyphenated
usernames like `@john-doe` to be split into `@john` (highlighted as `Name.Entity`)
and `-doe` (left as plain `Text`).

For example, a mention like:
```markdown
@john-doe mentioned this issue
```
would only highlight `@john`, leaving `-doe` unhighlighted.

## Changes

* Updated the mention regex in `pygments/lexers/markup.py` to include `-` as a valid character
* Added `tests/examplefiles/markdown/example.md` with test cases for hyphenated mentions and topics

## Result
Mentions with hyphens (e.g. `@john-doe`) are now correctly tokenised as a single
`Name.Entity` token, consistent with how GitHub handle hyphenated usernames.

Fixes #3135


birkenfeld:
Nice fix, thanks!

PR #3144, opened 2026-05-21
Base the CUDA lexer on C++ instead of C

### LINKED ISSUE #3127: Cuda lexer derived from C instead of C++
modern cuda code is typically based on C++ (and nvcc is a C++ compiler); the current cuda lexer is based on C, so it misses, e.g., `constexpr` .

mokashang:
Fixes #3127.

## Problem

CUDA code is compiled by `nvcc`, which is a C++ compiler, and modern CUDA routinely uses C++ constructs (templates, classes, namespaces, `constexpr`, `nullptr`, `auto`, etc.). However, `CudaLexer` derived from `CLexer`, so these were highlighted as plain `Name` tokens instead of keywords. For example, `template`, `class` and `namespace` all came out as `Token.Name`.

## Change

Derive `CudaLexer` from `CppLexer` instead of `CLexer` (and delegate `get_tokens_unprocessed` to `CppLexer` accordingly).

The CUDA-specific post-processing in `get_tokens_unprocessed` only rewrites tokens that come through as `Name`. C++ keywords are emitted as `Keyword` by `CppLexer`, so they don't collide with the CUDA detection, and the CUDA qualifiers, vector types, builtins and functions (`__global__`, `__restrict__`, `float4`, `dim3`, `threadIdx`, `__syncthreads`, …) are still recognized exactly as before.

## Testing

- Extended `tests/examplefiles/cuda/test.cu` with a modern CUDA C++ section (a templated `__global__` kernel, a `class`, a `namespace`, `constexpr`/`auto`/`nullptr`) and regenerated the golden output with `--update-goldens`. The snapshot diff is purely additive — the existing CUDA tokens are unchanged, and the new C++ constructs are now highlighted as keywords while `__global__` etc. remain `Keyword.Reserved`.
- `tox`/`pytest` full suite passes locally (the only failure is the LaTeX formatter test, which requires a working `latex` binary and is unrelated to this change).
- `ruff check --ignore UP031 .` passes.


birkenfeld:
LGTM, thanks!

PR #3152, opened 2026-06-03
Markdown: highlight bold-italics

### LINKED ISSUE #3067: Pygments MarkdownLexer does not handle bold-italics
***Bold italics*** (`***bold italics*** or ___bold italics___`) is not handled by MarkdownLexer.

Since there's already a suitable Token (Token.Generic.EmphStrong) adding support was trivial, so I created a fix, tested it; everything seems to work fine. Patch attached.

[markdown_bold_italics.patch](https://github.com/user-attachments/files/26253576/markdown_bold_italics.patch)

The patch only does one thing:

It emits Generic.EmphStrong when bold-italics tags are encountered.

code-with-rashid:
The `MarkdownLexer` highlights `**bold**` and `*italic*` spans, but the combined bold-italics form `***bold italics***` (and `___bold italics___`) was not recognized and fell through to plain `Text`:

```python
>>> from pygments.lexers.markup import MarkdownLexer
>>> [(str(t), v) for t, v in MarkdownLexer().get_tokens('***bold italics***') if v.strip()]
[('Token.Text', '***bold'), ('Token.Text', 'italics***')]
```

Pygments already has a dedicated `Generic.EmphStrong` token for exactly this case, so this adds two inline rules that emit it. They are placed before the existing `**`/`__` bold rules so the triple markers aren't partially consumed.

After:

```python
>>> [(str(t), v) for t, v in MarkdownLexer().get_tokens('***bold italics***') if v.strip()]
[('Token.Generic.EmphStrong', '***bold italics***')]
```

Existing `**bold**`, `__bold__`, `*italic*`, `_italic_` and the `snake***case***word` non-emphasis behavior are unchanged.

### Tests

Extended `tests/examplefiles/md/example.md` with bold-italics lines (both the emphasized form and the `snake***case***word` non-match) and regenerated the golden output. `pytest tests/examplefiles/md/ tests/test_basic_api.py` passes (2446 passed, 8 skipped).

Closes #3067


birkenfeld:
Nice, thanks!

PR #3157, opened 2026-06-14
Inline import of importlib.metadata

danielhollas:
`importlib.metadata` is a relatively slow to import so let's import it only when actually needed.
I found this while investigating the import time of `pip --help`.


birkenfeld:
Sure, why not.

PR #3158, opened 2026-06-16
Fix wrong token offsets for embedded code blocks (Markdown/RST/TiddlyWiki)

### LINKED ISSUE #3133: Update from 2.19.1 to 2.20.0 destroys markdown rendering of code blocks
The unilateral github update from 2.19.1 to 2.20.0 changed the rendering of all markdown code blocks (identifiled with ```cpp header) 

The entire code block was rendered as a single string, including the cpp marker) with no line breaks.

for example 

```cpp
foo(a);
bar(b);
```

is incorrectly rendered by 2.20.0 as  "cpp foo(a); bar(b);"

Adding Pygments==2.19.1 to requirements.txt corrects the problem.
  
See attached files for requirtements.txt and a typical .md file containing a code block.

[passing-loop.md](https://github.com/user-attachments/files/27854167/passing-loop.md)
[requirements.txt](https://github.com/user-attachments/files/27854168/requirements.txt)

sarathfrancis90:
Fixes #3133.

### Bug
The Markdown, reStructuredText and TiddlyWiki5 lexers all delegate the body of an embedded code block to a sub-lexer with `lexer.get_tokens_unprocessed(code)`. That call returns indices relative to the `code` snippet (starting at `0`), but the indices the outer lexer yields are supposed to be **absolute** positions within the whole input text. So every token inside a fenced/indented code block came back with a wrong offset. The `MarkdownLexer` even had a `# FIXME: aren't the offsets wrong?` comment about exactly this.

Minimal check:

```python
from pygments.lexers.markup import MarkdownLexer
text = "intro\n```python\nx = 1\n```\n"
for index, _tok, value in MarkdownLexer().get_tokens_unprocessed(text):
    assert text[index:index+len(value)] == value  # fails on the embedded code tokens
```

### Cause
The delegated token stream (and the `do_insertions(...)` output wrapping it) is never re-based by the start position of the code group, so the sub-lexer's 0-based indices leak out unchanged.

### Fix
Added a small `_shift_indices(tokens, offset)` helper and re-based the delegated tokens by `match.start(<code group>)` at the four affected sites (`RstLexer._handle_sourcecode`, `MarkdownLexer._handle_codeblock`, `TiddlyWiki5Lexer._handle_codeblock`, `TiddlyWiki5Lexer._handle_cssblock`). The plain `get_tokens` output and the rendered HTML are byte-for-byte unchanged — only the previously incorrect positions from `get_tokens_unprocessed` are corrected.

### Testing
Added `tests/test_markup.py` asserting that every token's index points at its own value in the input, for all three lexers (fails before, passes after). Full suite `5262 passed, 16 skipped`; `ruff check --ignore UP031 .` clean.


birkenfeld:
Nice, thanks!

PR #3163, opened 2026-06-21
D: Allow non-ASCII (Unicode) identifiers

### LINKED ISSUE #1088: D lang lexer doesn't parse correctly non Ascii characters
(Original issue 1384 created by None on 2017-10-22T13:32:46.655107+00:00)

Pygments D lang lexer doesn't parse non Ascii characters correctly and gives an error.

Consider this code snippet:

```
#!d

class Pencere: AnaPencere
{
    Düğme düğme;

    this()
    {
        super("Merhaba Dünya");
        düğme = new Düğme("Bana tıklayın");
        this.ekle(düğme);
        düğme.tıklandığında(&iletiGöster);
        tümünüGöster();
    }

    void ğĞıIiİşŞÜüçÇiletiGöster(Düğme düğme)
    {
        writeln("Bana tıkladınız");
    }
}
```
As you'd see some of the non ASCII characters çÇğĞşŞüÜıI in the source code are not rendered correctly.

ardittirana:
Fixes #1088.

### Problem
The D lexer matches identifiers with `[a-zA-Z_]\w*`. The continuation class `\w*` is Unicode-aware in Python 3, but the **leading** class `[a-zA-Z_]` is ASCII-only. D allows Unicode letters in identifiers (`UniversalAlpha` in `IdentifierStart`), so an identifier that *starts* with a non-ASCII letter is mis-tokenized — its leading characters become `Error` tokens.

Reproducer from the issue (the method name starts with `ğ`):
```d
void ğĞıIiİşŞÜüçÇiletiGöster(Düğme düğme) {}
```
Before this change the lexer emits `Error` tokens for `ğ`, `Ğ`, `ı`.

### Fix
Use `[^\W\d]\w*` — a word character that is not a digit, i.e. a Unicode letter or underscore — for the identifier rule. The same change is applied to the two other places a D identifier appears: the UDA decorator (`@name`) and the `q"name\n...\nname"` heredoc delimiter.

`[^\W\d]\w*` is a strict superset of `[a-zA-Z_]\w*`, so ASCII highlighting is unchanged (the existing `tests/examplefiles/d` goldens still pass).

### Test
Adds `tests/snippets/d/test_unicode_identifiers.txt`. It fails on `master` (Error tokens) and passes with this change; all Unicode identifiers tokenize as `Name`.


birkenfeld:
Nice and clean, thanks!

PR #3164, opened 2026-06-21
MATLAB, Octave, Scilab: Allow line continuations in function signatures

### LINKED ISSUE #659: MatlabLexer returns Token.Error if function definition has ellipsis anywhere in it
(Original issue 952 created by bwanamarko on 2014-01-31T07:30:39.245903+00:00)

the following are valid MATLAB/Octave/SciLab, but all fail:

    function [ ...
        o1, o1] = my_fun(a1, a2)

    function [o1, ...
        o1] = my_fun(a1, a2)

    function [o1, o1] = ...
        my_fun(a1, a2)

    function [o1, o1] = my_fun( ...
        a1, a2)

    function [o1, o1] = my_fun(a1, ...
        a2)

add this regex `(?:...\n[ \t]*)?` where ever there is a chance of ellipsis? It would be nice if they were lexed as Token.Comments with whatever characters follow the ellipsis up to but not including the newline character

thanks!

ardittirana:
Fixes #659.

### Problem
The `MatlabLexer`, `OctaveLexer` and `ScilabLexer` all matched a function-definition signature with a single regex that assumed the whole `function ... = name(...)` declaration was on one line:

```python
(r'(\s*)(?:(\S+)(\s*)(=)(\s*))?(.+)(\()(.*)(\))(\s*)', ...)
```

A `...` line continuation — valid anywhere in the signature in MATLAB/Octave/Scilab — makes that match fail, so the signature is emitted as `Error` tokens. All of these (from the issue) were broken:

```matlab
function [ ...
    o1, o2] = my_fun(a1, a2)
function [o1, o2] = ...
    my_fun(a1, a2)
function [o1, o2] = my_fun(a1, ...
    a2)
```

### Fix
Replace the single-regex `deffunc` state (which was duplicated verbatim in all three lexers) with a small shared state machine (`deffunc` + `deffunc_args`) that tokenises the signature incrementally and treats `...` continuations as `Comment`, as suggested in the issue. Because the three lexers shared the identical rule, the fix is defined once and reused.

Single-line definitions tokenise **exactly as before** — all existing `examplefiles`/`snippets` goldens for matlab/octave/scilab are unchanged.

### Test
Adds `tests/snippets/matlab/test_function_definition_line_continuation.txt` (continuation in both the output list and the argument list). It fails on `master` (Error tokens) and passes with this change.


birkenfeld:
LGTM, thanks!

PR #3165, opened 2026-06-22
Tcl: stop `$name` variable references at a dash

### LINKED ISSUE #1720: tcl variable name dash
```tcl
set foo lol
puts a-$foo-b
```

This prints `a-lol-b`, treating `foo` as a variable name, but Pygments highlight this as if `foo-b` was a variable name:

ffff

ardittirana:
Fixes #1720.

### Problem
The Tcl lexer matches `$` variable names with `[\w.:-]+`, which includes `-`. So in:

```tcl
set foo lol
puts a-$foo-b
```

it highlights `foo-b` as the variable, but Tcl reads the variable `foo` here (`a-$foo-b` evaluates to `a-lol-b`). As confirmed in the issue thread, a `$name` reference ends at a dash; only the braced `${foo-bar}` form may contain dashes.

### Fix
Drop `-` from the unbraced `$name` pattern (`\$[\w.:]+`), leaving the braced `${...}` pattern unchanged so `${foo-bar}` still lexes as one variable. Namespace (`$ns::var`) references are unaffected.

### Test
Adds `tests/snippets/tcl/test_variable_dash.txt` covering both `a-$foo-b` (now `$foo` + `-` + `b`) and `${foo-bar}` (unchanged). It fails on `master` and passes with this change; existing Tcl snippets are unchanged.


birkenfeld:
Nice, thanks for the PR!

PR #3166, opened 2026-06-24
fix: handle style=None in _lookup_style to avoid cryptic TypeError

gaoflow:
Passing `style=None` to any formatter crashes with a cryptic
`TypeError: 'NoneType' object is not iterable` because `_lookup_style()`
returns `None` verbatim when it should return the default style.

**Root cause:** `_lookup_style()` in `pygments/formatter.py` has no
`None` guard. `None` is a valid default for the `style` parameter
in all formatters, which default to `'default'`. But `None` is not a
string so it bypasses the `isinstance(str)` check and is returned
as-is, causing every formatter to crash when iterating style attributes.

**Fix (+2 lines):** `if style is None: style = 'default'` at the top
of `_lookup_style()`.

**Reproduction:**

5268/5268 tests pass.

This pull request was prepared with the assistance of AI, under my
direction and review.

### DIFF
--- a/pygments/formatter.py
+++ b/pygments/formatter.py
@@ -17,6 +17,8 @@ __all__ = ['Formatter']
 
 
 def _lookup_style(style):
+    if style is None:
+        style = 'default'
     if isinstance(style, str):
         return get_style_by_name(style)
     return style


birkenfeld:
But why would you pass `style=None`? Is that mentioned somewhere?

binggao1230:
It isn't, and I checked before answering rather than guessing. The docstring says string or Style subclass; nothing in Pygments passes `None`; `_lookup_style` has exactly one caller; and there's no issue behind this. It came out of a None-safety sweep, not a real report, so the premise doesn't hold up.

The only part that's arguably real is how far the failure lands from the cause — `HtmlFormatter(style=None)` raises `TypeError: 'NoneType' object is not iterable` well after construction, and `TerminalFormatter(style=None)` doesn't complain at all. But that's someone passing an undocumented value, not a bug in Pygments.

Happy to close this.

birkenfeld:
Thanks Claude.

PR #3167, opened 2026-06-26
Add missing word boundary to GoogleSqlLexer 'set' keyword

mahirhir:
GoogleSqlLexer's keyword pattern for `set` lacks a trailing `\b`, so an identifier like `settings` is lexed as the keyword `set` followed by `tings` instead of a single name. The fix adds the missing word boundary.

This is the same fix merged in #3110 for the adjacent lexer; GoogleSql was the case left unmirrored, which the maintainer noted as still-unfixed. Existing suite stays green (796 passed).


birkenfeld:
Thanks for the PR!

PR #3168, opened 2026-06-26
Smithy: fix text blocks and add missing syntax

JordonPhillips:
The Smithy lexer in pygments predates Smithy 2.0, and so it was missing support for the new syntax added by it. This fixes those gaps:

  - enum / intEnum are now recognized as shape keywords
  - the with and for keywords (mixins and resource binding)
  - the := inline operator for operation input/output shapes
  - target elision, e.g. $id inside a structure
  - control statements other than $version (e.g. $operationInputSuffix); the rule was hardcoded to $version

Additionally, this fixes text blocks, which had no branch for ordinary characters. That made it so any block containing actual text failed to match and got shredded into a run of String.Double tokens, with words from the body leaking out as identifiers.

([smithy docs for reference](https://smithy.io))


birkenfeld:
Thanks for the PR!

PR #3171, opened 2026-07-02
Ruby: don't duplicate the body of an unterminated heredoc (#2998)

### LINKED ISSUE #2998: Ruby snippet leads to duplicated output
When formatting (partial) ruby, the output contains one bit twice.

This input (7 lines long)

```
        code = """        #  - format_type includes the column size constraint, e.g. varchar(50)
        #  - ::regclass is a function that gives the id for a table name
        def column_definitions(table_name) #:nodoc:
          exec_query(<<-end_sql, 'SCHEMA').rows
              SELECT a.attname, format_type(a.atttypid, a.atttypmod),
                     pg_get_expr(d.adbin, d.adrelid), a.attnotnull, a.atttypid, a.atttypmod
                FROM pg_attribute a LEFT JOIN pg_attrdef d"""
```

Leads to this output, which is longer (10 lines)

```
<span class="w">        </span><span class="c1">#  - format_type includes the column size constraint, e.g. varchar(50)</span>
<span class="w">        </span><span class="c1">#  - ::regclass is a function that gives the id for a table name</span>
<span class="w">        </span><span class="k">def</span><span class="w"> </span><span class="nf">column_definitions</span><span class="p">(</span><span class="n">table_name</span><span class="p">)</span><span class="w"> </span><span class="c1">#:nodoc:</span>
<span class="w">          </span><span class="n">exec_query</span><span class="p">(</span><span class="o">&lt;&lt;-</span><span class="dl">end_sql</span><span class="p">,</span><span class="w"> </span><span class="s1">&#39;SCHEMA&#39;</span><span class="p">)</span><span class="o">.</span><span class="n">rows</span>
<span class="err">              SELECT a.attname, format_type(a.atttypid, a.atttypmod),</span>
<span class="err">                     pg_get_expr(d.adbin, d.adrelid), a.attnotnull, a.atttypid, a.atttypmod</span>
<span class="err">                FROM pg_attribute a LEFT JOIN pg_attrdef d</span>
<span class="w">              </span><span class="no">SELECT</span><span class="w"> </span><span class="n">a</span><span class="o">.</span><span class="n">attname</span><span class="p">,</span><span class="w"> </span><span class="n">format_type</span><span class="p">(</span><span class="n">a</span><span class="o">.</span><span class="n">atttypid</span><span class="p">,</span><span class="w"> </span><span class="n">a</span><span class="o">.</span><span class="n">atttypmod</span><span class="p">),</span>
<span class="w">                     </span><span class="n">pg_get_expr</span><span class="p">(</span><span class="n">d</span><span class="o">.</span><span class="n">adbin</span><span class="p">,</span><span class="w"> </span><span class="n">d</span><span class="o">.</span><span class="n">adrelid</span><span class="p">),</span><span class="w"> </span><span class="n">a</span><span class="o">.</span><span class="n">attnotnull</span><span class="p">,</span><span class="w"> </span><span class="n">a</span><span class="o">.</span><span class="n">atttypid</span><span class="p">,</span><span class="w"> </span><span class="n">a</span><span class="o">.</span><span class="n">atttypmod</span>
<span class="w">                </span><span class="no">FROM</span><span class="w"> </span><span class="n">pg_attribute</span><span class="w"> </span><span class="n">a</span><span class="w"> </span><span class="no">LEFT</span><span class="w"> </span><span class="no">JOIN</span><span class="w"> </span><span class="n">pg_attrdef</span><span class="w"> </span><span class="n">d</span>
```

To clarify, it's the SELECT statement that is contained in the output twice.

Source of the input:

https://github.com/rails/rails/blob/0f969a989c87872935ecdd4db5ab700e35226238/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L919C1-L925C59

Sanjays2402:
## Summary

Fixes #2998. An unterminated Ruby heredoc caused its body to be emitted **twice** in the lexer output — once as `Error` tokens and again re-lexed as ordinary Ruby.

## Reproduction

```python
from pygments.lexers.ruby import RubyLexer

# heredoc <<-end_sql is never terminated (truncated snippet)
src = (
    'exec_query(<<-end_sql)\n'
    '  SELECT a.attname\n'
    '  FROM pg_attribute a'
)
out = ''.join(v for _, v in RubyLexer().get_tokens(src))
print(out.count('SELECT a.attname'))   # -> 2  (before)   1  (after)
```

Before the fix the `SELECT ... / FROM ...` lines appear twice — matching the reporter's observation that a 7-line snippet renders as 10 lines with the `SELECT` block duplicated.

## Root cause

In `RubyLexer.heredoc_callback` (`pygments/lexers/ruby.py`), the outermost heredoc is resolved by scanning forward for each terminator:

```python
for match in line_re.finditer(ctx.text, ctx.pos):
    ...
    if check == hdname:
        ...
        ctx.pos = match.end()      # success: advance past consumed body
        break
    else:
        lines.append(match)
else:
    # end of heredoc not found -- error!
    for amatch in lines:
        yield amatch.start(), Error, amatch.group()
        # <-- ctx.pos is NOT advanced here
ctx.end = len(ctx.text)
```

The success branch advances `ctx.pos = match.end()`, but the "terminator not found" (`else`) branch yields the remaining lines as `Error` **without advancing `ctx.pos`**. Immediately after, `ctx.end = len(ctx.text)` re-opens the full text, so the `ExtendedRegexLexer` main loop resumes at the un-advanced `ctx.pos` and re-lexes the exact same lines a second time as normal Ruby — hence the duplication.

## Fix

Advance `ctx.pos` past the last consumed line in the error branch, mirroring what the success branch already does:

```python
else:
    # end of heredoc not found -- error!
    for amatch in lines:
        yield amatch.start(), Error, amatch.group()
    # advance past the consumed lines so they are not
    # re-lexed (and thus duplicated) once ctx.end is reset
    if lines:
        ctx.pos = lines[-1].end()
```

The unterminated body is now emitted exactly once, as `Error` tokens (the intended "unterminated heredoc" semantics).

## Testing

Added `test_unterminated_heredoc_not_duplicated` to `tests/test_ruby.py`. It asserts every input character is preserved exactly once (no duplication) and that the body is reported as `Error`.

Proof the test guards the bug (stash the source fix, keep the test):

| state | result |
|-------|--------|
| with fix | **PASS** |
| fix stashed (bug present) | **FAIL** — output contains the `SELECT`/`FROM` lines twice |

Full suite is green (the only excluded module, `tests/contrast`, fails to *collect* on a bare venv due to the optional `wcag_contrast_ratio` dependency — unrelated to this change):

```
5260 passed, 16 skipped
```

Ruby example-file golden-output tests all pass, and `ruff check` on the changed files is clean.

## Notes

- Diff is +33/−1 across `pygments/lexers/ruby.py`, `tests/test_ruby.py`, and a `CHANGES` entry.
- Verified the fix does not affect properly-terminated heredocs, multiple stacked heredocs, or trailing lines without a newline.


birkenfeld:
Nice, thanks for the PR!

PR #3172, opened 2026-07-02
TypeScript: require whitespace after `module` keyword (#2823)

### LINKED ISSUE #2823: `module` Is Incorrectly Highlighted as a Keyword When It Is Used at the Beginning of an Unquoted Object Key
In JavaScript and related languages like TypeScript, objects can use any string, including keywords of the language, as keys. A subset of keys can also be written without surrounding quotes. Following these rules, Pygments correctly handles the following object definition using the keyword `module` as a key, meaning `module` is highlighted like any other key.

```javascript
{ module: 0 }
```

However, if the key is unquoted and _starts with_ the word `module` followed by any other characters, `module` is interpreted as a keyword. The following code is highlighted as if `module` were a keyword.

```javascript
{ modules: 0 }
```

The produced tokens are:

| Type | Token |
| --- | --- |
| Punctuation | `{` |
| Text.Whitespace | ` ` |
| Keyword.Reserved | `module` |
| Name.Other | `s` |
| Operator | `:` |
| Text.Whitespace | ` ` |
| Literal.Number.Float | `0` |
| Text.Whitespace | ` ` |
| Punctuation | `}` |

Sanjays2402:
## Summary

Fixes #2823.

The `TypeScriptLexer` rule intended to match `module Foo {...}` namespace
declarations mis-tokenizes ordinary identifiers that merely *start with*
`module`, and the extremely common CommonJS `module.exports` idiom.

The rule is:

```python
# Match stuff like: module name {...}
(r'\b(module)(\s*)([\w?.$]+)(\s*)',
 bygroups(Keyword.Reserved, Whitespace, Name.Other, Whitespace), 'slashstartsregex'),
```

The whitespace group `(\s*)` allows **zero** characters between `module`
and the following identifier, so `module` is greedily consumed as
`Keyword.Reserved` whenever it is immediately followed by more word
characters or a dot.

## Reproduction (before)

```python
from pygments.lexers.javascript import TypeScriptLexer
lx = TypeScriptLexer()
for tok, val in lx.get_tokens("{ modules: 0 }"):
    if val.strip(): print(str(tok), repr(val))
```

| Input | Before | After |
| --- | --- | --- |
| `{ modules: 0 }` | `module` → `Keyword.Reserved`, `s` → `Name.Other` | `modules` → `Name.Other` |
| `module.exports = foo` | `module` → `Keyword.Reserved`, `.exports` → `Name.Other` | `module` → `Name.Other`, `.` → `Punctuation`, `exports` → `Name.Other` |
| `module Foo { }` | `module` → `Keyword.Reserved` | `module` → `Keyword.Reserved` (unchanged) |
| `module Foo.Bar { }` | `module` → `Keyword.Reserved` | `module` → `Keyword.Reserved` (unchanged) |

## Root cause & fix

In TypeScript the `module`/`namespace` contextual keyword must be
separated from the namespace name by whitespace — you cannot write
`moduleFoo` or `module.exports` as a declaration. Requiring at least one
whitespace character (`\s*` → `\s+`) makes non-declaration uses fall
through to the normal identifier rules, while genuine namespace
declarations (`module Foo {}`, `module Foo.Bar {}`, `module\nFoo {}`)
are unaffected.

```diff
-            (r'\b(module)(\s*)([\w?.$]+)(\s*)',
+            (r'\b(module)(\s+)([\w?.$]+)(\s*)',
```

## Tests

Adds a golden regression snippet `tests/snippets/typescript/2823.txt`
covering all three cases (`modules`, `module.exports`, `module Foo`).
Proven to guard the bug:

- With the source change **stashed**, the new snippet **fails** (emits
  `module` + `s` instead of `modules`).
- With the change applied, it **passes**.

Full local run: `3241 passed, 8 skipped` (entire `tests/snippets/` +
`tests/test_basic_api.py`), `ruff check pygments/lexers/javascript.py`
clean. Diff is +4/−1 in the lexer plus the new test and a CHANGES entry.


birkenfeld:
Looks good, thanks for the PR!

PR #3174, opened 2026-07-02
Improve scripts/get_vimkw.py

jvoisin:
The main change here lies in r_line, where a bunch of keywords were added. For convenience's sake, the conditions below it were re-ordered. Everything else is autogenerated.

I've been using
https://github.com/vim/vim/blob/v7.4.1588/runtime/syntax/vim.vim as reference.


jvoisin:
I'll likely try to bump the vim7.4 file one this is merged, as this version was released in August 2013

birkenfeld:
Sure, LGTM! Thanks!

PR #3177, opened 2026-07-02
Swift: keep `func` a keyword in `class func` methods

### LINKED ISSUE #1125: Swift Lexer: `func` mis-classified when following `class`
(Original issue 1421 created by renfeisong on 2018-03-02T08:52:37.919989+00:00)

Here is a snippet that illustrate the issue: [(view the same snippet on pygments.org)](http://pygments.org/demo/6718197/)

```
#!swift

func test1() {
    print("heh")
}

public func test2() {
    print("hah")
}

private func test3() {
    print("yo")
}

class func test4() {
    print("meh")
}

public class func test5() {
    print("nah")
}

private class func test6() {
    print("hmm")
}
```

## Expected behavior

The token `func` in `test{4,5,6}` should be treated as a keyword and thus have the same color as `func` in `test{1,2,3}`

## Actual behavior

The token `func` in `test{4,5,6}` are treated as a custom type name and colored as such. The error is more obvious when using a color scheme similar to that of Xcode (see attachment).

apoorva-01:
`class func`/`class var` are type-member modifiers, but the lexer's class-name rule grabbed the word after `class`, so `func` got tagged as a type name instead of a keyword. Fixed with a negative lookahead so those fall through to the keyword rules. `class Foo` still works.

Heads up on scope: I only excluded the modifier keywords instead of handling `class`/`static` modifiers properly, and the lookahead's on the whole `class|enum|...` group. Kept it to one rule, can go broader if you want.

Test added (fails before, passes after) + CHANGES.

Fixes #1125.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -34,6 +34,7 @@ Version 2.21.0
   * Kotlin: Don't let a nullable type marker (``?``) consume the following
     character, so ``Foo?,`` and ``a?:b`` tokenize correctly (#2964)
   * YAML: Add ``yml`` alias (#3169)
+  * Swift: Keep ``func`` a keyword in ``class func`` type methods (#1125)
 
 - Fix catastrophic backtracking in ``looks_like_xml`` (#2931, #3112)
 - Improve monokai theme colors (#3100)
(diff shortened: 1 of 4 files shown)


birkenfeld:
If `class func` and `class var` are the only ones to consider here, I think I'd prefer an explicit pattern handling those before the generic one, and not needing a lookahead assertion.

apoorva-01:
Makes sense, done. Dropped the lookahead and added explicit `class func`/`class var` rules ahead of the generic `class …` one, so the same token output falls out without the assertion. Also rebased on master to clear the CHANGES conflict.

birkenfeld:
Thanks, lgtm now!

PR #3180, opened 2026-07-05
Allow transparent in style color definitions

### LINKED ISSUE #3170: Can't set transparent color
`transparent` isn't a valid color according to the [`colorformat`](https://github.com/pygments/pygments/blob/2.20.0/pygments/style.py#L66-L79) function. Since it doesn't accept 4 or 8 character hex colors either, it can't set as `#0000` or `#00000000` either.

But at least it can be tricked by using `var(--some-variable,transparent)` since it doesn't really check the format of `var` or `calc` CSS functions.

ahmojo:
﻿## Summary

Fixes #3170 by allowing the CSS `transparent` keyword in style color definitions.

## What changed

- Allowed exact `transparent` style color values in `StyleMeta.colorformat()`.
- Preserved `transparent` when emitting HTML CSS instead of treating it as a hex value.
- Added a regression test covering foreground, background, and border style color output.
- No new dependencies.

## Why

Previously, defining a style such as `Name: "transparent"` failed with `AssertionError: wrong color format 'transparent'`. Since Pygments already passes through CSS-specific `var(...)` and `calc(...)` color values for HTML output, this keeps the change narrow by accepting only the `transparent` keyword rather than broadening validation to arbitrary named colors or RGBA formats.

## Tests

- [x] `tox -- tests/test_html_formatter.py::test_get_style_defs_allows_transparent_color`
- [x] `tox -- tests/test_html_formatter.py`
- [x] `tox -e check`
- [x] `tox`

Closes #3170

### DIFF
--- a/pygments/formatters/html.py
+++ b/pygments/formatters/html.py
@@ -27,7 +27,7 @@ __all__ = ['HtmlFormatter']
 
 
 def webify(color):
-    if color.startswith('calc') or color.startswith('var'):
+    if color == 'transparent' or color.startswith('calc') or color.startswith('var'):
         return color
     else:
         # Check if the color can be shortened from 6 to 3 characters
--- a/pygments/style.py
+++ b/pygments/style.py
@@ -74,6 +74,8 @@ class StyleMeta(type):
                     return col[0] * 2 + col[1] * 2 + col[2] * 2
             elif text == '':
                 return ''
+            elif text == 'transparent':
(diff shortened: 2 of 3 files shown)


birkenfeld:
Looks good to me. I don't quite know what will happen in non-HTML formatters, but since we're checking for `var` here as well, which is a CSS-only feature, I don't think we're making it worse.

PR #3183, opened 2026-07-05
Bump scripts/get_vimkw.py to vim9.2

jvoisin:
This commit bumps the vim syntax file used by scripts/get_vimkw.py to vim9.2, by removing a workaround (output_info['option'].append…), adding a handful of matchgrounps, and changing the vim.vim file path.

I used https://github.com/vim/vim/blob/v9.2.0782/runtime/syntax/vim.vim as reference.


birkenfeld:
:+1:

PR #3184, opened 2026-07-05
Speed up HtmlFormatter._format_lines per-token loop

jvoisin:
Skip the parts[:-1] slice for single-line tokens (the common case) by guarding the all-but-last-line loop with len(parts) > 1, hoist the span_element_openers / _translate_parts lookups to locals, and bind parts[-1] once.

Local microbenchmarks are saying ~16% faster formatting on a 50k-token corpus with byte-identical output, but only 2% faster for a full lex+format highlight as lexing dominates.


birkenfeld:
Why not!

PR #3185, opened 2026-07-06
Speed up HtmlFormatter token escaping with quote=False

jvoisin:
Token values are emitted as element text content (inside `<span>/<pre>`), where the " and ' characters do not need HTML-escaping. Escape them with html.escape(value, quote=False) directly instead of going through pygments.util.html_escape (which forces quote=True and adds a str() + extra call layer). This skips two of the five .replace() passes on the per-token hot path.

Interleaved, core-pinned benchmarking on pre-tokenized C/C++/Java/Python shows the full HtmlFormatter pipeline get ~7% faster, with every new sample beating every baseline sample.

The __init__ option escaping (cssclass, cssstyles, filename, lineseparator, lineanchors, linespans) keeps quote=True since those may end up inside HTML attributes.

Note: this is a visible output change: " and ' inside highlighted code now render as literal characters rather than &quot; / &#x27;. The result is equivalent, valid HTML (verified byte-for-byte across the example-file corpus: only quote characters differ, tag structure is unchanged). test_correct_output is updated to compare against quote=False escaping.

### DIFF
--- a/pygments/formatters/html.py
+++ b/pygments/formatters/html.py
@@ -13,6 +13,7 @@ import os
 import sys
 import os.path
 from io import StringIO
+from html import escape as _escape
 
 from pygments.formatter import Formatter
 from pygments.token import Token, Text, STANDARD_TYPES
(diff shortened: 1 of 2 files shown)


birkenfeld:
7%? That's pretty crazy :)

@Anteru any concerns about compatibility? I'm kinda happy to emit raw " and ' since we nowhere guarantee those to be escaped.

Anteru:
I don't think escaping those quotes was ever intentional. I wouldn't mind having a small unit test for this though, just to document the new behavior.

jvoisin:
Sure :)

PR #3186, opened 2026-07-06
Simplify redundant whitespace regexes to \s+

jvoisin:
Several lexers matched whitespace with a nested/redundant quantifier over an alternation, e.g. (\n|\s+)+, (\n|\s)+ or (\s|\n)+. Since \s already matches \n in re.MULTILINE, the \n branch, the outer quantifier and the capturing group are all redundant: each pattern is equivalent to plain \s+, but does more work (an alternation, a group and a needless second quantifier) both when matching a whitespace run and when failing at a non-whitespace position.

Replace them with \s+ in a bunch of lexers. The rules yield a bare Whitespace token and never read the group (no bygroups), so tokenisation is unchanged.

On dumb local microbenchmark (import re, timeit; re.compile; match), it speeds the matching by 25%-33%.

### DIFF
--- a/pygments/lexers/asm.py
+++ b/pygments/lexers/asm.py
@@ -285,7 +285,7 @@ class HsailLexer(RegexLexer):
             (r'[=<>{}\[\]()*.,:;!]|x\b', Punctuation)
         ],
         'whitespace': [
-            (r'(\n|\s)+', Whitespace),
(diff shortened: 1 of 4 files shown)


birkenfeld:
Yup, looks like a no brainer.

PR #3187, opened 2026-07-06
Speed up multiline comment matching with [\s\S]*?

jvoisin:
The Rell and Cheetah lexers matched multiline comment bodies with (.|\n)*? (Rell also had a redundant \r branch). Since . does not match \n under re.MULTILINE, this alternation is the usual idiom for "any character including newlines", but it is evaluated per character across the whole comment. [\s\S]*? is equivalent (\s and \S together cover every character) and matches via a single character class instead of an alternation.

The rules yield a bare Comment token and never read the group (no bygroups, no backref), so tokenisation is unchanged.

Microbenchmark: matching a long /* ... */ comment is ~4x faster; on a comment-heavy source the end-to-end lexing speedup is ~13%.

### DIFF
--- a/pygments/lexers/ampl.py
+++ b/pygments/lexers/ampl.py
@@ -30,7 +30,7 @@ class AmplLexer(RegexLexer):
             (r'\n', Text),
             (r'\s+', Whitespace),
             (r'#.*?\n', Comment.Single),
-            (r'/[*](.|\n)*?[*]/', Comment.Multiline),
+            (r'/[*][\s\S]*?[*]/', Comment.Multiline),
             (words((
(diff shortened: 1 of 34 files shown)


birkenfeld:
Are those really the only two occurrences?

jvoisin:
For the other ones, the performance changes were in the noise level.

birkenfeld:
Might still be worth being consistent?

jvoisin:
I chose minimal invasiveness over consistency, but happy to change them all :)

jvoisin:
Done!

birkenfeld:
LGTM, thanks!

PR #3188, opened 2026-07-06
Speed up line-scanning regexes with [^\n]* instead of lazy .*?

jvoisin:
Many lexers match to end-of-line with the lazy idioms .*?\n / .*?$ (line comments, line continuations, etc.). Lazy .*? advances one character at a time, whereas [^\n]* consumes the line in a single character-class scan. In isolation on microbenchmarks [^\n]*\n is ~3.4x faster than //.*?\n, and on comment-heavy input the affected lexers (zig, ada, prolog, hare, yara, ...) lex ~20% faster end to end.

Under re.MULTILINE without re.DOTALL (the RegexLexer default), . never matches \n, so .*?\n and .*?$ are exactly equivalent to [^\n]*\n and [^\n]*$. Lexers using re.DOTALL/re.S are left untouched.

This was a mechanical change, then tediously manually reviewed, and the testsuite wasn't touched, so it shouldn't™ break anything.

### DIFF
--- a/pygments/lexers/ada.py
+++ b/pygments/lexers/ada.py
@@ -36,7 +36,7 @@ class AdaLexer(RegexLexer):
     tokens = {
         'root': [
             (r'[^\S\n]+', Text),
-            (r'--.*?\n', Comment.Single),
+            (r'--[^\n]*\n', Comment.Single),
             (r'[^\S\n]+', Text),
             (r'function|procedure|entry', Keyword.Declaration, 'subprogram'),
             (r'(subtype|type)(\s+)(\w+)',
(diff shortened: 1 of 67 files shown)


jvoisin:
Seems like [regexlint](https://github.com/thatch/regexlint/blob/6f385043ca78d37679b77db6825f1f1c70dce805/regexlint/checkers.py#L59-L71) isn't happy

birkenfeld:
You have an actual newline in the regex (which apparently Jythom didn't support), probably should be `\\n`

jvoisin:
Seems like my horrible vim-search'n'replace did have some misfirings indeed. I added `r` prefixes instead

birkenfeld:
Just to be sure - these lexers all don't have re.DOTALL in the flags?

jvoisin:
> Just to be sure - these lexers all don't have re.DOTALL in the flags?

They don't have re.DOTALL in the flags, no

PR #3189, opened 2026-07-06
Use words() for large keyword lists in a few lexers

jvoisin:
Several lexers matched large keyword/builtin sets with a hand-written (a|b|c|...) alternation. words() compiles the same set through regex_opt, which factors shared prefixes into a trie: faster to match and, more importantly, faster to reject.

This commit converts the lexers with ~40 words or more, and since regex_opt preserves the matched set tokenisation is unchanged.

The benefit scales with list size: on keyword-heavy input, an isolated within-process benchmark shows the keyword rule get ~8-14% faster for these lexers (console +13.6%, C# +10.3%, Modelica +7.7%). Smaller lists gain little, so only the large ones are converted here.


jvoisin:
```
/home/runner/work/pygments/pygments/pygments/lexers/dotnet.py: (BooLexer:root:pat#11) W117: Overlap in character class: ['s']
  '(a(?:bstract|nd|[ss])|break|c(?:a(?:llable|st)|on(?:...
                   ^ here
```

now that's a fun one, seems like a bug in the `words()` implementation

jvoisin:
#3190 fixes the issue, and I've removed the duplicate `as` from the Boo parser

birkenfeld:
Nice!

PR #3190, opened 2026-07-07
Fix regex_opt emitting redundant character classes for duplicate words

jvoisin:
regex_opt() sorted its input but never deduplicated it. When a word list contained duplicates (e.g. the Boo lexer keyword list has 'as' twice), the duplicate one-character remainders left after prefix-stripping were fed into make_charset, producing a redundant character class such as [ss]. This triggered a W117 "Overlap in character class" lint warning (surfaced in pygments/pygments#3189).

Deduplicate the input with sorted(set(strings)) and add regression tests.

### DIFF
--- a/pygments/regexopt.py
+++ b/pygments/regexopt.py
@@ -98,5 +98,5 @@ def regex_opt(strings, prefix='', suffix=''):
 
     *prefix* and *suffix* are pre- and appended to the final regex.
     """
-    strings = sorted(strings)
+    strings = sorted(set(strings))
     return prefix + regex_opt_inner(strings, '(') + suffix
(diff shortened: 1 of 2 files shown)


Anteru:
Good catch, thanks!

birkenfeld:
I'm actually not sure this needed to be fixed here; having duplicate words in `words()` is a bug and should be fixed there, no? Sometimes it's really just a duplicated word but it could also be less benign, indicating corruption in the list of words.

jvoisin:
Maybe, I can see arguments for both sides.
If `word()` should detect duplicate keywords, it should break/yell in a nice and understandable way nevertheless, instead of a cryptic error :/

PR #3191, opened 2026-07-07
Remove redundant duplicate regexes in several lexers

jvoisin:
Each of these rules was an exact duplicate of an earlier pattern within the same RegexLexer state. They were found with the following python script:

```python
from pygments.lexers import get_all_lexers, find_lexer_class
from pygments.lexer import RegexLexer

report = []
seen_classes = set()

for name, aliases, filenames, mimetypes in get_all_lexers(plugins=False):
    cls = find_lexer_class(name)
    if cls is None or cls in seen_classes:
        continue
    if not (isinstance(cls, type) and issubclass(cls, RegexLexer)):
        continue
    seen_classes.add(cls)
    # look only at tokens literally defined on THIS class (not inherited)
    toks = cls.__dict__.get('tokens')
    if not toks:
        continue
    for state, rules in toks.items():
        seen = {}
        for idx, tdef in enumerate(rules):
            if not (type(tdef) is tuple):
                continue
            pat = tdef[0]
            if not isinstance(pat, str):
                continue
            if pat in seen:
                report.append((cls.__module__.split('.')[-1], cls.__name__,
                               state, pat, seen[pat], idx))
            else:
                seen[pat] = idx

for mod, clsname, state, pat, first, dup in report:
    print(f"{mod}.py {clsname} state={state!r} rule#{dup} dup of #{first}: {pat!r}")
print(f"\nTotal in-source duplicate regexes: {len(report)}")
```

### DIFF
--- a/pygments/lexers/ada.py
+++ b/pygments/lexers/ada.py
@@ -37,7 +37,6 @@ class AdaLexer(RegexLexer):
         'root': [
             (r'[^\S\n]+', Text),
             (r'--.*?\n', Comment.Single),
-            (r'[^\S\n]+', Text),
             (r'function|procedure|entry', Keyword.Declaration, 'subprogram'),
             (r'(subtype|type)(\s+)(\w+)',
              bygroups(Keyword.Declaration, Text, Keyword.Type), 'type_def'),
--- a/pygments/lexers/basic.py
+++ b/pygments/lexers/basic.py
@@ -457,7 +457,6 @@ class QBasicLexer(RegexLexer):
             (r'\-?\d*\.\d+[@|#]?', Number.Float),
             (r'\-?\d+[@|#]', Number.Float),
             (r'\-?\d+#?', Number.Integer.Long),
-            (r'\-?\d+#?', Number.Integer),
             (r'!=|==|:=|\.=|<<|>>|[-~+/\\*%=<>&^|?:!.]', Operator),
             (r'[\[\]{}(),;]', Punctuation),
             (r'[\w]+', Name.Variable.Global),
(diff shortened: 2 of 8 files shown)


Anteru:
I really love all of those changes you're doing here for performance, but can we please make those things tests that run during `make check` so we're not going to introduce things again? It's fantastic that you have the script, but if it's not hooked up to run regularly, you're only getting 1/2 of the value of your investigation.

birkenfeld:
I agree! For now, we can merge this of course.

birkenfeld:
(Some of the checks should probably be part of regexlint, like looking for `(.|\n)` or `.*?\n` or `.*?$`)

jvoisin:
It seems like regexplint is a tad unmaintained, and doesn't work on modern python :/

Anteru:
We maintain a copy of regexlint here: https://github.com/pygments/regexlint You can contribute this there.

jvoisin:
My changes [were surprisingly merged upstream](https://github.com/thatch/regexlint/commits/master/), yay. Are you planning on backporting/integrating upstream's changes, or shall I send pull-requests against your fork?

Anteru:
I'll try to update our fork, let me see how many conflicts that produces.

Anteru:
Updated our fork to match upstream again.

jvoisin:
Would it make sense to ask @thatch if he's willing to make the pygments' fork the upstream one, and delete his?

thatch:
Happy to, let's merge the regexlint PRs you've opened recently first and I'll sort out rename/pypi creds/trusted publisher for it.

Anteru:
Thanks @thatch . We'll cut a release for now then we can look into that next.

PR #3192, opened 2026-07-07
Simplify redundant regex lookarounds in lexers

jvoisin:
Replace zero-width lookarounds that wrap a single boundary/anchor assertion with the equivalent assertion itself:

- email.py: (?<=\b)/(?=\b) -> \b and (?!\b) -> \B across the 15 IPv4/IPv6 rules
- templates.py: (?<=^) -> ^ in the Myghty and Mason comment/line rules
- webmisc.py: (?=\b) -> \b in the XQuery operator-word rule

These are exact equivalents: a lookaround over a zero-width assertion is the assertion.

This was found by running `git grep -E '\(\?<?=(\\b|\^|\$)\)'`


birkenfeld:
Good catch!

PR #3195, opened 2026-07-08
Ruby: treat << after ), ], } as the shift operator, not a heredoc

### LINKED ISSUE #760: RubyLexer: overeager heredoc
(Original issue 1053 created by hynkle-taxjar on 2014-11-05T20:02:57.242083+00:00)

When you give RubyLexer `[]<<a`, `<<a` is misinterpreted as the beginning of a heredoc (which of course causes errors in subsequent lines).

Here's a failing test for it:
```python
def testShiftLeftWithoutSpace(self):
    fragment = u'[]<<a\n'
    tokens = [
        (Token.Operator, u'['),
        (Token.Operator, u']'),
        (Token.Operator, u'<<'),
        (Token.Name,     u'a'),
        (Token.Text,     u'\n'),
    ]
    self.assertEqual(tokens, list(self.lexer.get_tokens(fragment)))
```

AmadNaseem:
Fixes #760.

The bareword heredoc rule in `RubyLexer` used a `(?<!\w)` lookbehind. That rejects `a<<b` (word char before `<<`), but `)`, `]` and `}` are not word characters, so `<<` immediately after a closing bracket was wrongly started as a heredoc. After a closing `)`/`]`/`}` the parser is in an expression-end state, so `<<` there is the left-shift/append operator.

Example — `[]<<a` before:

```
Operator '['
Operator ']'
Operator '<<'
String.Heredoc ''
String.Delimiter 'a'
String.Heredoc ''
```

after:

```
Operator '['
Operator ']'
Operator '<<'
Name 'a'
```

The fix extends the lookbehind to `(?<![\w)\]}])` on the "normal heredocs" rule only. Verified that genuine heredocs are unaffected, including ones that legitimately follow `(` or `[` (e.g. `foo(<<A)`, `[<<A]`), as well as `<<~` / `<<-` forms.

Added a snippet test `tests/snippets/ruby/test_shift_operator.txt` and a `CHANGES` entry.

Scope: this targets the `)`/`]`/`}` case from the issue. Two related cases are intentionally left for separate follow-ups — `<<` after a string literal (`"foo"<<bar`) and the empty-delimiter `<<""` heredoc rule — since they involve different rules and tradeoffs.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -31,6 +31,8 @@ Version 2.21.0
     whitespace, so identifiers like ``modules`` and ``module.exports``
     are highlighted correctly (#2823)
   * Ruby: Don't duplicate the body of an unterminated heredoc (#2998)
+  * Ruby: Treat ``<<`` after ``)``, ``]`` or ``}`` as the shift operator
+    rather than the start of a heredoc (#760)
   * Kotlin: Don't let a nullable type marker (``?``) consume the following
     character, so ``Foo?,`` and ``a?:b`` tokenize correctly (#2964)
   * YAML: Add ``yml`` alias (#3169)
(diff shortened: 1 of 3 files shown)


birkenfeld:
Makes sense. Thanks!

Issue #3077, opened 2026-03-30
Superfence warning with pygments 2.20.0


afroch:
Hi!

We are using `mkdocs` to produce our documentation, which includes a bit of code. We usually don't fix the pygments version, but with 2.20.0, we get the following warning (which kills our strict build of our docs):

```bash
WARNING -  mkdocs_autorefs: software/tutorials/tutorial-upp.md: Could not find cross-reference target ':'
WARNING -  mkdocs_autorefs: software/tutorials/tutorial-upp.md: Could not find cross-reference target '"pt_btagJes"'
WARNING -  mkdocs_autorefs: software/tutorials/tutorial-upp.md: Could not find cross-reference target '"pt_btagJes"'
WARNING -  mkdocs_autorefs: software/tutorials/tutorial-upp.md: Could not find cross-reference target '"pt_btagJes"'
```

The code markdown is the following:

```markdown
??? warning "Solution: Plot Input Distributions"

    You can plot for example the $p_\text{T}$ distribution with the following snippet.
    Note that you have to replace `<path_to_ttbar_h5_file>` with the actual path to the ttbar file you downloaded.

    ```python
    import h5py
    from puma import Histogram, HistogramPlot

    ttbar_filepath = "<path_to_ttbar_h5_file>"

    # load the jets dataset from the h5 file
    with h5py.File(ttbar_filepath, "r") as h5file:
        jets = h5file["jets"][:]

    # defining boolean arrays to select the different flavour classes
    is_light = jets["HadronConeExclTruthLabelID"] == 0
    is_c = jets["HadronConeExclTruthLabelID"] == 4
    is_b = jets["HadronConeExclTruthLabelID"] == 5

    # initialise the plot
    pt_plot = HistogramPlot(
        bins_range=(0, 250_000),
        xlabel="$p_T$ [MeV]",
        ylabel="Normalised number of jets",
    )

    # add the histograms
    pt_plot.add(Histogram(jets[is_light]["pt_btagJes"], flavour="ujets"))
    pt_plot.add(Histogram(jets[is_c]["pt_btagJes"], flavour="cjets"))
    pt_plot.add(Histogram(jets[is_b]["pt_btagJes"], flavour="bjets"))


    pt_plot.draw()
    pt_plot.savefig("tutorial_histogram_pT.png")
    ```
```

It seems that somehow the admonition + superfence breaks now mkdocs_autorefs.

[...]

PR #3090, opened 2026-04-06
Update #3078 - Force any non-NoneType to string

styx0x6:
Hi @Anteru , @birkenfeld ,

Following up #3078 (which fixes #3076), I would like to improve a quite bite the `html_escape` function for the 2.20.1 release.

@Anteru you said in #3078:

> Eventually, we can even add strong types to it (str | None) in this case.

I found this issue recently that would justify strong types : https://github.com/lepture/shibuya/issues/113

This time an `int` type is initiated. Some attributes from `HtmlFormatter` definitely shouldn't be inititiated with other types than string (right?), but I suggest, as you mentioned it above @Anteru , to force strong type `str` when it's not `NoneType`, then no more of this kind of issue may be raised.

PR #3196, opened 2026-07-08
Remove redundant validation on already escaped filename field

styx0x6:
Hi,

Just a quick PR to address my comment on closed issue https://github.com/pygments/pygments/issues/3077#issuecomment-4380341716.

As a reminder , the comment:

> Hi all,
> 
> Sorry to raise again this topic, just read it now, but I think there is an overlap between #3076 and #3077, and their resolutions.
> 
> @pellepim, the solution in PR #3090 wasn't merged at the time of your analysis. So, regarding PR #3016, ` or '')` in `self.filename = html.escape(self._decodeifneeded(options.get('filename') or ''))` will work but is redondant to the `html_escape` solution and respond only to the filename issue, not to other escaped fields. Also, it will not handle other type of values, like `int` (see lepture/shibuya/issues/113). I did the same mistake in #3078, amended by PR #3090 to address it.
> 
> I use zensical rather than mkdocs, but I am fairly sure that this is the same thing. superfences swallows the exception silently and the raw backtick markup passes through to the Markdown block parser, which renders it as inline < code > with the language hint visible as text.
> 
> Correct and a good catch, didn't saw this relationship.
> 
> So #3090 will handle it in this case.
> 
> **Proposals to fix the redondant fix** @Anteru , @birkenfeld :
> - Revert commit b6f6dab78e13652741d5e5811b4256b4dcb5fcd8. Will also remove the test.
> - A new PR to remove ` or '')` on escaped filename field. We can let the test which is correct.
> 
> Best regards. 

 _Originally posted by @styx0x6 in [#3077](https://github.com/pygments/pygments/issues/3077#issuecomment-4380341716)_

### DIFF
--- a/pygments/formatters/html.py
+++ b/pygments/formatters/html.py
@@ -416,7 +416,7 @@ class HtmlFormatter(Formatter):
         self.noclobber_cssfile = get_bool_opt(options, 'noclobber_cssfile', False)
         self.tagsfile = self._decodeifneeded(options.get('tagsfile', ''))
         self.tagurlformat = self._decodeifneeded(options.get('tagurlformat', ''))
-        self.filename = html_escape(self._decodeifneeded(options.get('filename', '') or ''))
+        self.filename = html_escape(self._decodeifneeded(options.get('filename', '')))
         self.wrapcode = get_bool_opt(options, 'wrapcode', False)
         self.span_element_openers = {}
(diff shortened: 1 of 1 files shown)


styx0x6:
Fixes https://github.com/pygments/pygments/issues/3077 (to link the related issue to this PR).

pellepim:
:+1:

PR #3197, opened 2026-07-09
Mathematica: recognize \[Name] named-character escapes

### LINKED ISSUE #3097: Mathematica parsing warnings since 2.20
Hi pygments team!

Since the update to 2.20, the build of our documentation using Sphinx is generating warnings that were not here in 2.19.2.

The `.rst` snippet of code is as follow

```.rst
.. code:: wl

   (* [command] *)

   gap = 7/4 + flagNLOSO*(L^2*(-15/32 - (45*\[Nu])/32))/r^2
```

and this raises the following warning:

```
WARNING: Lexing literal_block '(* [command] *)\n\ngap = 7/4 + flagNLOSO*(L^2*(-15/32 - (45*\\[Nu])/32))/r^2' as "wl" resulted in an error at token: '\\'. Retrying in relaxed mode. [misc.highlighting_failure]
```

The only location I am seeing a `\\` is for `\\[Nu]` which comes from `\[Nu]` on the original input. The other package we are using is [pygments-mathematica](https://github.com/rsmenon/pygments-mathematica). I believe this comes from `pygments` as the failures started with the release date of 2.20, but I might be completely wrong, please let me know (using sphinx 9.0.4).

Thank you!

apoorvdarshan:
Fixes #3097.

## Summary

Since 2.20, highlighting valid Mathematica/Wolfram code that contains a named-character escape such as `\[Nu]` produces a `Token.Error`, which surfaces as a Sphinx `misc.highlighting_failure` warning. For example, lexing `(45*\[Nu])/32` yielded `(Token.Error, \)` followed by `[`, `Nu`, `]` tokenized separately.

The `MathematicaLexer` (in `pygments/lexers/algebra.py`) had no rule for the `\[Name]` named-character escape syntax, so the leading backslash was left unmatched and emitted as an `Error` token.

## Fix

Add a rule to the lexer's `root` state that matches `\[Name]` named-character escapes (e.g. `\[Nu]`, `\[CapitalAlpha]`) and tokenizes each as a single `String.Escape` token instead of an `Error`.

```python
# Named character escapes, e.g. \[Nu], \[CapitalAlpha].
(r'\\\[[A-Za-z][A-Za-z0-9]*\]', String.Escape),
```

## Tests

- Added `tests/snippets/mathematica/test_named_character.txt`, a golden snippet test for the issue's reproducer `(45*\[Nu])/32`. It fails on the pristine lexer (records `'\\' Error`) and passes with the fix (`'\\[Nu]' Literal.String.Escape`).
- Full golden suite (`tests/snippets/` + `tests/examplefiles/`) and the rest of `tests/` pass; `ruff check --ignore UP031 .`, `scripts/check_sources.py`, and `regexlint` are clean.
- Added a `CHANGES` entry under "Updated lexers".

Disclosure: prepared with AI assistance; reviewed and verified locally.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -37,6 +37,8 @@ Version 2.21.0
     character, so ``Foo?,`` and ``a?:b`` tokenize correctly (#2964)
   * YAML: Add ``yml`` alias (#3169)
   * Swift: Keep ``func`` a keyword in ``class func`` type methods (#1125)
+  * Mathematica: Recognize ``\[Name]`` named-character escapes such as
+    ``\[Nu]`` instead of emitting an ``Error`` token (#3097)
 
 - Fix catastrophic backtracking in ``looks_like_xml`` (#2931, #3112)
 - Improve monokai theme colors (#3100)
--- a/pygments/lexers/algebra.py
+++ b/pygments/lexers/algebra.py
@@ -177,6 +177,9 @@ class MathematicaLexer(RegexLexer):
         'root': [
             (r'(?s)\(\*.*?\*\)', Comment),
 
+            # Named character escapes, e.g. \[Nu], \[CapitalAlpha].
+            (r'\\\[[A-Za-z][A-Za-z0-9]*\]', String.Escape),
+
             (r'([a-zA-Z]+[A-Za-z0-9]*`)', Name.Namespace),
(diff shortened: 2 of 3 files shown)


birkenfeld:
Why "since 2.20“? Is this a regression?

birkenfeld:
In any case, lgtm.

PR #3199, opened 2026-07-12
Go lexer improvements

arminguenther:
Fixes following issues:
- Go has no `float` type, just [`float32` and `float64`](https://go.dev/ref/spec#Numeric_types)
- Expressions like `int64(x)` are [__type__ conversions](https://go.dev/ref/spec#Conversions)
- [Maps](https://go.dev/ref/spec#Map_types) and [channels](https://go.dev/ref/spec#Channel_types) should be seen as types
- [Octal literals](https://go.dev/ref/spec#Integer_literals) may include letter `o` or `O` after their leading zero
- Go has [binary literals](https://go.dev/ref/spec#Integer_literals) starting with `0b` or `0B`
- [Integer literals](https://go.dev/ref/spec#Integer_literals) can contain `_` separators
- `!`, `<`, `>`, `|`, and `^` are [operators](https://go.dev/ref/spec#Operators), `=` an [assignment operator](https://go.dev/ref/spec#Assignment_statements) rather than punctuation
- [Assignment operator](https://go.dev/ref/spec#Assignment_statements) `^=` was missing
- `...` and `~` are no operators but [punctuation](https://go.dev/ref/spec#Operators_and_punctuation)

Notes:
- Floating-point and imaginary literals can also contain `_` characters and can use hexadecimal notation, may add this in another pull request


birkenfeld:
Thanks for the PR! Looks like the test case doesn't exercise some of the changes yet - can you amend it?

PR #3200, opened 2026-07-12
[Bug] Fix hexdump regex to only match valid hex digits

### LINKED ISSUE #2847: Why are the letters `g`,` h`, `G`, and `H` allowed in hexdump regex?
Line 42 of `pygments/lexers/hexdump.py` reads as follows:

`hd = r'[0-9A-Ha-h]'`

Why are the letters `g`, `h`, `G`, and `H` allowed? As far as I understand, this is supposed to be a regular expression that only matches hexadecimal numbers. But characters allowed for hexadecimal numbers are `0123456789abcdefABCDEF`. So why are additional letters allowed here? Does it have something to do with a specific output format of any of the supported hexdump programmes? If not, the line should probably be changed to:

`hd = r'[0-9A-Fa-f]'`

hiro-nikaitou:
Fixes the character range in the hexdump lexer regex from `[0-9A-Ha-h]` to `[0-9A-Fa-f]`.

The letters G, H, g, h are not valid hexadecimal digits and were incorrectly included in the pattern.

Fixes #2847

### DIFF
--- a/pygments/lexers/hexdump.py
+++ b/pygments/lexers/hexdump.py
@@ -39,7 +39,7 @@ class HexdumpLexer(RegexLexer):
     url = 'https://en.wikipedia.org/wiki/Hex_dump'
     version_added = '2.1'
 
-    hd = r'[0-9A-Ha-h]'
+    hd = r'[0-9A-Fa-f]'
 
     tokens = {
         'root': [


birkenfeld:
Fun :)

PR #3204, opened 2026-07-12
TeX: recognize math environments (#3034)

### LINKED ISSUE #3034: Latex math environments not shown as math
When using an environment whose inside is math-mode, it's content is not treated differently (at least not as math-mode).

Example:
https://en.wikipedia.org/wiki/User:Joshinils/tex-pygments#latex,_no_math-tag
where the environments _align_, _equation_, _math_ and _displaymath_ are shown.

Sanjays2402:
LaTeX math environments such as `equation` and `align` were lexed as ordinary text, unlike `$...$` and `\[...\]`. Enter a math state for the standard `math`, `displaymath`, `equation`, and `align` environments (including starred forms), with regression coverage.

Fixes #3034.


birkenfeld:
Thanks for the PR!

PR #3206, opened 2026-07-13
Vala: recognize verbatim strings

### LINKED ISSUE #2861: Vala: verbatim are causing errors
```
string verbatim = """This is a so-called "verbatim string".
   Verbatim strings don't process escape sequences, such as \n, \t, \\, etc.
   They may contain quotes and may span multiple lines.""";
```
Verbatim strings aren't recognized and are causing various errors: https://pygments.org/demo/?lexer=vala&style=bw&formatter=tokens&code=string+verbatim+%3D+%22%22%22This+is+a+so-called+%22verbatim+string%22.%0A+++Verbatim+strings+don%27t+process+escape+sequences%2C+such+as+%5Cn%2C+%5Ct%2C+%5C%5C%2C+etc.%0A+++They+may+contain+quotes+and+may+span+multiple+lines.%22%22%22%3B

Sanjays2402:
Vala's ordinary quote rule matched the first `"` of a triple-quoted verbatim string, making its contents fall back to ordinary code tokens and producing errors for valid backslashes. Match the existing verbatim-string rule first so the full multiline value is emitted as one string token; the reported example is covered by a golden test.

Fixes #2861

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -39,6 +39,7 @@ Version 2.21.0
   * Swift: Keep ``func`` a keyword in ``class func`` type methods (#1125)
   * Mathematica: Recognize ``\[Name]`` named-character escapes such as
     ``\[Nu]`` instead of emitting an ``Error`` token (#3097)
+  * Vala: Recognize verbatim strings before ordinary quoted strings (#2861)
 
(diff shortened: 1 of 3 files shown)


birkenfeld:
Good catch!

PR #3209, opened 2026-07-13
YAML: recognize non-breaking spaces

### LINKED ISSUE #2827: YAML lexer has trouble with non-breaking spaces
According to the [YAML spec](https://yaml.org/spec/1.2.2/#51-character-set), non-breaking spaces are printable characters which don't need escaping. However, `pygments` doesn't seem to support them. For example, a plain scalar containing a non-breaking space produces `Token.Error`:

```console
$ printf 'foo: bar\u00a0baz' | pygmentize -l yaml -f raw
Token.Name.Tag	'foo'
Token.Punctuation	':'
Token.Text.Whitespace	' '
Token.Literal.Scalar.Plain	'bar'
Token.Error	'\xa0'
Token.Literal.Scalar.Plain	'baz'
Token.Text.Whitespace	'\n'
```

Sanjays2402:
YAML permits U+00A0 in scalar content, but the lexer emitted an `Error` token for it. This recognizes non-breaking spaces in block, flow, and quoted scalars and adds a golden regression test.

Fixes #2827

### DIFF
--- a/pygments/lexers/data.py
+++ b/pygments/lexers/data.py
@@ -349,10 +349,8 @@ class YamlLexer(ExtendedRegexLexer):
             (r'[ ]+$', Whitespace),
             # line breaks are ignored
             (r'\n+', Whitespace),
-            # non-breaking spaces are a part of the value
-            (r'\xa0+', Name.Variable),
             # other whitespaces are a part of the value
-            (r'[ ]+', Name.Variable),
+            (r'[ \xa0]+', Name.Variable),
         ],
 
         # single-quoted scalars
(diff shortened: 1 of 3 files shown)


pygments/lexers/data.py
```diff
@@ -349,6 +349,8 @@ def callback(lexer, match, context):
             (r'[ ]+$', Whitespace),
             # line breaks are ignored
             (r'\n+', Whitespace),
+            # non-breaking spaces are a part of the value
+            (r'\xa0+', Name.Variable),
```
birkenfeld: Can we combine this with the rule below?

birkenfeld (changes requested):

Sanjays2402:
@birkenfeld Combined the adjacent quoted-scalar rules in `c45b3c1f`; regular and non-breaking spaces now share one `[ \\xa0]+` expression with the same `Name.Variable` action.

Verification: the YAML regression passed, and the full suite is green: **5,296 passed, 16 skipped**.

PR #3211, opened 2026-07-13
Kusto: highlight object member access (#2779)

### LINKED ISSUE #2779: Kusto: Error when using objects
pygments is reporting an error at ``tags[.]owner`` for the `.` but this is a valid syntax for kusto.

```kusto
    resources
    | project id, type, name, resourceGroup, location, owner=tags.owner
```

see https://pygments.org/demo/?lexer=kql&style=bw&formatter=html&code=++++resources%0A++++%7C+project+id%2C+type%2C+name%2C+resourceGroup%2C+location%2C+owner%3Dtags.owner%0A

Sanjays2402:
Kusto member access such as `tags.owner` currently emits an `Error` token for the dot because `.` is missing from the punctuation table. This adds both dot punctuation forms, preserves leading-dot floats by matching numbers first, and adds a golden regression case.

Fixes #2779

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -39,6 +39,7 @@ Version 2.21.0
   * Swift: Keep ``func`` a keyword in ``class func`` type methods (#1125)
   * Mathematica: Recognize ``\[Name]`` named-character escapes such as
     ``\[Nu]`` instead of emitting an ``Error`` token (#3097)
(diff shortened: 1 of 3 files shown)


birkenfeld:
Thanks!

PR #3213, opened 2026-07-13
C#: recognize interpolated verbatim strings (#2685)

### LINKED ISSUE #2685: C#: $@"foo"
```c#
Console.WriteLine($@"C:\Users\{username}\Desktop");
```

the `$` shouldn't be highlighted as an error

Sanjays2402:
C# interpolated verbatim strings allow both `$@"..."` and `@$"..."`, but the lexer left one prefix character as an `Error` token. Recognize both forms as strings and cover the reported examples with a golden lexer test.

Fixes #2685


birkenfeld:
Nice!

PR #3216, opened 2026-07-14
Clojure: recognize named, octal and unicode character literals as a single token

### LINKED ISSUE #979: incorrect highlighting for clojure \space character
(Original issue 1275 created by jcorneli on 2016-08-12T14:27:09.060389+00:00)

I've put an example here, and below.
http://pygments.org/demo/5863783/

```
#!clojure

(defn test-string-is-word
  "Allow a string that represents an word, defined by not containing spaces."
  [x]
  {:pre [(string? x)]
   :post [(boolean? %)]}
  (if (some #(= \space %) (seq x))
    false
    true))
```

For comparison, here is how Emacs highlights it:

AmadNaseem:
Fixes #979.

`ClojureLexer` splits named, octal, and unicode character literals into two tokens. For example `\space` lexes as `String.Char '\s'` followed by `Name.Variable 'pace'`, instead of a single character literal.

The cause is the rule `(r"\(.|[a-z]+)", String.Char)`: the `.` alternative always matches first, so the `[a-z]+` branch is dead and multi-character names are never matched as a unit.

The fix replaces it with an ordered alternation covering the six named literals (`newline`, `space`, `tab`, `formfeed`, `backspace`, `return`), unicode `\uNNNN` (4 hex digits), octal `\oNNN` (1–3 octal digits), and finally the single-character fallback:

```python
(r"\(?:newline|space|tab|formfeed|backspace|return"
 r"|u[0-9a-fA-F]{4}|o[0-7]{1,3}|.)", String.Char),
```

Now each of these lexes as a single `String.Char`:

```
\a  \A  \1  \(  \space  \newline  \tab  \formfeed  \backspace  \return  ÿ  \o377
```

Single-character literals are unchanged, and literals stop correctly at delimiters (e.g. `\newline)` → `\newline` + `)`).

Added a snippet test (`tests/snippets/clojure/test_characters.txt`) and a `CHANGES` entry. The existing Clojure example-file goldens are unchanged.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -39,6 +39,8 @@ Version 2.21.0
   * Swift: Keep ``func`` a keyword in ``class func`` type methods (#1125)
   * Mathematica: Recognize ``\[Name]`` named-character escapes such as
     ``\[Nu]`` instead of emitting an ``Error`` token (#3097)
+  * Clojure: Recognize named, octal and unicode character literals such as
+    ``\space`` and ``\o377`` as a single token (#979)
 
 - Fix catastrophic backtracking in ``looks_like_xml`` (#2931, #3112)
 - Improve monokai theme colors (#3100)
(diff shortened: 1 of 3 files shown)


birkenfeld:
Thanks for the PR!

PR #3218, opened 2026-07-15
Go: Full float, imaginary, rune and string literals

arminguenther:
Fixes the following issues:
- [Float literals](https://go.dev/ref/spec#Floating-point_literals) may contain `_` characters as separators
- Float literals may use hex notation
- In float literals, `.` can be immediately followed by the exponent character
- [Imaginary literals](https://go.dev/ref/spec#Imaginary_literals) may contain `_` characters as separators
- Imaginary literals may use binary, octal or hex notation
- Escape sequences in [rune](https://go.dev/ref/spec#Rune_literals) and [string](https://go.dev/ref/spec#String_literals) literals should be recognized as such
- Octal byte values like `\377` in runes and strings consist of exactly three octal digits
- `'\"'` is not a valid rune
- Raw string literals `` `...` `` can be classified as `String.Backtick` tokens


arminguenther:
Code `mi` is not available for `Number.Imag`, already taken by `Number.Integer`. I opted for `mc` as in **c**omplex. Not sure if there is some better code.

birkenfeld:
Are you sure it's worth giving `Number.Imag` its own token type? And even if you do, there is no need to add it to the short token abbreviation - as long as no other lexer (or style) is using it, there's no good reason to add this.

arminguenther:
Maybe we can treat imaginary literals as two parts like described in the spec:

> The value of an imaginary literal is the value of the respective integer or floating-point literal multiplied by the imaginary unit _i_

So `i` could be its own token, maybe `Name.Builtin`?
Is `Number.Float.Hex` okay?

arminguenther:
I also added some improvements to rune and string literals, let me know what you think :)

PR #3221, opened 2026-07-16
Simplify single-character regex alternations to character classes

jvoisin:
Replace single-character alternations like (a|b) and (?:a|b) with the equivalent character class [ab] across several lexers.

While technically a character class compiles to a single IN op in CPython's sre engine, whereas an alternation compiles to a slower BRANCH, there is no measurable performance improvements, even from removing the useless capture groups. The change is purely cosmetic, as reading regexps is tedious enough as it is without the need to add useless grouping.

This was found via `git grep '(.|.)' | grep -v bygroup`


Anteru:
Merged, thanks!

PR #3228, opened 2026-07-22
Fix catastrophic backtracking in string and regex lexers

Support-itecz:
Several string and regex rules build their body from an alternation like `(\\\\|\\"|[^"])*`. The `[^"]` branch also matches a backslash, so a run of backslashes can be split between it and the `\\\\` branch in exponentially many ways, and an unterminated string of backslashes makes the regex engine try every split before it fails. A few dozen backslashes already cost seconds on the affected lexers, and each extra character roughly doubles that, so any untrusted snippet these lexers highlight can hang the process.

The fix drops the backslash from the negated class and matches escapes with `\\[^\\]`, the same body form already used elsewhere in the tree and by the sibling rules in these files (Perl's other `s@...@` delimiters, urbi's closing-quote rule, Smithy's triple-quoted string, Rell's single-quoted string). That keeps the matched language identical for valid input while removing the ambiguous split. Existing example and snippet goldens are unchanged; I added a few `test_string_cataback` snippets as regression guards.


Support-itecz:
The lint failure came from the regexlint update that landed on master a few hours after this branch's last CI run, not from this change. Merged master to pick up bd22577; regexlint and the full test suite pass locally on the merged tree.

Support-itecz:
any update?

PR #3230, opened 2026-07-26
Optimize regexopt: exploit sorted order to cut regex_opt CPU by ~34%

jvoisin:
regex_opt_inner runs at words()-matcher compile time and is recursive. Its input list is always sorted (every recursive call preserves order), which lets several per-node operations be replaced with cheaper ones that produce byte-identical output:

- Prefix detection: the longest common prefix of a sorted list is just the common prefix of its first and last element. Replace commonprefix(strings) (a full min()/max() scan) with a direct comparison of first and strings[-1].

- Last-resort branch: since the list is sorted and first is its minimum, the strings sharing first's initial character form a contiguous front block. Replace groupby(..., lambda s: s[0] == first[0]) plus a generator/list/join with a single boundary scan and two slices, and drop the now-unused itertools.groupby import.

- make_charset: replace the per-match lambda in CS_ESCAPE.sub with a r'\\\g<0>' replacement template.

commonprefix is kept (now sharing a helper) for the suffix path, whose reversed list is not sorted and still needs min()/max().

Those changes are not only making regex_opt ~34% faster, but also making the code a bit less cryptic.

### DIFF
--- a/pygments/regexopt.py
+++ b/pygments/regexopt.py
@@ -11,27 +11,29 @@
 
 import re
 from re import escape
-from itertools import groupby
 from operator import itemgetter
 
 CS_ESCAPE = re.compile(r'[\[\^\\\-\]]')
 FIRST_ELEMENT = itemgetter(0)
 
 
-def commonprefix(m):
-    """Given an iterable of strings, returns the longest common leading substring"""
-    if not m:
-        return ""
-    s1 = min(m)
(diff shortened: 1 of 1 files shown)


birkenfeld:
I'm guessing the regexlint diagnostics are not related to this?

jvoisin:
I don't think it is, no

PR #3232, opened 2026-07-26
Remove superfluous parentheses from PostgresExplainLexer

jvoisin:
Unfortunately, I don't think there's an easy way to have a regexlinter check for that :/


birkenfeld:
LGTM! But why not? It already has Pygments specific info about the token assignment if I'm not wrong, e.g. to check for the correct number of groups for bygroups()

PR #3236, opened 2026-07-28
Escape the remaining HtmlFormatter and SvgFormatter attribute values

Support-itecz:
`HtmlFormatter` escapes `cssclass`, `cssstyles`, `filename`, `lineseparator`, `lineanchors` and `linespans` because they end up inside attributes, but `title`, `classprefix`, `prestyles` and `cssfile` reach the output the same way and go out raw. So do the token-type name components that build the `class` attribute (and the `title` attribute under `debug_token_types`), which a caller reaches through `NameHighlightFilter`'s `tokentype` option or a document lexed with `RawTokenLexer`. That is plain HTML attribute injection: `classprefix='"><b y="'` breaks out of every span, `HtmlFormatter(full=True, title='</title><script>alert(1)</script>')` lands a live script tag in the head, and `SvgFormatter` has the same hole in `fontfamily`/`fontsize`, in a document format browsers will run script from.

Route those values through `html_escape` at the same points the neighboring options already use it. The formatter is the only layer that can cover this, since it decides which attribute a value goes into and an embedding application cannot pre-escape for a placement it does not control. Valid input is unchanged: ordinary class prefixes, CSS declarations and font names carry no HTML metacharacters, and the ones that legitimately can (a quoted `font-family`, an apostrophe in a style) decode back to the same string inside the attribute.


Support-itecz:
any update?

PR #3239, opened 2026-07-29
Improved Solidity lexer

gandaro:
I used SolidityLexer.g4 and SolidityParser.g4 from ethereum/solidity to improve the rules in the Solidity lexer.

Removed keywords:

```
['block', 'inherited', 'msg', 'require', 'suicide', 'this', 'throw', 'tx']
```

Added keywords:

```
['abstract', 'after', 'alias', 'anonymous', 'apply', 'as', 'assembly', 'auto', 'byte', 'calldata', 'case', 'catch', 'copyof', 'days', 'default', 'define', 'delete', 'emit', 'enum', 'ether', 'event', 'fallback', 'final', 'gwei', 'hex', 'hours', 'immutable', 'implements', 'in', 'indexed', 'inline', 'interface', 'let', 'macro', 'match', 'minutes', 'mutable', 'null', 'of', 'override', 'partial', 'pragma', 'promise', 'pure', 'receive', 'reference', 'relocatable', 'sealed', 'seconds', 'sizeof', 'static', 'storage', 'supports', 'switch', 'try', 'type', 'typedef', 'typeof', 'unchecked', 'unicode', 'using', 'view', 'virtual', 'weeks', 'wei', 'years']
```

Improved the number detection and elementary types.


gandaro:
Note: More improvements could be:
* add fixed/ufixed types (but those aren't implemented yet, they just exist as lexer/parser rules)
* use a YUL lexer for `assembly` blocks
  * without a YUL lexer, disable Solidity rules within `assembly` blocks

PR #3241, opened 2026-08-01
escape Token.Escape in LatexFormatter unless escapeinside is set

Support-itecz:
LatexFormatter passes Token.Escape through without escaping so that LatexEmbeddedLexer can hand it verbatim LaTeX for the `escapeinside` option, but two ordinary lexers also emit that token type for their own escape syntax: apdlexer for `%...%` parameter substitution and ldap for backslash escapes inside a DN. An APDL file containing `%\immediate\write18{id}%` lands in the Verbatim block as-is, and that block declares `commandchars=\\\{\}`, so the backslash and braces are still active TeX. Typesetting the result with `-shell-escape`, which minted requires, runs the command; `\input{/etc/passwd}` works without it.

Gate the passthrough on `escapeinside`, since that is the only option that puts a LatexEmbeddedLexer in front of the formatter. The formatter is the layer that knows whether it asked for verbatim LaTeX, and the token type on its own cannot tell the two cases apart, so a check in the lexers would still leave third-party ones open. LDIF escapes come out correct now too instead of a stray control sequence.


Support-itecz:
any update?

PR #3254, opened 2026-08-15
Write RTF unicode escapes as signed 16 bit values

Sreekant13:
The RTF specification defines the argument of `\uN` as a signed 16 bit integer, and states that Unicode values greater than 32767 must be expressed as negative numbers. The RTF formatter writes them unsigned, so any code point above U+7FFF is emitted out of range:

```python
>>> from pygments import highlight
>>> from pygments.lexers.special import TextLexer
>>> from pygments.formatters import RtfFormatter
>>> highlight('\u9fa5', TextLexer(), RtfFormatter())   # CJK 龥
...{\u40869}...          # should be {\u-24667}
>>> highlight('\U0001F600', TextLexer(), RtfFormatter())   # emoji
...{\u55357}{\u56832}...  # should be {\u-10179}{\u-8704}
```

This affects a large part of the CJK ideograph range (everything from U+8000 up) and every non-BMP character, since the surrogate pair values are themselves above 32767. Word and other writers emit the negative form for these.

The fix converts the value where it is written, so anything at or below 32767 is untouched.

Note that this updates `test_double_characters`, which asserted the unsigned output for 힣, the variation selector and 鼖. Their expected values change from `55203`, `65038`, `55422`, `56859` to `-10333`, `-498`, `-10114`, `-8677`. Characters below the limit in that test (`4096`, `8597`) are unchanged.

Added `test_rtf_unicode_is_signed_16_bit` covering values either side of the boundary, and `test_rtf_unicode_never_exceeds_signed_16_bit` asserting no escape falls outside the range. The full RTF suite passes; five tests fail without the change.


Sreekant13:
Gentle nudge on this whenever the team has capacity. It makes the RTF formatter write `\uN` escapes as the signed 16 bit values the RTF spec requires, so CJK characters from U+8000 upward and all non-BMP characters are no longer emitted out of range. It does update an existing test that had encoded the unsigned output, which I have called out in the description. Thanks for maintaining pygments.

Sreekant13:
Gentle follow-up on this whenever the team has capacity. It is a small, tested fix for the RTF formatter writing `\uN` escapes as unsigned instead of the signed 16 bit value the RTF spec requires, so every CJK character from U+8000 up and all non-BMP characters currently go out of range. Still green on current main. Thanks for your work on Pygments.

PR #3264, opened 2026-08-18
PostScript: recognize immediately evaluated names as a single token (#1487)

### LINKED ISSUE #1487: PostScriptLexer -- token of the form //name
In PS2 we have "immediately evaluated name" tokens (section 3.12.2 in LPRM - "PS reference manual"):

//name

which actually is erronous highlighted.

Suggestion:

In lexers/graphics.py  replace line 350 (at '#References', for tokens in PostScriptLexer class):

(r'\/%s', ...

with:

(r'\/{1,2}%', ...

advitrocks9:
Fixes #1487.

`PostScriptLexer` emits an `Error` token for the first slash of `//name`, the Level 2 immediately evaluated name form, and then lexes the rest as an ordinary literal name:

```
>>> from pygments.lexers import get_lexer_by_name
>>> print(*get_lexer_by_name('postscript').get_tokens('//name'), sep='\n')
(Token.Error, '/')
(Token.Name.Variable, '/name')
(Token.Text.Whitespace, '\n')
```

The References rule in the `root` state only ever encoded the single-slash form:

```python
(rf'\/{valid_name}', Name.Variable),
```

`valid_name` starts with `[^{delimiter}]` and `/` is one of the delimiters, so the rule cannot consume a second slash. No other `root` rule starts with `/` either, so the engine takes its no-match path, emits `Error` for that one character and resumes at `/name`. Making the second slash optional is the whole fix:

```python
(rf'\/\/?{valid_name}', Name.Variable),
```

After:

```
(Token.Name.Variable, '//name')
(Token.Text.Whitespace, '\n')
```

`/name`, `/`, `//` and `/ name` are unchanged, and `tests/examplefiles/postscript/cells.ps` still produces the same 2736 tokens. One knock-on beyond the reported case: `///name` now sheds one `Error`, because the trailing `//name` forms a single name. It still leads with `Error`, and it is malformed PostScript either way. I left the neighbouring question of a bare `/` before a delimiter (arguably an empty literal name, currently `Error`) alone, that is a different rule and a separate PR.

Test is `tests/snippets/postscript/test_immediately_evaluated_name.txt`, covering the defining `/ND`, the reference `//ND`, and `///ND` to pin that a longer slash run still leads with `Error`. On master that file fails: the token diff shows a stray `'/' Error`. With the regex change it passes.

```
$ tox -- tests/snippets/postscript/   1 passed
$ tox -- -W error                     5331 passed, 16 skipped
$ tox -e check                        OK
$ tox -e regexlint                    OK
```


advitrocks9:
CI hasn't run on this yet, I think it's waiting on first-time contributor approval.

advitrocks9:
Could a maintainer approve the workflow run? CI hasn't run on this since it was opened.

PR #3266, opened 2026-08-19
JavaScript: recognize regex literals after `=>` and `...`

ekanshul:
The JavaScript root state matched `=>` and `...` as punctuation without entering the `slashstartsregex` state, so a regex literal right after an arrow (a very common shape, e.g. `const isUrl = s => /^https?:\/\//.test(s)`) was lexed as two division operators, and every backslash inside it became an `Error` token. The same happens after a spread.

Both tokens are always followed by an expression, so a slash after them can only start a regular expression literal. This change makes both rules push `slashstartsregex`, like the other operators do. It applies to all lexers derived from `JavascriptLexer` (TypeScript, JSX, TSX, ...).

Found by lexing real-world code (gitea's `web_src`) and looking for `Error` tokens; for example [gitea/web_src/js/utils/url.ts](https://github.com/go-gitea/gitea/blob/main/web_src/js/utils/url.ts#L2):

```ts
export const urlRawRegex = () => /\bhttps?:\/\/[^\s<>[\]]+/gi;
```

A snippet test is added covering an arrow with a block-less body, an arrow followed by a method call on the regex, and a spread.


ekanshul:
Data point: with this change, lexing all 4,346 `.ts` files of gitea's `web_src` plus a Next.js monorepo produces zero `Error` tokens (the TypeScript lexer inherits the rule). Before, a single `.filter((line) => /^[a-z0-9]+: '/.test(line))` desynchronized the lexer for the remaining ~800 lines of the file because the regex was read as division and the `'` inside it opened a string.

PR #3314, opened 2026-09-13
Fix catastrophic backtracking in TADS 3 single-line comment lexing

### LINKED ISSUE #3311: Super-linear (exponential) execution time in Tads3Lexer on adversarial input
Per the guidance in `doc/docs/security.rst` ("When reporting a bug where you
suspect super-linear execution times, please make sure to attach an input to
reproduce it") — here is a reproducing input and the measured scaling.

**Affected:** `Tads3Lexer` (`pygments/lexers/int_fiction.py`)
**Tested against:** Pygments 2.21.0 (current `master`, commit `38f426a`)

### Root cause

`pygments/lexers/int_fiction.py:770`:

```python
_comment_single = r'(?://(?:[^\\\n]|\\+[\w\W])*$)'
```

`[^\\\n]` explicitly excludes `\`, so a run of *N* backslashes right after
`//` can be split between the two alternatives (`[^\\\n]` matches nothing on
a backslash, `\\+[\w\W]` matches 1+ backslashes) in `2^(N-1)` different ways.
Whenever the rest of the pattern this fragment is embedded in fails to match
(e.g. no closing token, no newline, end of input reached mid-comment), the
engine exhaustively tries every split — classic catastrophic backtracking.

This fragment is reused (via `_ws`/`_ws_pp`) in roughly 20 token rules across
the lexer, e.g. line 978: `(extern)({_ws}+)(object\b)`.

### Reproduction

A single unterminated `//` comment followed by a run of backslashes is
enough — no other TADS 3 syntax needed:

```python
import time
from pygments.lexers.int_fiction import Tads3Lexer
from pygments.formatters import NullFormatter

lexer = Tads3Lexer()
fmt = NullFormatter()
prev = None
for n in [20, 24, 26, 28, 30, 32, 34]:
    src = "//" + "\\" * n
    t0 = time.perf_counter()
    highlight(src, lexer, fmt)
    dt = time.perf_counter() - t0
    ratio = f" (x{dt/prev:.2f})" if prev else ""
    print(f"n={n:3d}  len={len(src):3d}  time={dt:9.4f}s{ratio}")
    prev = dt
```

Measured (isolated regex, `_comment_single` + a trailing literal that can
never match — same shape, faster to iterate):

| n (backslashes) | input length | time |
|---:|---:|---:|
| 20 | 22 | 0.0008 s |
| 24 | 26 | 0.0052 s |
| 26 | 28 | 0.0138 s |
| 28 | 30 | 0.0359 s |
| 30 | 32 | 0.0936 s |

Full token rule `(extern)({_ws}+)(object\b)` (the actual pattern used in the
lexer's `tokens` table, still isolated `re.match`, not going through
`highlight()`):

| n (backslashes) | time |
|---:|---:|
| 30 | 0.15 s |
| 32 | 0.43 s |
| 34 | 1.01 s |
| 36 | 2.62 s |
| 38 | 7.05 s |
| 40 | 18.16 s |

And confirmed end-to-end through the public `pygments.highlight()` API
(the reproduction script above, not just the isolated regex): a 36-byte
source file (`"//"` + 34 backslashes, nothing else) already takes **1.22s**;
growth is consistently ~1.5-1.6x per additional byte. Extrapolating, a
comment of well under 100 bytes is enough to occupy a worker for minutes.

The trigger is a single unterminated `//` comment ending in backslashes —
plausible as an accidental paste/truncation, and trivial to craft
adversarially for any service that syntax-highlights user-supplied `.t`/TADS
3 source (e.g. code-snippet sites, "paste with highlighting" tools using
`get_lexer_by_name`/`guess_lexer`).

I have not modified anything in the checkout — this is read-only analysis
plus a standalone timing script.

devtechedge:
`Tads3Lexer._comment_single` matched each character of a `//` comment with either of two alternatives, `[^\\n]` or `\+[\w\W]`. Since `\+[\w\W]` can also match a backslash, a run of N backslashes after `//` could be split between the two branches in 2**(N-1) ways, and whenever no newline ended the comment the regex engine tried every split. Highlighting `"//" + "\\" * n` therefore scaled exponentially, as shown in the reproduction script in #3311: on this machine n=28 took 0.06s and n=40 took 19.7s.

The fix changes `\+[\w\W]` to `\[\w\W]`, so the second alternative consumes exactly one backslash plus one more character. The alternatives are now disjoint on the first character and matching is linear; the same input now highlights in 0.013s.

Tests:

- A bounded regression test in `tests/test_int_fiction.py` (same style as the recently added `tests/test_lisp.py`) times highlighting of 40 backslashes after `//` and asserts it completes in under 2s.
- Snippets under `tests/snippets/tads3/` pin tokenization of comments ending in even and odd backslash runs, terminated by newline and by end of input.

One behavior change worth noting: because `$` matches before a newline under `re.MULTILINE`, the old pattern could pull the line following a backslash-terminated `//` comment into the comment token. The refreshed golden for `tads3_example.t` shows the affected lines; for example `#define Room Unthing` after such a comment is now highlighted as `Comment.Preproc` instead of being swallowed by the comment.

Fixes #3311


devtechedge:
Hi - gentle check-in on the TADS 3 comment RedOS fix whenever you have a moment.

`Tads3Lexer._comment_single` in `pygments/lexers/int_fiction.py` paired `[^\\n]` with `\+[\w\W]`, so a long run of backslashes after `//` could be split between those branches in exponentially many ways (`Fixes #3311`).

The one-character change to `\[\w\W]` makes the alternatives disjoint; on the repro from the issue, n=40 dropped from about 19.7s to about 0.013s here.

`tests/test_int_fiction.py` adds a RedOS-style regression, and three snippets under `tests/snippets/tads3/` cover odd, even, and EOF backslash runs; the refreshed `tads3_example.t` golden reflects `#define` lines after backslash-terminated `//` comments tokenizing as `Comment.Preproc` instead of being swallowed.

Local replication was green (`pytest -W error`, ruff, check_sources, analyse_text, token refs, regexlint); GitHub Actions is still waiting on first-contribution approval, which is expected.

Happy to tighten the snippets, adjust the golden, or rework the note about `$` under `re.MULTILINE` if that would help review.

No rush at all - still ready from my end whenever it fits.

